PR #71, opened 2019-05-15
Stubbing out protocPath so that it can be overridden in google3

robcapo:


robcapo:
We can’t use runfiles before google.Init() is called, so it has to be called after the test starts.

Thanks,
Rob Capo (via iPhone)

PR #74, opened 2019-05-21
Updated installation to use `go get`.

mingzhi:


README.md
```diff
@@ -1,30 +1,27 @@
...
+
+* Install `api-linter` using `go get`:
+
 ```sh
-cd $HOME/Downloads/api-linter/cmd/api-linter
-go install
+go get github.com/googleapis/api-linter/cmd/api-linter
 ```
-* Ensure that your operating system's `PATH` contains `go install`ed binaries, as described in the
-  [Go install guide](https://golang.org/doc/install#install).
```
robcapo: Should we keep this link? Also, I don't think `$HOME/go/bin` is always the right location. For example in [the Linux / Mac instructions](https://golang.org/doc/install#tarball), it recommends `/usr/local/go`

mingzhi: I believe /usr/local/go/bin contains the go binaries, but not the ones installed by a user through `go get`, which will be in `$GOPATH/bin`. By default, `$GOPATH` is `$HOME/go`. If a user sets `$GOPATH`, he/she is experienced enough to figure out where the go binaries are installed.

See https://golang.org/doc/install#testing

robcapo: Ah I see, makes sense.

PR #76, opened 2019-06-10
Added support for resolving dependencies among FileDescriptors

robcapo:


rules/testutil/testutil.go
```diff
@@ -13,95 +13,119 @@ import (
...
-	return protoset.GetFile()[0], nil
+	protoset.GetFile()[0].Name = &filename
+
+	return protoset.GetFile()[0]
 }
 
-func mustCloseAndRemoveFile(f *os.File) {
-	if err := f.Close(); err != nil {
-		log.Fatalf("Error closing proto file: %v", err)
+func mustCreateDescSetFileFromFileDescriptorProtos(descs []*descriptorpb.FileDescriptorProto) *os.File {
```
mingzhi: Let's rewrite this function to `writeFileDescriptorSet(filename string, s *descriptorpb.FileDescriptorSet) error`, so the caller can create the temp file in a consistent place and remember to remove it.

robcapo: Any particular reason for this preference? I think it's a bit cleaner for the caller to not have to deal with the temp file as well as handling two extra errors (one from ioutil.TempFile and one from this function), but it really just seems like a matter of preference. Also, this is an unexported method with only one usage, so this is an extremely low impact decision.

rules/check_field_names_use_lower_snake_case_test.go
```diff
@@ -45,9 +46,13 @@ func TestFieldNamesUseLowerSnakeCaseRule(t *testing.T) {
 	rule := checkFieldNamesUseLowerSnakeCase()
 
 	for _, test := range tests {
-		req := testutil.MustCreateRequestFromTemplate(tmpl, test)
-
 		errPrefix := fmt.Sprintf("Check field name `%s`", test.FieldName)
+
+		req, err := lint.NewProtoRequest(testutil.MustCreateFileDescriptorProtoFromTemplate("", tmpl, test, nil), nil)
```
mingzhi: I have no clue figuring out what the arguments are in `MustCreateFileDescriptorProtoFromTemplate`. Maybe we need to encompass those arguments to a struct.

lint/registry.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	var err error
+	if e.desc, err = protodesc.NewFile(e.descProto, f.reg); err != nil {
+		return err
+	}
+	if err = f.reg.Register(e.desc); err != nil {
+		return err
+	}
+	e.registering = false
```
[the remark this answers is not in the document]
robcapo: It's just an extra check to guard against cycles. Otherwise we risk infinite loop.

lint/registry_test.go
```diff
@@ -0,0 +1,69 @@
+package lint
+
+import (
+	"github.com/googleapis/api-linter/rules/testutil"
+	"google.golang.org/protobuf/types/descriptorpb"
+	"testing"
+)
+
+func TestMakeRegistryFromAllFiles(t *testing.T) {
```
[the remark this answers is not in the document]
robcapo: is there any particular case you'd like to see covered?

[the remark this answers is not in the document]
mingzhi: -> means depends
A -> B -> C
A->B, A->C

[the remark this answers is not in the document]
robcapo: Done

lint/linter.go
```diff
@@ -18,6 +18,7 @@ package lint
 
 import (
 	"errors"
+	"google.golang.org/protobuf/reflect/protoregistry"
```
mingzhi: Fix the import order and grouping.

lint/registry.go
```diff
@@ -0,0 +1,88 @@
...
+	"google.golang.org/protobuf/reflect/protodesc"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// MakeRegistryFromAllFiles creates a *protoregistry.Files with all dependencies present in descs resolved.
+// Missing dependencies will be filled with placeholders. If there are cyclic dependencies, an error will
+// be returned.
+func MakeRegistryFromAllFiles(descs []*descriptorpb.FileDescriptorProto) (*protoregistry.Files, error) {
```
mingzhi: Make this function private -- this will be replaced once upstream protoreflect support it.

lint/registry.go
```diff
@@ -0,0 +1,91 @@
...
+type files struct {
+	reg *protoregistry.Files
+
+	filesSet map[string]*entry
+}
+
+type entry struct {
+	descProto   *descriptorpb.FileDescriptorProto
+	desc        protoreflect.FileDescriptor
+	registering bool
```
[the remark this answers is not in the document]
mingzhi: `visited` is a common word in graph search, which will help others understand the algorithm here.

lint/registry.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (f *files) register(path string) error {
+	e, ok := f.filesSet[path]
+
+	// if it's already been registered (from another import) do nothing. if it doesn't exist,
```
mingzhi: Remove comments here, instead, create a function `e.isRegistered()` that returns true if `e.desc != nil`, i.e., codes should speak itself.

lint/registry.go
```diff
@@ -0,0 +1,88 @@
...
+	// also do nothing, and a placeholder will be filled in for those that import it
+	if !ok || e.desc != nil {
+		return nil
+	}
+
+	if e.registering {
+		// this should never happen, because protoc doesn't allow cyclic imports
```
mingzhi: Remove comments here.

lint/registry.go
```diff
@@ -0,0 +1,88 @@
...
+		if err := f.register(dep); err != nil {
+			return err
+		}
+	}
+
+	var err error
+	if e.desc, err = protodesc.NewFile(e.descProto, f.reg); err != nil {
+		return err
+	}
+	if err = f.reg.Register(e.desc); err != nil {
```
mingzhi: Change to `if err := ...`

lint/registry.go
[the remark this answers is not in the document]
mingzhi: In that case, use `defer` -- what happens if it returns error above and registering is not reset?

lint/registry.go
```diff
@@ -0,0 +1,88 @@
...
+	e.registering = false
+
+	return nil
+}
+
+func makeFilesSet(files []*descriptorpb.FileDescriptorProto) (map[string]*entry, error) {
+	filesSet := make(map[string]*entry, len(files))
+
+	for _, f := range files {
+		filesSet[f.GetName()] = &entry{descProto: f}
```
mingzhi: Will the file names are all unique? Does `protoc` guarantee it?

lint/registry_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Fatalf("foo.Fields().Len()=%d; want 1", foo.Fields().Len())
+	}
+
+	if foo.Fields().Get(0).Message() == nil {
+		t.Fatalf("foo.Fields().Get(0).Message() was nil")
+	}
+
+	if foo.Fields().Get(0).Message().Name() != "Bar" {
+		t.Fatalf("foo.Fields().Get(0).Message().Name() = %q; want %q", foo.Fields().Get(0).Message().Name(), "Bar")
+	}
```
mingzhi: Check if it is a placeholder.

lint/registry_test.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func TestMakeRegistryFromAllFiles_DirectAndIndirectDependencies(t *testing.T) {
```
mingzhi: Check placeholder.

lint/registry_test.go
```diff
@@ -0,0 +1,146 @@
...
+	if baz.Fields().Get(1).Message() == nil {
+		t.Fatalf("baz.Fields().Get(1).Message() was nil")
+	}
+
+	if baz.Fields().Get(1).Message().Name() != "Bar" {
+		t.Fatalf("baz.Fields().Get(1).Message().Name() = %q; want %q", baz.Fields().Get(1).Message().Name(), "Bar")
+	}
+}
+
+func TestMakeRegistryFromAllFiles_MissingImports(t *testing.T) {
```
mingzhi: Check if it is a placeholder.

lint/registry.go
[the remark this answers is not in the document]
robcapo: `visted` does not accurately describe what's happening. It's past tense, which will be confusing, because after a given file is registered it will be set to false. `visiting` is more accurate, but the function in named `register`, so this more accurately describes "the algorithm is currently in the `register` function for this `entry`.

lint/registry.go
robcapo: I'm not a fan of this level of indirection. Adding more functions takes up more space and makes it harder for the reader to click through what's happening. A function like `isRegistered` seems obvious to us, but a reader will still have to click through to understand what that means. I will remove the comment if you think this code is obvious enough.

lint/registry.go
robcapo: Why?

lint/registry.go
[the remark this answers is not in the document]
robcapo: It's still safe, because errors propagate all the way up, but sure, I'll use defer.

lint/registry.go
robcapo: I think the explanation that a placeholder will be filled is useful.

rules/check_field_names_use_lower_snake_case_test.go
robcapo: IntelliJ automatically fills in the names of the parameters, so it's easy to see "filename", "srcTmpl", "data", "deps". I'm not against the idea of using a struct, but that also makes it a bit more cumbersome to use. Maybe I can restructure the functions to be more obvious.

robcapo: Decided to go with struct.

lint/registry.go
robcapo: protoregistry will ignore duplicates, so this assumption is not imposing any additional constraints than its dependencies. I'm not sure whether protoc guarantees uniqueness or not, but the entire thing will not function without uniqueness anyway. From the proto descriptor, it's the "file name, relative to root of source tree". It sounds like this should be unique, but they don't define the "root of the source tree".

lint/registry.go
mingzhi: nit point.

PR #80, opened 2019-06-14
[WIP] Initial AIP-131 rules.

lukesneeringer:
This is not ready to be merged (there are no tests, formatting is likely wrong, there may even be syntax errors as I am on a plane, etc. -- this is my getting my toe in the water). I would appreciate feedback on whether this is more or less correct, though.

I also have a few comments and concerns, which we should figure out how/where to tacke:

* **P0**: There are no rule codes. We really need these -- I need to be able to say that the "there must be a `name` field in Get methods" rule is "E0131A" or something similar.
* **P1**: Can we add a link field on `Problem`? We should be including the citation links.
* **P1**: There is still quite a bit of boilerplate per rule. We want rules to be very small so they can be disabled individually (it would be bad if the three rules in this PR were one rule with multiple `if` statements). I understand if nothing can be done about this, but I want to ask.
* **P2**: All linter rules should be enforcement of AIPs. Can we double down on a naming convention that reflects this?


robcapo:
Regarding rule codes: it sounds like you're suggesting something similar to the intention of `Name`. Is there some way you can formulate rules with codes in a namespace that you can parse? I don't really understand the intention here.
`RuleInfo` has a `URI` field that should work for your purpose. It turns out, we're not actually using that field, so I created #81.
I'm not sure if we can reduce it much more than we already have. There's a certain amount of meta information that we need in order to post findings effectively. Could you point to specific areas that feel like boilerplate?
Are you referring to naming conventions of `RuleInfo.Name`? Could you explain the distinction between this and rule codes?

lukesneeringer:
> Regarding rule codes: it sounds like you're suggesting something similar to the intention of `Name`. Is there some way you can formulate rules with codes in a namespace that you can parse? I don't really understand the intention here.

Linters like [ESLint](https://eslint.org/) and [flake8](http://flake8.pycqa.org/en/latest/) assign codes to particular errors (like `E408` or `W533`, as strawmen). The general breakdown is that `E` and `W` mean error and warning, and then the numeric codes are quasi-arbitrary. ESLint uses short codes without numbers, like `semi`.

I am not really sure what `lint.NewRuleName` does, so it is tough for me to know whether it serves this purpose or how to use it correctly. I think we are likely to have a _lot_ of rules though (AIP-131 alone will probably have ten), and I think using numeric codes will be valuable. Essentially, a code like `E0131A` should communicate that you have an error, and that it is related to AIP-131. (The final letter would be a disambiguator since single AIPs will need multiple rules.)

That also lets someone disable rules intelligently; disabling `E0131B` allows you to silence a very specific rule, while leaving the others, while disabling `E0131*` could disable all rules related to that AIP, if an API is ignoring the AIP wholesale.

> `RuleInfo` has a `URI` field that should work for your purpose.

Ah, perfect. I missed that!

> I'm not sure if we can reduce it much more than we already have. There's a certain amount of meta information that we need in order to post findings effectively. Could you point to specific areas that feel like boilerplate?

Lines 37-49 feel like a lot, but idiomatic Go tends to be verbose. I do understand what each line is doing. I am content here, I think.

> Are you referring to naming conventions of `RuleInfo.Name`? Could you explain the distinction between this and rule codes?

I actually had filenames in mind when I said this. Basically, say there is a bug in a linter rule, or I get a lint error and I want to understand it better. If the linter says I am breaking AIP-131, I should be able to quickly find the AIP-131 file in the linter, and I would expect it to be named something like `aip_0131.go` or `E0131.go` or something similar, and _not_ `check_get_methods.go`.

My rationale here is again that there are very likely to be many, many linter rules. We have an indexing system, and as the number of linter rules get large, it will be increasingly important to be able to find the one you want in the tree.

rules/aip_0131.go
```diff
@@ -0,0 +1,138 @@
...
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get- methods for the purpose of this rule;
+				// ignore everything else.
+				if strings.HasPrefix(string(m.Name()), "Get") {
+					return
+				}
+
+				// Rule check: Establish that there are no other fields besides `name`.
+				fields := m.Input().Fields()
+				for i := 0; i < fields.Len(); i++ {
```
lukesneeringer: Is this really the best way to do this? (I tried to `range` over the object to get fields, but it does not work.)

robcapo: Unfortunately this is the way the protoreflect API is defined for now. I think it was initially written by C++ devs so it's not quite idiomatic Go.

I believe they mentioned they are adding a more idiomatic API, but until then, this is how I've been doing it.

rules/0131.go
```diff
@@ -0,0 +1,185 @@
...
+			Description:  "check that Get RPCs have appropriate request messages",
+			RequestTypes: []lint.RequestType{lint.ProtoRequest},
+			URI:          "https://aip.dev/131#guidance",
+		},
+		Callback: descriptor.Callbacks{
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get- methods for the purpose of this rule;
+				// ignore everything else.
+				methodName := string(m.Name())
+				if !strings.HasPrefix(methodName, "Get") {
```
alin04: Should probably introduce a library that encapsulates this type of logic for example, "isGetMethod", "isUpdateMethod", or maybe "isStandardMethod", "isCustomMethod".  Dunno, thinking out loud.

rules/0131.go
```diff
@@ -0,0 +1,185 @@
...
+				// ignore everything else.
+				if !strings.HasPrefix(string(m.Name()), "Get") {
+					return
+				}
+
+				// Rule check: Establish that there are no other fields besides `name`.
+				fields := m.Input().Fields()
+				for i := 0; i < fields.Len(); i++ {
+					field := fields.Get(i)
+					if string(field.Name()) != "name" {
```
alin04: Some fields are accepted, e.g. show_deleted, or a view field.

rules/0131.go
```diff
@@ -0,0 +1,185 @@
...
+					}
+				}
+
+				return problems, nil
+			},
+		},
+	}
+}
+
+// Get messages should use the resource as the respose message
```
alin04: s/respose/response

rules/0131_test.go
```diff
@@ -0,0 +1,223 @@
...
+		startLine    int
+	}{
+		{"GetFooRequest", 0, "", -1},
+		{"GetFooReq", 1, "GetFooRequest", 5},
+	}
+
+	rule := checkRequestMessageName()
+
+	for _, test := range tests {
+		req := testutil.MustCreateRequestFromTemplate(tmpl, test)
```
alin04: Consider using t.Run to run each test.

rules/0131.go
lukesneeringer: Oooh I forgot about views and read masks, thanks.
Is `show_deleted` valid for Get? I thought it was just List. (Honestly asking.)

rules/0131_test.go
lukesneeringer: I blindly copied this section. I defer to you and @mingzhi.

rules/0131.go
lukesneeringer:
```suggestion
// Get messages should use the resource as the response message
```

rules/aip_0131.go
lukesneeringer: Blargh, but thanks. :)

rules/0131.go
lukesneeringer: Added everything except `show_deleted`, which I _think_ does not belong in Gets. (Totally willing to add it if needed.)

rules/0131.go
lukesneeringer: I went to write this, but it seems like I would be replacing...

```go
if !strings.HasPrefix(methodName, "Get") {
	return
}
```

with...

```go
if isGetMethod(m) {
	return
}
```

...which does not actually seem better (particularly since I need `methodName` for other reasons frequently).

100% willing to be told I am missing something; I am out of my element here.

lukesneeringer: One argument for this might be exceptions. For example, @garrettjonesgoogle recommended using `GetEffective-` as a slightly different prefix, and we would want to exempt it. Same for `GetIamPolicy`.

Okay, I talked myself into it.

lukesneeringer: Done. Thanks!

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+				if !isGetMethod(m) {
+					return
+				}
+
+				// Rule check: Establish that a name field is present.
+				nameField := m.Input().Fields().ByName("name")
+				if nameField == nil {
+					problems = append(problems, lint.Problem{
+						Message:    fmt.Sprintf("method %q has no `name` field", m.Name()),
+						Descriptor: m.Input(),
```
mingzhi: The lint warning should show up in the RPC method, not in the input message, i.e., we should return `m` here.

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+					})
+					return problems, nil
+				}
+
+				// Rule check: Establish that the name field is a string.
+				if nameField.Kind() != p.StringKind {
+					problems = append(problems, lint.Problem{
+						Message:    "`name` field on Get RPCs should be a string",
+						Descriptor: nameField,
```
mingzhi: Same question here -- where should the lint warning show up?

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+	return &descriptor.CallbackRule{
+		RuleInfo: lint.RuleInfo{
+			Name:         lint.NewRuleName("core", "0131", "request_message", "unknown_fields"),
+			Description:  "check that there are no unknown fields",
+			RequestTypes: []lint.RequestType{lint.ProtoRequest},
+			URI:          "https://aip.dev/131#request-message",
+		},
+		Callback: descriptor.Callbacks{
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get methods for the purpose of this rule;
```
mingzhi: Remove comments.

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+				// ignore everything else.
+				if !isGetMethod(m) {
+					return
+				}
+
+				// Rule check: Establish that there are no unexpected fields.
+				allowedFields := map[string]struct{}{
```
mingzhi: Why not use `map[string]bool`? And we can then write `if !allowedFields[string(field.Nam())] {...}`

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+				fields := m.Input().Fields()
+				for i := 0; i < fields.Len(); i++ {
+					field := fields.Get(i)
+					if _, ok := allowedFields[string(field.Name())]; !ok {
+						problems = append(problems, lint.Problem{
+							Message: fmt.Sprintf(
+								"Get RPCs should only only contain fields explicitly described in AIPs, not %q.",
+								string(field.Name()),
+							),
+							Descriptor: field,
```
mingzhi: Same question -- where should the lint warning show up? In the service method section, or the message field?

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+	return &descriptor.CallbackRule{
+		RuleInfo: lint.RuleInfo{
+			Name:         lint.NewRuleName("core", "0131", "response_message", "name"),
+			Description:  "check that Get RPCs have appropriate response messages",
+			RequestTypes: []lint.RequestType{lint.ProtoRequest},
+			URI:          "https://aip.dev/131#guidance",
+		},
+		Callback: descriptor.Callbacks{
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get methods for the purpose of this rule;
```
mingzhi: Remove comments.

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+				// ignore everything else.
+				if !isGetMethod(m) {
+					return
+				}
+
+				// Rule check: Establish that for methods such as `GetFoo`, the response
+				// message is named `Foo`.
+				responseMessageName := string(m.Output().Name())
+				methodName := string(m.Name())
+				if methodName != "Get"+responseMessageName {
```
mingzhi: Compare the response message name instead: `if suggestion := methodName[3:]; responseMessageName != suggestion {...}`

rules/0131.go
alin04: The empty struct is more memory efficient, and the map value (true) is not valuable. It's not useful to ever have a value of false - only the presence of map key matters. I would wrap this is in an isAllowed() func for better readability.

mingzhi: I think the current list is small, that does not impact readability.

rules/0131.go
lukesneeringer: I want to to show up on the field in this case.

robcapo:
> > Regarding rule codes: it sounds like you're suggesting something similar to the intention of `Name`. Is there some way you can formulate rules with codes in a namespace that you can parse? I don't really understand the intention here.
> 
> Linters like [ESLint](https://eslint.org/) and [flake8](http://flake8.pycqa.org/en/latest/) assign codes to particular errors (like `E408` or `W533`, as strawmen). The general breakdown is that `E` and `W` mean error and warning, and then the numeric codes are quasi-arbitrary. ESLint uses short codes without numbers, like `semi`.
> 
> ~I am not really sure what `lint.NewRuleName` does, so it is tough for me to know whether it serves this purpose or how to use it correctly.~ Okay, I see that `lint.NewRuleName` essentially provides namespacing to serve a similar purpose. It just means that you effectively have to have a lookup table in your head between AIPs and their rules.
> 
> However, I think we are likely to have a _lot_ of rules though (AIP-131 alone will probably have ten), and I think using alphanumeric short codes will be valuable. Essentially, a code like `E0131A` should communicate that you have an error, and that it is related to AIP-131. (The final letter would be a disambiguator since single AIPs will need multiple rules.)
> 
> Maybe we can do this with rule names as it exists, e.g. `core::0131::{descriptive_name}`. I would probably argue for a mandated structure in that case -- e.g. the AIP number should _always_ be one of the components.
> 
> > `RuleInfo` has a `URI` field that should work for your purpose.
> 
> Ah, perfect. I missed that!
> 
> > I'm not sure if we can reduce it much more than we already have. There's a certain amount of meta information that we need in order to post findings effectively. Could you point to specific areas that feel like boilerplate?
> 
> Lines 37-49 feel like a lot, but idiomatic Go tends to be verbose. I do understand what each line is doing. I am content here, I think.
> 
> > Are you referring to naming conventions of `RuleInfo.Name`? Could you explain the distinction between this and rule codes?
> 
> I actually had filenames in mind when I said this. Basically, say there is a bug in a linter rule, or I get a lint error and I want to understand it better. If the linter says I am breaking AIP-131, I should be able to quickly find the AIP-131 file in the linter, and I would expect it to be named something like `aip_0131.go` or `E0131.go` or something similar, and _not_ `check_get_methods.go`.
> 
> My rationale here is again that there are very likely to be many, many linter rules. We have an indexing system, and as the number of linter rules get large, it will be increasingly important to be able to find the one you want in the tree.

Ok, I want to try to distill a few important points / questions here:

1. What meta information (if any) might vary among `Problem`s emitted by a single `Rule` (keeping in mind that we may have many `Rule`s per AIP). Obviously `Location` is one (i.e. if a `Rule` finds the same error in a bunch of different places, the `Problem`s will be posted with different `Location`s. Is there anything else that can be different?
2. `RuleInfo.Name` is used to identify an implementation of the `Rule` interface. We can then customize the output from that `Rule` via the configuration file. This can be used to change the output from a `Rule` based on the files that it's operating on (e.g. maybe some `google::corp::*` rules show up as `error`s for anything in `google/corp/` but only `warning`s elsewhere).
3. In theory, the config mentioned in (2) can be changed at runtime or even request time. This is not the case in google3 currently (we're using a blaze rule to embed it in the MPM), but it's just a yaml file, so improvements can be made to change the config without rebuilding the binary. In the CLI, we can customize it with the `--cfg` flag.

I wanted to explicitly mention those so we can try to satisfy your requirements leveraging as much of the existing functionality as possible and avoid rework / throwing things away unless it's necessary.

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+				// ignore everything else.
+				if !isGetMethod(m) {
+					return
+				}
+
+				// Rule check: Establish that for methods such as `GetFoo`, the request
+				// message is named `GetFooRequest`.
+				methodName := string(m.Name())
+				requestMessageName := string(m.Input().Name())
+				if requestMessageName != methodName+"Request" {
```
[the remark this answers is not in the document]
lukesneeringer: Done.

rules/0131.go
lukesneeringer: `m.Input()` is correct. The problem is on the input message, not the method; having the lint warning show up on the RPC would be more difficult for the producer.

rules/0131.go
lukesneeringer: On the field is intentional and correct.

rules/0131.go
lukesneeringer: Done.

rules/0131.go
```diff
@@ -0,0 +1,202 @@
...
+	return &descriptor.CallbackRule{
+		RuleInfo: lint.RuleInfo{
+			Name:         lint.NewRuleName("core", "0131", "request_message", "name"),
+			Description:  "check that Get RPCs have appropriate request messages",
+			RequestTypes: []lint.RequestType{lint.ProtoRequest},
+			URI:          "https://aip.dev/131#guidance",
+		},
+		Callback: descriptor.Callbacks{
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get methods for the purpose of this rule;
```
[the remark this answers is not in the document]
lukesneeringer: I do not want to remove comments. I find narrative comments useful when I look at code later (and engineers consistently tell me this is a valuable practice).

rules/0131.go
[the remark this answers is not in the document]
lukesneeringer: Done.

lukesneeringer:
> What meta information (if any) might vary among Problems emitted by a single Rule (keeping in mind that we may have many Rules per AIP)? Obviously Location is one (i.e. if a Rule finds the same error in a bunch of different places, the Problems will be posted with different Locations. Is there anything else that can be different?

I think location is the main thing. I have written a couple of rules that might warn against a message or a field (e.g. "field X is missing" goes on the message, "field X is present but wrong" goes on the field).

I have also used different messages ("field missing" vs. "field wrong").

> RuleInfo.Name is used to identify an implementation of the Rule interface.

I am not sure what this sentence means.
That said, I think I am mostly convinced of the wisdom of this plan. However, we probably need to be prescriptive about exactly what the proper namespace is.

robcapo:
> > RuleInfo.Name is used to identify an implementation of the Rule interface.
> 
> I am not sure what this sentence means.
> That said, I think I am mostly convinced of the wisdom of this plan. However, we probably need to be prescriptive about exactly what the proper namespace is.

Sure, so what I mean is that, each `Rule` implementation has a `Info() RuleInfo` method. And the `Name` identifies that `Rule`. I just wanted to clarify that `Name` identifies a `Rule` and not a `Problem`, so it would be hard, for example, to emit E0101A and E0101B from the same `Rule`.

Furthermore, `Name` can be used to map the `Problem`s emitted by a `Rule` to a specific [`Category`](https://github.com/googleapis/api-linter/blob/master/lint/response.go#L47), by way of the yaml configuration, and this can be customized per file that's being linted.

lukesneeringer:
Oh! 💡 That makes complete sense.

In the existing linter, is it possible to enable / disable _rules_, _problems_, or both?

rules/0131.go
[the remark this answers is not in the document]
mingzhi: Got it. Thanks!

robcapo:
> Oh!  That makes complete sense.
> 
> In the existing linter, is it possible to enable / disable _rules_, _problems_, or both?

You can disable `Rule`s via yaml configuration, and you can disable `Problem`s via comments in your proto source for a specific location.

lukesneeringer:
> You can disable Rules via yaml configuration, and you can disable Problems via comments in your proto source for a specific location.

That makes sense, although Problems do not seem to have names, so it is not entirely clear to me yet how to disable them.

lukesneeringer:
Okay, this is now ready to merge, I think. I still expect some tweaking as the team figures out what rules should look like, but nothing is served by holding this as a PR anymore.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	github.com/bmatcuk/doublestar v1.1.1
+	github.com/golang/protobuf v1.2.1-0.20190516215712-ae2eaafab405
```
mingzhi: Remove this line. I don't think we depend on this package in this PR.

rules/0131.go
mingzhi: If a user wants to disable this rule, it will be on the input message as well, right? Currently, we are using the returned descriptor (here is the `m.Input()`) to check if the rule is disabled on the leading comments of the descriptor.

lukesneeringer:
> If a user wants to disable this rule, it will be on the input message as well, right? Currently, we are using the returned descriptor (here is the `m.Input()`) to check if the rule is disabled on the leading comments of the descriptor.

I would assume that users would disable rules wherever the problem shows up.

PR #85, opened 2019-06-17
Executing rules inside a function that recovers from panics.

robcapo:
This will prevent a panic inside a rule from impacting the rest of the linter rules.

### DIFF
--- a/lint/config.go
+++ b/lint/config.go
@@ -109,6 +109,16 @@ func ReadConfigsYAML(f io.Reader) (Configs, error) {
 	return c, nil
 }
 
+// MakeConfigsForAllPaths takes an input RuleConfig and applies it to all Rules and all file paths.
+func MakeConfigsForAllRulesAndPaths(rc RuleConfig) Configs {
+	return []Config{{
+		IncludedPaths: []string{"**"},
+		RuleConfigs: map[string]RuleConfig{
+			"": rc,
+		},
+	}}
+}
+
 // getRuleConfig returns a RuleConfig that matches the given path and rule.
 // Returns an error if a config is not found for the path.
 func (c Configs) getRuleConfig(path string, rule RuleName) (result RuleConfig, err error) {
--- a/lint/config.go
+++ b/lint/config.go
@@ -109,16 +109,6 @@ func ReadConfigsYAML(f io.Reader) (Configs, error) {
 	return c, nil
 }
 
-// MakeConfigsForAllPaths takes an input RuleConfig and applies it to all Rules and all file paths.
-func MakeConfigsForAllRulesAndPaths(rc RuleConfig) Configs {
-	return []Config{{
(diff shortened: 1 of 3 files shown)


lint/linter.go
```diff
@@ -102,3 +103,17 @@ func (l *Linter) run(req Request) (Response, error) {
 
 	return resp, err
 }
+
+func (l *Linter) runAndRecoverFromPanics(rl Rule, req Request) (probs []Problem, err error) {
```
mingzhi: Can you add a test? Thanks!

robcapo: Done.

lint/config.go
```diff
@@ -109,6 +109,16 @@ func ReadConfigsYAML(f io.Reader) (Configs, error) {
 	return c, nil
 }
 
+// MakeConfigsForAllPaths takes an input RuleConfig and applies it to all Rules and all file paths.
+func MakeConfigsForAllRulesAndPaths(rc RuleConfig) Configs {
```
mingzhi: Do we need to make it public? Can we move it to linter_test.go?

robcapo: Yeah, we can move it in. I made it public because I figured it would be useful to other tests, but we can move it if we need it.

mingzhi: I think it is better to make it private and only for tests, so we don't need to maintain it.

lint/linter_test.go
```diff
@@ -109,3 +110,29 @@ func TestLinter_run(t *testing.T) {
...
+func (r *panickingRule) Lint(_ Request) ([]Problem, error) { panic("panic") }
+
+func TestLinter_LintProtos_RulePanics(t *testing.T) {
+	r, err := NewRules(&panickingRule{})
+	if err != nil {
+		t.Fatalf("Failed to create Rules: %q", err)
+	}
+
+	fd := new(descriptorpb.FileDescriptorProto)
+	fd.SourceCodeInfo = new(descriptorpb.SourceCodeInfo)
```
mingzhi: Do we need this line?

lint/linter_test.go
```diff
@@ -109,3 +110,29 @@ func TestLinter_run(t *testing.T) {
...
+
+	// linter with only one rule, and a default configuration which enables all rules for all files
```
mingzhi: Remove this comment -- the code is clear enough.

lint/linter.go
```diff
@@ -102,3 +103,17 @@ func (l *Linter) run(req Request) (Response, error) {
...
+	defer func() {
+		if r := recover(); r != nil {
+			if rerr, ok := r.(error); ok {
```
mingzhi: Can we test this condition?

lint/linter_test.go
robcapo: Yes, otherwise creating request will fail.

PR #86, opened 2019-06-17
[WIP] AIP-132 rules.

lukesneeringer:
These are AIP-132 rules similar to the AIP-131 rules added in #80.
As in AIP-131, they are incomplete, because we are not yet able to
read annotations.


rules/0132.go
```diff
@@ -0,0 +1,195 @@
...
+				// Rule check: Establish that there are no unexpected fields.
+				allowedFields := []string{
+					"parent", "page_size", "page_token",
+					"filter", "order_by", "group_by", "show_deleted",
+				}
+				fields := m.Input().Fields()
+				for i := 0; i < fields.Len(); i++ {
+					field := fields.Get(i)
+					allowed := false
+					for _, allowedField := range allowedFields {
```
lukesneeringer: @alin04 @mingzhi This feels gross. Is there really no way to do the equivalent of Python's `if field.name in allowed_fields`?
(I checked Stack Overflow and it suggested there really is no better way to do it, but the question was old.)

alin04: Turn allowedFields into a map. Iterate over fields and check existence (or non-existence) in the map.

lukesneeringer: Yay thanks!
Could I be super lazy and ask you to provide sample code? Stack Overflow did not give me a useful answer.

alin04:
```
allowedFields := map[string]struct{}{
  "parent": struct{}{},
  "page_size": struct{}{},
  ...
}

if _, ok := allowedFields[field]; !ok {
  problems = append(problems, ...}
}
```

The goofy empty struct is a space saving optimization.

rules/0132.go
```diff
@@ -0,0 +1,195 @@
...
+				// We only care about List- methods for the purpose of this rule;
+				// ignore everything else.
+				methodName := string(m.Name())
+				if !strings.HasPrefix(methodName, "List") {
+					return
+				}
+
+				// Rule check: Establish that for methods such as `ListFoos`, the response
+				// message is named `ListFoosResponse`.
+				responseMessageName := string(m.Output().Name())
```
alin04: Suggest encapsulating this into a re-usable function. The requirement that the response match the method name + response is common to Create, List, and all custom methods.  Don't love this copy&paste...

rules/0132.go
lukesneeringer: Thanks!

robcapo: You can also eliminate the `struct{}` part, so that it looks like `"parent": {},`.

robcapo: Also, just to elaborate a little bit on the justification for this. One of the philosophies behind Go is to generally be explicit about computation (there are exceptions to this). In other words, if your list was really long, and you had `if field.name in allowed_fields`, you might not realize that it's an O(N) operation.

It can lead to boilerplate, but after reading about why some things like this are missing from the language, I felt better about the decisions. There's a lot about Go that seems a ugly or verbose at first, but there's _usually_ a decent reason for it.

lukesneeringer: Well, except `__contains__` is not necessarily O(N). On `set`, for example, it is the same hash lookup being used here (and I would have used a set in this case).

But that is neither here nor there. If this is the right thing, and what a Go programmer expects to see, then I am content. :)

rules/0132.go
lukesneeringer: Done. :)

rules/0132.go
lukesneeringer: Done. :)

rules/0132.go
```diff
@@ -0,0 +1,201 @@
...
+		},
+		Callback: descriptor.Callbacks{
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about List- methods for the purpose of this rule;
+				// ignore everything else.
+				if !isListMethod(m) {
+					return
+				}
+
+				// Rule check: Establish that a name field is present.
```
alin04: s/name/parent

rules/0132.go
```diff
@@ -0,0 +1,201 @@
...
+				parentField := m.Input().Fields().ByName("parent")
+				if parentField == nil {
+					problems = append(problems, lint.Problem{
+						Message:    fmt.Sprintf("method %q has no `parent` field", m.Name()),
+						Descriptor: m.Input(),
+					})
+					return problems, nil
+				}
+
+				// Rule check: Establish that the name field is a string.
```
alin04: s/name/parent

rules/0132.go
```diff
@@ -0,0 +1,201 @@
...
+					})
+				}
+
+				return problems, nil
+			},
+		},
+	}
+}
+
+// Get methods should not have unrecognized fields.
```
alin04: s/Get/List

rules/0132.go
```diff
@@ -0,0 +1,201 @@
...
+func init() {
+	registerRules(
+		checkListRequestMessageName(),
+		checkListRequestMessageParentField(),
+		checkListRequestMessageUnknownFields(),
+		checkListResponseMessageName(),
+	)
+}
+
+// Get messages should have a properly named Request message.
```
alin04: s/Get/List

lukesneeringer:
> This all seems like it can be config driven. E.g. check that a message definition contains required field(s) of specified type. Also, that a message only contains allowed fields from a given set.
> 
> All this boilerplate is so much code duplication.

I largely agree, it feels like a _lot_ of work to wire up pretty simple rules. Also, the error handling seems excessive (3-4 error checks per rule) -- is there really no way to make that better?

rules/0132.go
```diff
@@ -0,0 +1,201 @@
...
+
+				return problems, nil
+			},
+		},
+	}
+}
+
+// Return true if this is a List method, false otherwise.
+func isListMethod(m p.MethodDescriptor) bool {
+	return strings.HasPrefix(string(m.Name()), "List")
```
robcapo: Borderline nitpick: I imagine the corner cases are small, but this might be safer with a regex like `List(?:[A-Z]|$)` in other words, you may hit false positives with words that begin with `List`. `Listen`, `Listener`, etc. come to mind.

robcapo:
> > This all seems like it can be config driven. E.g. check that a message definition contains required field(s) of specified type. Also, that a message only contains allowed fields from a given set.
> > All this boilerplate is so much code duplication.
> 
> I largely agree, it feels like a _lot_ of work to wire up pretty simple rules. Also, the error handling seems excessive (3-4 error checks per rule) -- is there really no way to make that better?

Are you guys talking about the test for the actual rules? Also, is this a Go thing or an api-linter thing? I'm not sure I can see all the boilerplate you're talking about.

Regarding error checks (I assume you're talking about the tests, because I don't see any error checks in your rule), this is one of the biggest complaints of Go. It's largely unavoidable, unless [this](https://golang.org/issue/32437) happens.

One area that looks a bit like duplication is the RuleInfo, but even this has different values for 3 out of the 4 fields in all cases. Aside from that, you're checking different fields, and your checking them for different things. It doesn't feel much like boilerplate to me, but then again, I've been writing Go all quarter :)

Request: can you link me to the implementations of these rules in the python linter?

rules/0132.go
lukesneeringer:
```suggestion
				// Rule check: Establish that a parent field is present.
```

rules/0132.go
lukesneeringer:
```suggestion
				// Rule check: Establish that the name field is a string.
```

rules/0132.go
lukesneeringer:
```suggestion
// List methods should not have unrecognized fields.
```

rules/0132.go
lukesneeringer: I buy that.

lukesneeringer:
> Request: can you link me to the implementations of these rules in the python linter?

Bad citizenship on my part: I am not actually sure the equivalent rules exist. Sorry....
(I used the AIPs as my launch point, not existing rules.)

> Are you guys talking about the tests or the actual rules? Also, is this a Go thing or an api-linter thing? I'm not sure I can see all the boilerplate you're talking about.

Some of both.

I have been discussing this with @alin04 in another venue and I expect that for rules, we can make a series of utility functions that would reduce the effort. For example, it will be really common to want to check that a message has a given type of field, so we might need to have an `EnsureField` function we can call that takes a message, checks its fields, and does the right thing. You can generalize from that example to probably ~15 or so utility functions that might make this a lot easier.

The tests feel much worse. The setup involves a lot of repetitive error handling, and I have a hard time believing it is completely a Golang thing. I _expect_ (and I want to dig in myself and try it) that what we need is a utility function that takes a template and a struct (and maybe a string prefix) and handles all that error work.

These are really items that I think I should be going and trying to make better and not a "gripe list".

rules/0132.go
lukesneeringer: Done.

rules/0132.go
```diff
@@ -0,0 +1,201 @@
...
+					"read_mask":    {}, // AIP-157
+					"view":         {}, // AIP-157
+				}
+				fields := m.Input().Fields()
+				for i := 0; i < fields.Len(); i++ {
+					field := fields.Get(i)
+					if _, ok := allowedFields[string(field.Name())]; !ok {
+						problems = append(problems, lint.Problem{
+							Message:    "List RPCs should only contain fields explicitly described in AIPs.",
+							Descriptor: field,
```
mingzhi: Do we all agree that a rule should be disabled at the place the warning is shown up?

lukesneeringer: Yes.

Issue #84, opened 2019-06-15
Common imports.


lukesneeringer:
We need to be able to resolve imports for common protos (e.g. `google/api/annotations.proto`) in order to write many rules.

I assert that any protos defined in [this repository](https://github.com/googleapis/api-common-protos) should be able to be imported in test templates.

mingzhi:
Agree. But we need to be find a way to have a local (and synced) copy of the mentioned repository (we have one in Google3).

lukesneeringer:
That repository downsyncs from google3. :) It copies all the directories listed in [AIP-213](https://aip.dev/213).

mingzhi:
Rob, can you please take a look at it? Thank you!

lukesneeringer:
@robcapo I could believe that this might be tricky. If you want, I can ask some of the Golang heavy hitters on my team to help, since they have experience writing golang / protoc stuff.

robcapo:
I have some thoughts (and open questions) on the different ways we can do this, and I'm actually writing a test in google3 that needs it now, so it's become priority 1 for me. I think your idea about a submodule externally makes sense, and we can just use the corresponding google3 paths internally.

The mechanics of passing multiple files into protoc is also straightforward. We can either compile the entire repository into a descriptor set and pass that in, or we can pass the base directories in via proto_path.

If you want to reach out to the Golang team and get some advice, that'd be great. In the meantime I'll try to throw together some prototypes and see if we can get something working (since I'm blocked by it anyway). Thanks!

lukesneeringer:
Adding @noahdietz and @pongad for vis -- Noah and Michael, if you have advice for us on this, we would really appreciate it.

noahdietz:
Let me reiterate what I think is being discussed and y'all tell me if I'm off:

* `api-common-protos` is a dependency for many of the rules that will be written (e.g. validating a `google.api.resource` annotation).
* It would be nice if the linter could provide the `api-common-protos` (for both binary & Go package usage) so that a user doesn't have to clone the repo & provide them (e.g. via `--proto_path`). 

The questions are:
* how do we package/ship `api-common-protos` with a linter binary?
* how do we expose `api-common-protos` to the linter (pkg & binary protoc exec)?

lukesneeringer:
> It would be nice if the linter could provide the `api-common-protos` (for both binary & Go package usage) so that a user doesn't have to clone the repo & provide them (e.g. via `--proto_path`).

[...]

PR #88, opened 2019-06-18
[chore] Install stdlib protos in test environment.

lukesneeringer:
This adds the `google/protobuf/` protos in the place where
protoc looks for them, allowing them to be imported in tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
 
             curl -OL https://github.com/google/protobuf/releases/download/v3.7.1/$PROTOC_ZIP
 
-            sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc
+            sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc include/*
 
             rm -f $PROTOC_ZIP
       - run: go test -mod=readonly ./...


.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
             curl -OL https://github.com/google/protobuf/releases/download/v3.7.1/$PROTOC_ZIP
 
-            sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc
+            sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc include/*
```
robcapo: I don't really like the idea of getting the stdlib via install instructions. I think we'd be better off with a submodule so that we can version control them, and keep it in sync with our rules.

robcapo: Ah you know what, I was thinking that these were the api-common-protos that we were discussing in #84. These ones are probably much less likely to change?

robcapo: This will break for zsh, change `include/*` `'include/*'`.

lukesneeringer: > I don't really like the idea of getting the stdlib via install instructions. I think we'd be better off with a submodule so that we can version control them, and keep it in sync with our rules.

I think you figured this out already, but for clarity: this is the actual install instructions in protoc (unzip the archive you are downloading and look at the readme and it tells you to do precisely this).

lukesneeringer: > This will break for zsh, change `include/*` to `'include/*'`.

Who cares? The image is not using zsh.

robcapo: Ah, fair. I'm using zsh and I copy pasted it :)

PR #89, opened 2019-06-18
Allowing clients of testutil to add custom `--proto_path` arguments

robcapo:
Added api-common-protos as a submodule under the testutil dir (this should be the only time we need it AFAIK). This is a solution to issue #84.

Also added a test that uses it.

FWIW I also had to follow @lukesneeringer's update to the installation instructions (#88), because there were deps on them. I think I'd rather include them as a submodule if there's an easy way of doing it.


lukesneeringer:
> FWIW I also had to follow @lukesneeringer's update to the installation instructions (#88), because there were deps on them. I think I'd rather include them as a submodule if there's an easy way of doing it.

There is not really -- those protos are part of protocol buffers, not part of their own repo.
However, what I am doing in #88 is the canonical protoc installation instructions. (protoc is not installed properly if it is not done.)

rules/testutil/testutil.go
```diff
@@ -39,10 +41,13 @@ func descriptorProtoFromSource(source io.Reader) (*descriptorpb.FileDescriptorPr
 	}
 	defer mustCloseAndRemoveFile(descSetF)
 
+	_, thisFilePath, _, _ := runtime.Caller(0)
```
mingzhi: Will it work in Google3? Should we use the hack way as `protocPath`?

mingzhi: Also make `testutil` an internal package -- it is no longer suitable for other libraries outside this repo to use it.

robcapo: This is a good point, and actually requires some restructuring that I've wanted to do for a while. I'm going to move this out of the `rules` package and into `api-linter/internal/testutil`.

I also think that the descriptor helpers should be in `api-linter/` and that `rules` should be strictly for `Rule` implementations and their tests, but we can decide that separately.

mingzhi: I prefer to creating a new testing util package in and only for google3 -- as we already saw that the internal infrastructure is quite different. Otherwise, we will always have concerns about compatibility.

testutil/testutil.go
```diff
@@ -19,25 +19,27 @@ var protocPath = func() string {
...
+	// Filename is the output of the returned FileDescriptorProto.GetName().
 	Filename string
+	// Template defines a text/template to use for the proto source.
 	Template string
-	Data     interface{}
-	Deps     []*descriptorpb.FileDescriptorProto
+	// Data is plugged into the template to create the full source code.
+	Data interface{}
+	// Deps are any additional FileDescriptorProtos that the protocol compiler will need for the source.
+	Deps []*descriptorpb.FileDescriptorProto
```
mingzhi: Maybe we should make it consistent to `protoc` arguments:
`Deps` -> `DescriptorSetIn` (`FileDescriptorSet`)
`AdditionalProtoPaths` -> `ProtoPaths`

testutil/testutil.go
```diff
@@ -19,25 +19,27 @@ var protocPath = func() string {
...
+	AdditionalProtoPaths []string
 }
 
-// MustCreateFileDescriptorProtoFromTemplate creates a *descriptorpb.FileDescriptorProto from a string template and data.
-func MustCreateFileDescriptorProtoFromTemplate(spec FileDescriptorSpec) *descriptorpb.FileDescriptorProto {
-	tmpl := template.Must(template.New("test").Parse(spec.Template))
-	b := new(bytes.Buffer)
-	if err := tmpl.Execute(b, spec.Data); err != nil {
+// MustCreateFileDescriptorProtoFromSpec creates a *descriptorpb.FileDescriptorProto from a string template and data.
+func MustCreateFileDescriptorProtoFromSpec(spec FileDescriptorSpec) *descriptorpb.FileDescriptorProto {
```
mingzhi: I think we can shorten the function name by removing `FromSpec`.

testutil/testutil.go
robcapo: I considered it. One thing that's nice about this setup is that this function outputs FileDescriptorProtos already, so it's easy to do something like:
```
f1 := MustCreateFileDescriptorFromSpec(...)
f2 := MustCreateFileDescriptorFromSpec(FileDescriptorSpec{ ..., Deps: []*descriptorpb.FileDescriptorProto{f1}})
```

testutil/testutil.go
robcapo: Done.

testutil/testutil.go
```diff
@@ -62,8 +64,12 @@ func mustCreateDescriptorProtoFromSource(filename string, source io.Reader, deps
...
 	}
 
-	if len(deps) > 0 {
-		descSetIn := mustCreateDescSetFileFromFileDescriptorProtos(deps)
+	for _, p := range spec.AdditionalProtoPaths {
+		args = append(args, fmt.Sprintf("--proto_path=%s", p))
+	}
+
+	if len(spec.Deps) > 0 {
+		descSetIn := mustCreateDescSetFileFromFileDescriptorProtos(spec.Deps)
```
mingzhi: I think we can remove `FromFileDescriptorProtos`, otherwise the function name is really, really long.

testutil/testutil_test.go
```diff
@@ -0,0 +1,30 @@
+package testutil
+
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func TestDescriptorFromProtoSource_CustomProtoPaths(t *testing.T) {
```
mingzhi: Can you add a test for the `descriptor_set_in`?

testutil/testutil.go
robcapo: Done.

testutil/testutil_test.go
robcapo: Done.

rules/testutil/testutil.go
robcapo: Resolving because this code is now removed

PR #91, opened 2019-06-21
Created map from import directories to full import paths

robcapo:
When syncing internally with google3, the locations of proto dependencies (like api-common-protos) are in different locations. In this PR we create a function that maps from a dependency name to its location, so that we can override it when we sync to google3.

### DIFF
--- a/rules/0131_test.go
+++ b/rules/0131_test.go
@@ -170,10 +170,7 @@ func TestGetRequestMessageUnknownFields(t *testing.T) {
 	for _, test := range tests {
 		errPrefix := "AIP-131 Request Unknown Fields"
 		req, err := lint.NewProtoRequest(testutil.MustCreateFileDescriptorProto(
-			testutil.FileDescriptorSpec{
-				Template: tmpl,
-				Data:     test,
-			},
+			testutil.FileDescriptorSpec{Template: tmpl, Data: test},
 		), nil)
 		if err != nil {
 			t.Errorf("%s: lint.NewProtoRequest returned error %v", errPrefix, err)
--- a/rules/0131_test.go
+++ b/rules/0131_test.go
@@ -171,9 +171,8 @@ func TestGetRequestMessageUnknownFields(t *testing.T) {
 		errPrefix := "AIP-131 Request Unknown Fields"
 		req, err := lint.NewProtoRequest(testutil.MustCreateFileDescriptorProto(
 			testutil.FileDescriptorSpec{
-				Template:             tmpl,
-				Data:                 test,
-				AdditionalProtoPaths: []string{testdatadir("api-common-protos")},
+				Template: tmpl,
+				Data:     test,
 			},
 		), nil)
 		if err != nil {
--- a/rules/0131_test.go
+++ b/rules/0131_test.go
@@ -1,0 +1,253 @@
+// Copyright 2019 Google LLC
+//
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rules
+
+import (
+	"testing"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/testutil"
+)
+
+func TestGetRequestMessageName(t *testing.T) {
+	tmpl := `
+	syntax = "proto3";
+
+	service Aip131 {
+		rpc GetFoo({{ .RequestName }}) returns (Foo);
+	}
+
+	message {{.RequestName }} {
+		string name = 1;
+	}
+
+	message Foo {}`
+
+	tests := []struct {
+		RequestName  string
+		problemCount int
+		suggestion   string
+		startLine    int
+	}{
+		{"GetFooRequest", 0, "", -1},
+		{"GetFooReq", 1, "GetFooRequest", 5},
+	}
+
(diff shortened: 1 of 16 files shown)


rules/0131_test.go
```diff
@@ -170,7 +170,11 @@ func TestGetRequestMessageUnknownFields(t *testing.T) {
 	for _, test := range tests {
 		errPrefix := "AIP-131 Request Unknown Fields"
 		req, err := lint.NewProtoRequest(testutil.MustCreateFileDescriptorProto(
-			testutil.FileDescriptorSpec{Template: tmpl, Data: test},
+			testutil.FileDescriptorSpec{
+				Template:             tmpl,
+				Data:                 test,
+				AdditionalProtoPaths: []string{testdatadir("api-common-protos")},
```
mingzhi: Why adding this?

rules/rules_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+
+	"github.com/googleapis/api-linter/testutil"
+)
+
+// This function maps testdata directories to their real directories. The convoluted mechanism for
+// doing so is an unfortunate consequence of maintaining compatibility when syncing internally.
+var testdatadir = func(lib string) string {
+	_, f, _, _ := runtime.Caller(0)
+	testdataDir := filepath.Dir(f) + string(os.PathSeparator) + "testdata" + string(os.PathSeparator)
```
mingzhi: Why not return the `testdataDir` and they can use it as `filepath.Join(testdataDir, "api-common-protos")`? So we don't need to hard code "api-common-protos" or add more cases.

robcapo: my main concern is that in google3, the protos are spread out more, so it may not map cleanly to a single dir. for example, api-common-protos live in several directories spread out across google3/.

Also for this specific case, it wouldn't work, because the testdata dir would have to be `google3/` and they would then join `google3/api-common-protos` which doesn't exist

rules/0131_test.go
robcapo: Good catch, I think i was originally using that test to test the common protos. removed.

rules/rules_test.go
mingzhi: Can we make including api-common-protos to the `proto_path` implicitly, similar to how we define `$PATH`? In other words, we always add `--proto_path=${testdataDir}/api-common-protos`, so:
1.  in the tests, we don't need to worry about where is `google/type/date.proto`;
2.  it allows us to do some future changes, e.g., we may want to rename `api-common-protos` or merge two folders;
3. if some tests need other proto files, (similar to `$PATH`), we append the path.

testutil/testutil_test.go
```diff
@@ -1,17 +1,37 @@
...
-	"runtime"
+	"os"
 	"testing"
+
+	"google.golang.org/protobuf/types/descriptorpb"
 )
 
 func TestDescriptorFromProtoSource_CustomProtoPaths(t *testing.T) {
-	_, thisFilePath, _, _ := runtime.Caller(0)
+	dirName := os.TempDir() + string(os.PathSeparator) + "api-linter-fake-dir"
```
mingzhi: This folder name is not random -- it may conflict -- use `ioutil.TempDir` instead.

testutil/testutil_test.go
```diff
@@ -1,17 +1,37 @@
...
+	err := os.Mkdir(dirName, 0777)
+	if err != nil {
+		t.Fatalf("Failed to create dependency directory: %s", err)
+	}
+	defer os.RemoveAll(dirName)
+
+	fh, err := os.Create(dirName + string(os.PathSeparator) + "sample.proto")
```
mingzhi: Why not use `filepath.Join`?

testutil/testutil_test.go
```diff
@@ -1,17 +1,37 @@
...
+	if err != nil {
+		t.Fatalf("Failed to create sample.proto: %s", err)
+	}
+
+	_, err = fh.WriteString(`syntax = "proto3";
+package testdata;
```
mingzhi: Indent please, it affects readability.

testutil/testutil_test.go
```diff
@@ -1,17 +1,37 @@
...
+message Sample {
+	string foo = 1;
+}
+`)
```
mingzhi: Indent.

rules/rules_test.go
robcapo: I think what you're suggesting has some advantage in the form of convenience, but it's implicitly adding a dependency from testutil to api-common-protos. My argument against it is the same as the one against compiling the protos into the binary. I don't think we should couple these protos to our implementation.

There are very few tests that actually use those protos right now, so keeping it explicit is not costly at all (1 line of code per test that uses it). It would also be trivial to create a factory function in the `rules` package that creates a `FileDescriptorContext` with `AdditionalProtoPaths` pre-filled. Then all of the references to `api-common-protos` would be contained to the `rules` package (where the directory already lives).

By keeping the dependency explicit, we remove the expectation that anything added to `testdata` will automatically be available in tests, and tests that have different dependencies don't have to worry about any naming conflicts with the protos in `api-common-protos`

I guess if we move `api-common-protos` to `testutil/` it would make a bit more sense, but otherwise there's a strange cyclic dependency between testutil and rules. I'm thinking of `testutil` as a simple utility that allows you to specify your source code and its dependencies, rather than a fully packaged proto environment. If we decide to move the protos into testutil, I think we should make it an internal package. WDYT?

mingzhi: I am confused -- I thought we were discussing about this `testdataDir` function -- I was suggesting something like 
```go
var common_proto_paths = func() []string {
 // register your proto paths here and we will return them
}
```
 to replace `testdataDir`.

mingzhi: I can live with the current version, so I am resolving this conversation.

PR #94, opened 2019-06-23
Allow multiple --proto_path

pcj:
Modifies the CLI arguments to allow multiple `--proto_path` flags that are propagated to the protocol compiler.


mingzhi:
I am not sure if package `cli` supports multiple presence of `--proto_path` (e.g., `--proto_path=a --proto_path=b`. But `protoc` supports `--proto_path=a:b:.`, which corresponds to `--proto_path=a --proto_path=b --proto_path=.`. Thanks!

pcj:
The `urfave/cli` library supports repeatable flags via the StringSlice flag type.  

I've always used multiple -I (--proto_path) flags to `protoc` rather than folding them into a colon-delimited list, but I guess that should work too.

cmd/api-linter/cli.go
```diff
@@ -55,10 +55,10 @@ func runCLI(rules lint.Rules, configs lint.Configs, args []string) error {
 					Value: "protoc",
 					Usage: "protocol compiler path",
 				},
-				cli.StringFlag{
+				cli.StringSliceFlag{
 					Name:  "proto_path",
-					Value: ".",
-					Usage: "the directory in which for protoc to search for imports",
+					Value: &cli.StringSlice{"."},
```
robcapo: What's the behavior of this value? If the user specifies value(s) will they override this slice or append to it?

lukesneeringer: Since 2016, it overrides: https://github.com/urfave/cli/issues/160

PR #108, opened 2019-07-11
[aip-132] make missing parent message more descriptive

noahdietz:
When I ran the linter, the AIP-132 rule for List* RPCs having a parent field in the input message was emitted. I felt that the message didn't make a lot of sense 

>method "ListProductsInProductSet" has no `parent` field

A method doesn't have fields and maybe it could be referring to the output message.

This PR makes the message more explicit/actionable by including the offending input message name.

### DIFF
--- a/rules/0132.go
+++ b/rules/0132.go
@@ -91,7 +91,7 @@ func checkListRequestMessageParentField() lint.Rule {
 				parentField := m.Input().Fields().ByName("parent")
 				if parentField == nil {
 					problems = append(problems, lint.Problem{
-						Message:    fmt.Sprintf("method %q has no `parent` field", m.Name()),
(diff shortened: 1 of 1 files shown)


noahdietz:
cc: @lukesneeringer as the rule author

noahdietz:
Thanks for both approving this PR. But it looks like CI is stuck? I can't see a status from it (unless I'm not supposed to?).

robcapo:
Try merging master and pushing it back up.

noahdietz:
I rebased against `master` and I'm still seeing a  `Waiting for status...`.

Screen Shot 2019-07-24 at 8 09 24 AM

lukesneeringer:
@noahdietz I fixed the CI to build forks. Can you push a dummy commit?

lukesneeringer:
Merging the Renovate PR gives me something to merge here, which solves the same problem.

PR #120, opened 2019-07-23
Extract DescriptorSource from FileDescriptor.SourceLocations.

mingzhi:
This is the first step to remove dependency on `FileDescriptorProto` in the `lint` package. In other words, the linter will expect a _resolved_ FileDescriptor that has source locations.

### DIFF
--- a/lint/request.go
+++ b/lint/request.go
@@ -39,12 +39,19 @@ func (r Request) DescriptorSource() DescriptorSource {
 }
 
 // NewProtoRequest creates a linting Request for a .proto file.
+//
+// Deprecated: Use NewRequestFromFileDescriptor instead.
 func NewProtoRequest(fd *descriptorpb.FileDescriptorProto, reg *protoregistry.Files) (Request, error) {
 	f, err := protodesc.NewFile(fd, reg)
 	if err != nil {
 		return Request{}, err
 	}
-	s, err := newDescriptorSource(fd)
+	return NewRequestFromFileDescriptor(f)
+}
+
+// NewRequestFromFileDescriptor creates a linting Request from the given FileDescriptor.
+func NewRequestFromFileDescriptor(f protoreflect.FileDescriptor) (Request, error) {
+	s, err := newDescriptorSource(f)
 	return Request{
 		fileDesc:   f,
 		descSource: s,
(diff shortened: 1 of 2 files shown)


robcapo:
Do we know if they've merged this internally yet? We have to be careful about using new functionality in protoreflect because it may block us from syncing the linter.

lint/source.go
```diff
@@ -63,34 +62,35 @@ func newLocPath(p ...int) locPath {
...
 
-	for _, loc := range sci.GetLocation() {
-		var path []int
-		for _, v := range loc.GetPath() {
-			path = append(path, int(v))
+	for i := 0; i < sci.Len(); i++ {
+		loc := sci.Get(i)
+		path := make([]int, len(loc.Path))
+		for k := range path {
+			path[k] = int(loc.Path[k])
```
robcapo: I'm not sure if this is a nitpick but I got yelled at for this in go readability review. The recommended way is using append, so you could do something like:

```
path := make([]int, 0, len(loc.Path))
for k := range path {
  path = append(path, int(loc.Path[k]))
}
```

Not blocking on this just wanted to let you know.

lint/request.go
```diff
@@ -39,12 +39,19 @@ func (r Request) DescriptorSource() DescriptorSource {
...
 	f, err := protodesc.NewFile(fd, reg)
 	if err != nil {
 		return Request{}, err
 	}
-	s, err := newDescriptorSource(fd)
+	return NewRequestFromFileDescriptor(f)
+}
+
+// NewRequestFromFileDescriptor creates a linting Request from the given FileDescriptor.
+func NewRequestFromFileDescriptor(f protoreflect.FileDescriptor) (Request, error) {
```
robcapo: Should we keep it private unless someone needs it?

mingzhi:
Protoreflect has not been synced since 3 weeks ago. So this is not ready to be merged to master. I will continue develop in this branch. Can you ping them to sync theirs?

lint/request.go
mingzhi: Since it will not been merged soon, I will keep it public.

lint/source.go
mingzhi: Make sense. It saves future mistakes that might result index out of bound errors. I will change it accordingly. Thanks!

robcapo:
LGTM with the exception that we can't yet sync it internally.

mingzhi:
I am merging it to the newly created `dev-sync-blocked` branch, which will be main place for commits that cannot be merged to the `master` branch because syncing with Google3 is an issue.

PR #126, opened 2019-08-06
Modify AIP-131 and AIP-132 rules to never cross a file.

lukesneeringer:
This commit modifies the 131 and 132 rules to never potentially
cross a file between what is being checked and where the rule
is reported.


rules/0132.go
```diff
@@ -195,7 +195,12 @@ func checkListResponseMessageName() lint.Rule {
 	}
 }
 
-// Return true if this is a List method, false otherwise.
+// Return true if this is an AIP-132 List method, false otherwise.
 func isListMethod(m p.MethodDescriptor) bool {
```
alin04: There is a lot of duplication throughout all this code. I wonder if some utility methods might help. E.g. https://play.golang.org/p/qFgJXFdDrMV

lukesneeringer:
Concur that duplication is high. Also concur that this PR is not the place to address it.

PR #132, opened 2019-08-06
[AIP-133] Add initial AIP-133 rules.

lukesneeringer:
Incomplete rules for AIP-133 (still need to parse annotations).


rules/0133.go
```diff
@@ -0,0 +1,227 @@
...
+
+				return problems, nil
+			},
+		},
+	}
+}
+
+// Return true if this is a AIP-133 Create method, false otherwise.
+func isCreateMethod(m p.MethodDescriptor) bool {
+	return regexp.MustCompile("^Create(?:[A-Z]|$)").MatchString(string(m.Name()))
```
alin04: Is there another lint check before this that ensures both method and message names don't have underscores and other disallowed characters? Otherwise this regexp won't capture things like Create_Foo.

Also, consider pulling the regexp out into a global variable. MustCompile panics if the expression can't be parsed - so better to do that earlier (and once) than later and each time this method is invoked.

rules/0133.go
```diff
@@ -0,0 +1,227 @@
...
+		Callback: descriptor.Callbacks{
+			MethodCallback: func(m p.MethodDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get methods for the purpose of this rule;
+				// ignore everything else.
+				if !isCreateMethod(m) {
+					return
+				}
+
+				// Sanity check: If the response type is an LRO, accept this.
+				// TODO: Check the response type annotation.
```
alin04: TODO should reference an issue? Unless you plan to do a very fast follow PR.

rules/0133.go
```diff
@@ -0,0 +1,227 @@
...
+				responseMessageName := string(m.Output().Name())
+				if responseMessageName == "Operation" {
```
alin04: Is the package stripped for responseMessageName? Given:

`rpc CreateFoo(CreateFooRequest) returns (google.longrunning.Operation)`

rules/0133.go
lukesneeringer: Yeah, my intent is for it to be imminent. I am just stuck because I am trying to figure out how to parse annotations, which I am actively working on.

rules/0133.go
lukesneeringer: Yes, it is, although I since learned that `FullName` exists and would do what I want. I will update this.

rules/0133.go
lukesneeringer: I think it is reasonable to have other checks to complain about "disallowed characters as a whole".

I would expect tests to fail if `MustCompile` panicked? Am I missing something?

alin04: Yes, tests will fail. It's also an optimization to compile the regexp once than repeatedly. Prefer global var declaration.

rules/0133.go
lukesneeringer: Done.

rules/0133.go
lukesneeringer: Done.

rules/0133.go
lukesneeringer: Going to leave this as-is. The follow-up will be fast.

rules/0133.go
```diff
@@ -0,0 +1,230 @@
...
+		Callback: descriptor.Callbacks{
+			MessageCallback: func(m p.MessageDescriptor, s lint.DescriptorSource) (problems []lint.Problem, err error) {
+				// We only care about Get methods for the purpose of this rule;
+				// ignore everything else.
+				if !isCreateRequestMessage(m) {
+					return
+				}
+
+				// Determine the name of the resource being created.
+				resource := regexp.MustCompile("^Create([A-Za-z0-9]+)Request$").FindStringSubmatch(string(m.Name()))[1]
```
alin04: This can use the global var now too.

lukesneeringer: Done!

PR #144, opened 2019-08-09
[AIP-131] Lint the google.api.http annotation

lukesneeringer:
This adds linting of the `google.api.http` annotation on Get requests.
It also adds a friendly "getHTTPRules" function to range over the http declarations.


rules/0131.go
```diff
@@ -71,6 +73,84 @@ func checkGetRequestMessageName() lint.Rule {
...
+					return
+				}
+
+				// Iterate over the HTTP rules.
+				for _, httpRule := range getHTTPRules(m) {
+					// Rule check: Ensure that the GET HTTP verb is used.
+					getURI := httpRule.GetGet()
+					if getURI == "" {
+						problems = append(problems, lint.Problem{
+							Message:    "Get RPCs should use the GET HTTP verb.",
```
alin04: I haven't been too thorough. Are you using "should" vs "must" as appropriate in these error messages?

rules/utils.go
```diff
@@ -0,0 +1,46 @@
...
+package rules
+
+import (
+	"github.com/golang/protobuf/proto"
+	"google.golang.org/genproto/googleapis/api/annotations"
+	p "google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// GetHTTPRules returns a array of HTTP rules for a given method descriptor.
```
alin04: getHTTPRules (i.e. lowercaseCamelCase) to match method name

rules/utils.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 Google LLC
```
alin04: utils.go is pretty generic, and could become a huge dumping ground. Maybe name this file http_annotations.go to keep all utilities around annotations co-located?

rules/0131.go
lukesneeringer: Probably not. Good call.

rules/utils.go
lukesneeringer: I have almost the opposite concern -- given that this is a `rules` package, the number of things that are not rules should be as limited as possible.

rules/utils.go
lukesneeringer:
```suggestion
// getHTTPRules returns a array of HTTP rules for a given method descriptor.
```

rules/0131.go
lukesneeringer: This one definitely should be must.

```suggestion
							Message:    "Get RPCs must use the GET HTTP verb.",
```

rules/utils.go
mingzhi: Maybe you want to create an internal package for util: `internal/util` under `rules`.

lukesneeringer: That makes me unable to read the `testdata`.

rules/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+		fdp := testutil.MustCreateFileDescriptorProto(
+			t,
+			testutil.FileDescriptorSpec{
+				AdditionalProtoPaths: []string{testdatadir("api-common-protos")},
+				Filename:             "test.proto",
+				Template:             tmpl,
+				Data:                 test,
+			},
+		)
+		reg, err := lint.MakeRegistryFromAllFiles([]*descriptorpb.FileDescriptorProto{fdp})
```
robcapo: We shouldn't expose this method just for the sake of this test. This functionality doesn't really even belong in the linter – I wrote it as a temporary helper until the proto team adds this feature to protodesc. Try using [protoregistry.NewFiles](https://github.com/golang/protobuf/blob/api-v2/reflect/protoregistry/registry.go#L76) directly.

Maybe you can make a new `testutil` method that takes a directory path and outputs a `[]*descriptorpb.FileDescriptorProto`. Or maybe there's some way to get the MethodDescriptor from the descriptorpb version of it.

lukesneeringer: `protoregistry.NewFiles` takes a `protoreflect.FileDescriptor`, which is not the thing I have.

robcapo: Yeah, the coordination between the two packages can be a bit annoying, but I still feel that this function should not be exposed from our package. We intend on removing it. I would almost prefer you copy paste it as a private function into this package if you can't find a cleaner way to test your helper.

lukesneeringer:
Replaced by #188.

PR #154, opened 2019-08-28
Migrate to third-party protoreflect.

lukesneeringer:
This PR migrates the linter to the third-party protoreflect package, as well as migrates existing rules.
Additionally, it moves rules into their own packages and changes up their organization a little bit.

Worth noting that this also removes the requirement for a git submodule, and in fact removes the dependency on having protoc at all (although it can still run as a protoc plugin, of course).


rules/aip0131/aip0131_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	"github.com/googleapis/api-linter/lint"
+)
+
+func TestAddRules(t *testing.T) {
+	rules := make(lint.Rules)
+	AddRules(rules)
+	for ruleName := range rules {
+		if !strings.HasPrefix(string(ruleName), "core::0131") {
```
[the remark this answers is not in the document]
mingzhi: Agree! Thanks!

cmd/api-linter/cli.go
```diff
@@ -84,12 +86,15 @@ func runCLI(rules lint.Rules, configs lint.Configs, args []string) error {
 					configs = append(configs, userConfigs...)
 				}
 
+				// Instantitate the linter object, and lint the protos.
```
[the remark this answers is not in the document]
lukesneeringer:
```suggestion
				// Instantiate the linter object, and lint the protos.
```

cmd/api-linter/summary_cli.go
```diff
@@ -62,6 +62,7 @@ func createSummary(responses []lint.Response) (summary LintSummary) {
 	return
 }
 
+// LintSummary summarized a lint run, including which files have violations.
```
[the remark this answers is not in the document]
lukesneeringer:
```suggestion
// LintSummary summarizes a lint run, including which files have violations.
```

lint/rule.go
```diff
@@ -14,10 +14,143 @@
...
+	LintField func(*desc.FieldDescriptor) []Problem
+
+	// LintService is called for each service in the file.
+	LintService func(*desc.ServiceDescriptor) []Problem
+
+	// LintMethod is called for each method in the file.
+	LintMethod func(*desc.MethodDescriptor) []Problem
+
+	// LintEnum is called for each enum in the file.
+	LintEnum func(*desc.EnumDescriptor) []Problem
```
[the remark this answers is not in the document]
lukesneeringer: YAGNI. We can always add it later.

lint/rule.go
```diff
@@ -14,10 +14,143 @@
...
+		}
+	}
+
+	// The rule is enabled.
+	return true
+}
+
+// getAllMessages returns a slice with every message (not just top-level
+// messages) in the file.
+func getAllMessages(f *desc.FileDescriptor) (messages []*desc.MessageDescriptor) {
```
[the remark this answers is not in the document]
lukesneeringer: I would rather not. Many other `desc.Descriptor` objects lack a `GetMessages` method.

lint/rule.go
```diff
@@ -14,10 +14,143 @@
...
+	for _, message := range f.GetMessageTypes() {
+		messages = append(messages, message)
```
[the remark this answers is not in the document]
lukesneeringer: Done.

rules/aip0131/message_rules.go
```diff
@@ -0,0 +1,91 @@
...
+		// We only care about Get methods for the purpose of this rule;
+		// ignore everything else.
+		if !isGetRequestMessage(m) {
+			return problems
+		}
+
+		// Rule check: Establish that a name field is present.
+		name := m.FindFieldByName("name")
+		if name == nil {
+			problems = append(problems, lint.Problem{
```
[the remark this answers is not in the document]
lukesneeringer: Done.

rules/aip0131/message_rules.go
```diff
@@ -0,0 +1,91 @@
...
+// The Get standard method should only have expected fields.
+var standardFields = lint.Rule{
+	Name:        lint.NewRuleName("core", "0131", "request-message", "name-field"),
+	Description: "The Get standard method must include expected fields.",
+	URI:         "https://aip.dev/131#request-message",
+	LintMessage: func(m *desc.MessageDescriptor) (problems []lint.Problem) {
+		// We only care about Get methods for the purpose of this rule;
+		// ignore everything else.
+		if !isGetRequestMessage(m) {
+			return problems
```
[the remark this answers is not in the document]
lukesneeringer: Done

rules/aip0131/message_rules.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/builder"
+)
+
+// The Get standard method should only have expected fields.
+var standardFields = lint.Rule{
+	Name:        lint.NewRuleName("core", "0131", "request-message", "name-field"),
+	Description: "The Get standard method must include expected fields.",
+	URI:         "https://aip.dev/131#request-message",
+	LintMessage: func(m *desc.MessageDescriptor) (problems []lint.Problem) {
```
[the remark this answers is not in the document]
lukesneeringer: Done except in the unknown fields rules, where we do aggregate problems and the existing style is much clearer.

rules/aip0191/syntax.go
```diff
@@ -0,0 +1,36 @@
...
+var syntax = lint.Rule{
+	Name:        lint.NewRuleName("core", "0191", "proto-version"),
+	Description: "APIs must use proto3.",
+	URI:         "https://aip.dev/191#guidance",
+	LintFile: func(f *desc.FileDescriptor) (problems []lint.Problem) {
+		if !f.IsProto3() {
+			problems = append(problems, lint.Problem{
+				Message:    "All API proto files must use proto3 syntax.",
+				Suggestion: "syntax = \"proto3\"",
+				Descriptor: f,
```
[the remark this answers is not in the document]
lukesneeringer: Fine.

lint/rule.go
```diff
@@ -14,10 +14,143 @@
...
+		}
+	}
+
+	// Finally, process rules for the file itself.
+	if rule.LintFile != nil {
+		problems = append(problems, rule.LintFile(f)...)
+	}
+
+	// Done aggregating problems for this file; return the list.
+	return problems
```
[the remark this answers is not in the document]
lukesneeringer: `return problems` is clearer.

rules/aip0132/method_rules.go
```diff
@@ -0,0 +1,80 @@
...
+	LintMethod: func(m *desc.MethodDescriptor) (problems []lint.Problem) {
+		// We only care about List- methods for the purpose of this rule;
+		// ignore everything else.
+		if !isListMethod(m) {
+			return
+		}
+
+		// Rule check: Establish that for methods such as `ListFoos`, the request
+		// message is named `ListFoosRequest`.
+		if got, want := m.GetInputType().GetName(), m.GetName()+"Request"; got != want {
```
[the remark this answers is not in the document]
lukesneeringer: Maybe? It is not really clear to me that a utility method is an improvement; this is a one line comparison.

lint/linter.go
```diff
@@ -76,27 +57,30 @@ func (l *Linter) LintProtosWithRegistry(files []*descriptorpb.FileDescriptorProt
 // It uses the proto file path to determine which rules will
 // be applied to the request, according to the list of Linter
 // configs.
-func (l *Linter) run(req Request) (Response, error) {
+func (l *Linter) run(fd *desc.FileDescriptor) (Response, error) {
```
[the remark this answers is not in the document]
lukesneeringer: I went with `lintFileDescriptor`.

lint/linter_test.go
```diff
@@ -88,62 +92,73 @@ func TestLinter_run(t *testing.T) {
...
 		l := New(rules, test.configs)
 
-		resp, _ := l.run(req)
-		if !reflect.DeepEqual(resp, test.resp) {
-			t.Errorf("Test #%d (%s): Linter.run()=%v; want %v", ind+1, test.desc, resp, test.resp)
+		// Actually run the linter.
+		resp, _ := l.run(fd)
+
+		// Assert that we got the problems we expected.
+		if !reflect.DeepEqual(resp.Problems, test.problems) {
```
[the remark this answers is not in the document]
lukesneeringer: This turned out not to be feasible for this particular test (it requires a circular import), but I will do this elsewhere.

rules/aip0132/method_rules.go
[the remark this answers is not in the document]
alin04: It's not one line. It's the entire block that is duplicated. Lines 38-47 here are identical to aip0131/method_rules.go. And this check will likely occur again for Update/Delete and custom methods. 

DRY

lint/rule.go
```diff
@@ -14,10 +14,172 @@
...
+
+// fileHeader attempts to get the comment at the top of the file, but it
+// is on a best effort basis because protobuf is inconsistent.
+//
+// Taken from https://github.com/jhump/protoreflect/issues/215
+func fileHeader(fd *desc.FileDescriptor) string {
+	var firstLoc *descriptor.SourceCodeInfo_Location
+	var firstSpan int64
+	for _, curr := range fd.AsFileDescriptorProto().GetSourceCodeInfo().GetLocation() {
+		if curr.LeadingComments == nil && len(curr.LeadingDetachedComments) == 0 {
```
mingzhi: I think it will make some mistakes in the following case:
```protobuf
syntax "proto3";

// This is not the file header, but it will be picked up by this function 
// and the following rule will be disabled for all messages in the file.
// api-linter: check-message-name = disabled
message M {}
```

So please add some tests for this function.

lint/rule.go
```diff
@@ -14,10 +14,172 @@
...
+	"github.com/jhump/protoreflect/desc"
+)
+
 // Rule defines a lint rule that checks Google Protobuf APIs.
-type Rule interface {
-	// Info returns metadata about a rule.
-	Info() RuleInfo
-	// Lint performs the linting process.
-	Lint(Request) ([]Problem, error)
+type Rule struct {
```
robcapo: This is a very major change. We have 20 rules in google3 that this will break. Can you propose a new struct with a different name that conforms to the Rule interface? It should be relatively easy to do and prevent the need for a bunch of rework.

```go
type LukesRule struct {
	// Name contains the rule's name.
	Info() RuleInfo		Name RuleName
	// Lint performs the linting process.	
 	Lint(Request) ([]Problem, error)		// Description is a short description of the rule,
	// used in documentation and elsewhere.
	Description string

 	// URI is the address where the guideline is documented.
	// This should be displayed to give API designers more information about
	// "how to do this right".
	URI string

 	// LintFile is called for files as a whole.
	LintFile func(*desc.FileDescriptor) []Problem

 	// LintMessage is called for any messages in the file.
	// This includes nested messages, regardless of depth.
	LintMessage func(*desc.MessageDescriptor) []Problem

 	// LintField is called for any fields in the file.
	// This includes fields within nested messages, regardless of depth.
	LintField func(*desc.FieldDescriptor) []Problem

 	// LintService is called for each service in the file.
	LintService func(*desc.ServiceDescriptor) []Problem

 	// LintMethod is called for each method in the file.
	LintMethod func(*desc.MethodDescriptor) []Problem

 	// LintEnum is called for each enum in the file.
	LintEnum func(*desc.EnumDescriptor) []Problem
}

func (l LukesRule) Info() RuleInfo {
  return RuleInfo{
    Name: l.RuleName,
    Description: l.Description,
    URI: l.URI,
  }
}

func (l LukesRule) Lint(r Request) ([]Problem, error) {
  // logic for traversing callbacks.
}
```

otherwise syncing will be blocked until someone refactors all the rules in google3

lukesneeringer: Syncing will be blocked anyway, considering the other changes.

I am willing to refactor the rules in google3 (we just need to stop making more of them). It was not difficult to refactor the ones here.

lint/rule.go
```diff
@@ -14,10 +14,172 @@
...
+
+	// LintService is called for each service in the file.
+	LintService func(*desc.ServiceDescriptor) []Problem
+
+	// LintMethod is called for each method in the file.
+	LintMethod func(*desc.MethodDescriptor) []Problem
+
+	// LintEnum is called for each enum in the file.
+	LintEnum func(*desc.EnumDescriptor) []Problem
+}
```
robcapo: Have you considered using interfaces instead of a big struct?

```go
type FileLinter interface {
  func LintFile(*desc.FileDescriptor) []Problem
}

type MessageLinter interface {
  func LintMessage(*desc.MessageDescriptor) []Problem
}
```

There would be very little burden on the rule implementer (i.e. just add the funcs that you want to implement), and go's [type switch](https://tour.golang.org/methods/16) makes this implementation really easy and type-safe for the linter core. It's also a bit more idiomatic in Go, and will enable users to implement our desired behavior with their own data structures if needed. To me, this struct looks more like a convenience helper than a library interface, and can be made available as such (i.e. in the linter core we don't actually care about the data structure used to hold state and lint descriptors, but rather the behavior, which is idiomatically described with interfaces).

If you do choose to go this route, I would suggest adding a `noPositional struct{}` argument here to prevent users from initializing this struct with positional arguments.

lint/rule.go
```diff
@@ -14,10 +14,172 @@
...
+	// used in documentation and elsewhere.
+	Description string
+
+	// URI is the address where the guideline is documented.
+	// This should be displayed to give API designers more information about
+	// "how to do this right".
+	URI string
+
+	// LintFile is called for files as a whole.
+	LintFile func(*desc.FileDescriptor) []Problem
```
robcapo: Why the removal of `error`? The inability for a rule to express that they failed to lint a proto seems like a limitation. Tricorder also has hooks to report linting failures, which are useful for debugging.

lint/rule.go
lukesneeringer: I would be willing to change this to use interfaces in the way that you suggest.

lint/rule.go
lukesneeringer: @robcapo and I discussed this in email, and decided that I will take responsibility for migrating their existing rules (and that they will not create new ones until this can go in).

lint/rule.go
lukesneeringer: Going down this road, this is less clear to me. It might be lack of knowledge on my part, though.

I feel like this is going to end in something like...

```go
type RuleInterface interface {
  Name string
  Description string
  URI string
  Lint func(m *desc.FileDescriptor) []Problem
}

type MessageRuleInterface {
  LintMessage func(m *desc.MessageDescriptor) []Problem
}
```

And then I still need the concrete variant...

```go
struct MessageRule {
  Name string
  Description string
  URI string
  LintMessage func(m *desc.MessageDescriptor) []Problem
}
```

And then the user instantiates a `MessageRule` rather than a `Rule`, otherwise identically. That does not seem like an improvement, it just seems like extra code.

An alternative variant might be to _only_ define the interfaces and then the user defines their own struct...

```go
// Current way...
myRule = Rule{
  Name: "..."
  Description: "..."
  URI: "..."
  LintMessage: func(m *desc.MessageDescriptor) []Problem {
    // implementation here...
  }
}

// New way...
myRule = struct{
  Name string
  Description string
  URI string
  LintMessage: func(m *desc.MessageDescriptor) []Problem
}{
  Name: "..."
  Description: "..."
  URI: "..."
  LintMessage: func(m *desc.MessageDescriptor) []Problem {
    // implementation here...
  }
}
```

...I assume that can not be what you mean, because that seems even more terrible.

What am I missing? :)

lint/rule.go
```diff
@@ -14,10 +14,143 @@
...
+		problems = append(problems, rule.LintFile(f)...)
+	}
+
+	// Done aggregating problems for this file; return the list.
+	return problems
+}
+
+// IsEnabled returns true if the rule is enabled (not disabled by the comments
+// for the given descriptor or its file), false otherwise.
+func (rule *Rule) IsEnabled(d desc.Descriptor) bool {
```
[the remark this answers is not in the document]
lukesneeringer: Done.

lint/rule.go
robcapo: Sorry, I should have sent a more detailed explanation. Basically what I was thinking was a main `rule` interface that would be embedded in each of the other more specific interfaces, and therefore you could have something like:
```go
type rule interface {
  Info() RuleInfo
}
type fieldLinter interface {
  rule
  LintField(d desc.FieldDescriptor) []Problem
}
type messageLinter interface { ... }
// ...
```
and rules like
```go
type myRule struct{}
func (r myRule) Info() RuleInfo {
  return RuleInfo{
    Name: "...",
    Description: "..."
    URI: "..."
  }
}
func (r myRule) LintField(m *desc.FieldDescriptor) []Problem {
  // implementation here...
}
```
and each rule could choose to implement any number of methods like `LintField` `LintMessage` etc. I'm not sure if it solves the same problem you're after here. It's still trivial to add a helper struct like the one you mentioned above, but by keeping the interfaces as the boundary, people can easily implement their own structures if they need to do things like persist state across calls to LintXxx, and if we want to add a new helper we can easily do so without breaking rules.
 
It's really just an option to consider, because you're making some pretty significant design changes as part of this PR. What I'd really like to see is a drop-in of the new protoreflect library for the old one. It's difficult to judge the impact that the change of library has with the impact of the larger structural changes in the repo, and I think these changes are distracting from the more important library swap. I think I understand why you did it like this (since you're breaking the interface regardless), but I think two smaller scale changes would be easier to review than one larger change. The library swap is pretty cut and dry, but the structural changes have some more important design considerations.

By the way, I started a [go-nuts discussion](http://g/go-nuts/gjPmImTdf94) a while ago about this topic. You might be interested in some of that discussion.

lukesneeringer: > and each rule could choose to implement any number of methods like LintField LintMessage etc.

I am trying to cut down on the amount of boilerplate in rule-writing and rule-reading, and this re-introduces a significant amount of it. I fail to see why having an `Info()` method as a level of indirection is valuable, for example. When I asked @alin04 about it, she felt similarly.

I guess if I want to do this, I want to understand why it is _better_, because it really just feels like a more complicated way to accomplish the same thing.

> I think I understand why you did it like this (since you're breaking the interface regardless), but I think two smaller scale changes would be easier to review than one larger change. The library swap is pretty cut and dry, but the structural changes have some more important design considerations.

Right, that requires two separate migrations before we can start writing rules. I am trying to address any interface-breaking concerns so we can "get this show on the road", so to speak.

I _can_ break this up into two PRs, but it would be a significant amount of work, and I am disinclined unless we need to address a design concern.

lukesneeringer: I talked to @noahdietz and he has somewhat convinced me.

I think implementing interfaces here is reasonable, but it is _really_ important to me that we do not have a lot of extra boilerplate code when defining most rules. I think I can make that happen though.

robcapo: `RuleInfo` is just a way to store structured data that doesn't have any behavior attached to it. It's essentially the same thing as a proto `message` – a declaration of data that we expect to be associated with rules. `Rule` is closer to a `service` definition – a declaration of behavior we expect without any ties to implementation. Adding `Info()` to the rule interface enables us to make sure that every rule implementation provides us with the data we care about. A side effect is that you can easily reuse the same set of info for many rules. For example, you can use a factory method or shared global struct across multiple rules and keep their shared info in one place (e.g. for an AIP that has a bunch of rules). The benefit of using an interface for rules is that it doesn't bind us to any implementation. That's why you can easily create the code you have here as a helper and satisfy the interface with it, and if I want to implement a rule using the code snippet in my previous comment, I can do that as well. Everyone's happy.

By my count, it's 7 lines of boilerplate versus 4. Is the enforced reduction of 3 lines of boilerplate enough to justify binding rule developers to a specific implementation and restructuring a bunch of existing code? I think the primary point that i'm trying (maybe poorly) to communicate here is that _interfaces are cheap_. Especially if they're unexported like the example I showed above. You can delete, rename, or restructure them as needed, because they're package private and no one depends on them but the code you already own. It's basically as cheap as python's duck typing, but you get free compile time checks with it.

Because you've already offered to refactor all the google3 code that's tied to the existing interface, I won't block this. And of course I'll be happy to help with any questions you have when you do the first sync.

robcapo: Just noticed you're other message. Yes, I think you're rightfully concerned about boilerplate. Helper libraries / method can help a lot here. Glad @noahdietz is more convincing than me :)

rules/aip0132/method_rules.go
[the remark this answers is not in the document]
lukesneeringer: I went to add this and I am struggling to come up with a utility method that feels worth the level of indirection. I wonder if you have something different in mind.

I can write something like...

```go
func checkRequestName(m *desc.MethodDescriptor) []Problem {
  	if got, want := m.GetInputType().GetName(), m.GetName()+"Request"; got != want {
		return []lint.Problem{{
			Message: fmt.Sprintf(
				"%q should have a request message named after the RPC, such as %q.",
				m.GetName(),
				want,
			),
			Suggestion: want,
			Descriptor: m,
		}}
	}
	return nil
}
```

I agree that cuts down on repetition somewhat, and I could factor it out a bit more to handle request and response messages, but I would have to put them in a utility functions package (which I kind of dislike) for them to be useful across the standard methods. I am conflicted.

lint/lint_test.go
```diff
@@ -0,0 +1,168 @@
...
+		{&FileRule{
+			Name: "panic-error",
+			LintFile: func(_ *desc.FileDescriptor) []Problem {
+				panic(fmt.Errorf("panic"))
+			},
+		}},
+	}
+
+	for _, test := range tests {
+		r, err := NewRuleRegistry(test.rule)
```
alin04: t.Run()?

rules/aip0132/method_rules.go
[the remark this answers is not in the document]
alin04: I can see the same DRY for checkResponseName and checkAllowedFields. Eventually a utility package might be worthwhile. I'm fine to add that later once it becomes more apparent there's a lot of cut&paste happening.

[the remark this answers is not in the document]
lukesneeringer: I think I would prefer add later if you will be okay with that. I think there is a decent chance you might be right, but I am not sure yet, and it is pretty orthogonal to this PR.

lint/lint_test.go
lukesneeringer: Fixed, thanks.

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/jhump/protoreflect/desc"
+)
+
 // Rule defines a lint rule that checks Google Protobuf APIs.
 type Rule interface {
```
mingzhi: Please consider that we will add rules for OpenAPIs -- it is better to change the name to `ProtoRule`.

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
-	// Info returns metadata about a rule.
-	Info() RuleInfo
-	// Lint performs the linting process.
-	Lint(Request) ([]Problem, error)
+	// GetName returns the name of the rule.
+	GetName() RuleName
+
+	// GetURI returns the URI where the applicable guideline
+	// is documented. (This should generally be an AIP on https://aip.dev/.)
+	GetURI() string
```
mingzhi: What is the purpose of this function? If you want to provide information about a rule in command lines, please note that a AIP (or a sub section) would contains multiple rules. For example, from the URI https://aip.dev/131#request-message, I am not certain if a rule is about "A resource name field must be included" or "It should be called name".

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+
+	// Lint accepts a FileDescriptor and lints it,
+	// returning a slice of Problem objects it finds.
+	Lint(*desc.FileDescriptor) []Problem
+}
+
+// FileRule defines a lint rule that checks a file as a whole.
+type FileRule struct {
+	Name RuleName
```
mingzhi: Consider using embedded struct to reduce boilerplate codes, for example,
```golang
type descRule struct {
  Name RuleName
  URI string
}

func (r *descRule) GetName() RuleName {
  return r.name
}

func (r *descRule) GetURI() string {
  return r.uri
}

type FileRule struct {
  descRule
  LintFile func(*desc.FileDescriptor) []Problem
}
// And now you only need to implement Lint function.
```

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+// GetURI returns the URI where the applicable guideline is documented.
+func (r *FieldRule) GetURI() string {
+	return r.URI
+}
+
+// Lint accepts a FileDescriptor and lints every field in the file.
+func (r *FieldRule) Lint(fd *desc.FileDescriptor) []Problem {
+	problems := []Problem{}
+
+	// Iterate over each message and process rules for each field in that
```
mingzhi: What about extensions?

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+	return r.Name
+}
+
+// GetURI returns the URI where the applicable guideline is documented.
+func (r *ServiceRule) GetURI() string {
+	return r.URI
+}
+
+// Lint accepts a FileDescriptor and lints every service in the file.
+func (r *ServiceRule) Lint(fd *desc.FileDescriptor) (problems []Problem) {
```
mingzhi: Please don't use named result parameters, which will stutter in godoc.

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+	return r.Name
+}
+
+// GetURI returns the URI where the applicable guideline is documented.
+func (r *MethodRule) GetURI() string {
+	return r.URI
+}
+
+// Lint accepts a FileDescriptor and lints every method in the file.
+func (r *MethodRule) Lint(fd *desc.FileDescriptor) (problems []Problem) {
```
mingzhi: Please don't use named result parameters, which will stutter in godoc.

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+}
+
+// GetURI returns the URI where the applicable guideline is documented.
+func (r *EnumRule) GetURI() string {
+	return r.URI
+}
+
+// Lint accepts a FileDescriptor and lints every enum in the file
+// (including enums nested within messages).
+func (r *EnumRule) Lint(fd *desc.FileDescriptor) (problems []Problem) {
```
mingzhi: Please don't use named result parameters, which will stutter in godoc.

lint/rule.go
mingzhi: See jhump's comment in the original issue.

lint/rule.go
```diff
@@ -14,10 +14,295 @@
...
+)
+
 // Rule defines a lint rule that checks Google Protobuf APIs.
 type Rule interface {
-	// Info returns metadata about a rule.
-	Info() RuleInfo
-	// Lint performs the linting process.
-	Lint(Request) ([]Problem, error)
+	// GetName returns the name of the rule.
+	GetName() RuleName
```
mingzhi: Please follow Go idioms: https://golang.org/doc/effective_go.html#Getters -- use `Name()` instead.

lint/rule.go
lukesneeringer: As best as I can tell, an embedded struct _increases_ boilerplate code because every single rule has to know about the level of indirection.

lint/rule.go
lukesneeringer: What about them?

lint/rule.go
lukesneeringer: Changed to non-named parameters, albeit for different reasons (`reflect.DeepEqual` barfs when we return `nil`).

lint/rule.go
lukesneeringer: Switched.

lint/rule.go
lukesneeringer: Since rule authors will be interacting with the protoreflect library that uses `GetFoo`, local consistency is more important.

lint/rule.go
mingzhi: Will an extension be considered as a field?

lint/rule.go
lukesneeringer: Internally, we should be including the URI at the end of the message we post in Critique.
If we can not do that, we should just start putting this in the problem messages.

lint/rule.go
mingzhi: I don't think the following codes will increase any boilerplate code for a rule:
```golang
f := FileRule{}
f.Name = "checkFieldName"
f.URI = "http://some-where"
f.LintFile = func(d *desc.FileDescriptor) []Problem {}
```
But you may just want:
```golang
f := FileRule{
Name: "checkFieldName"
URI: "http://some-where"
LintFile: func(d *desc.FileDescriptor) []Problem {}
}
```
which an embedded struct will be an issue.

lint/rule.go
lukesneeringer: I doubt that ever happens in real life, but done regardless.

lint/rule.go
mingzhi: protoreflect is a special case due to their decision to maintain consistence with protobuf. But in our godoc, we will not see any `GetFoo` from the protoreflect library. So I think we should keep Go idioms.

lint/rule.go
mingzhi: I think we should put the URI into the problem messages, as the URI does not serve as an identity, but more like "for more context/details, see this link".

mingzhi: Should we give some description about a rule, i.e., `Description() string`, so a user can understand a rule before running and getting the problem messages (which may be tricky to trigger one before understanding the rule)?

lint/rule_registry.go
```diff
@@ -18,12 +18,12 @@ import (
 	"fmt"
 )
 
-// Rules is a registry for registering and looking up rules.
-type Rules map[RuleName]Rule
+// RuleRegistry is a registry for registering and looking up rules.
+type RuleRegistry map[RuleName]ProtoRule
```
mingzhi: Consider change it to `ProtoRuleRegistry` as well or `Register` to `RegisterProtoRule`.

mingzhi: I think `Register` to `RegisterProtoRules` is better, considering how you use `RuleRegistry` in the `AddRules` function in each rule package.

lint/rule_registry.go
```diff
@@ -32,33 +32,33 @@ func (r Rules) Copy() Rules {
...
-		r[rl.Info().Name] = rl
+		r[rl.GetName()] = rl
 	}
 	return nil
 }
 
 // All returns all rules.
-func (r Rules) All() []Rule {
-	rules := make([]Rule, 0, len(r))
+func (r RuleRegistry) All() []ProtoRule {
```
mingzhi: Consider change `All` to `AllProtoRules` as well, if you stick with `RuleRegistry`.

lint/rule.go
lukesneeringer: No. Among many other reasons, it would be unlikely to be in the same file.

lint/rule.go
lukesneeringer: Yes, I want the latter, and embedded structs break it. I would rather have more code in the rule declaration file to hold rule authors harmless.

lint/rule.go
lukesneeringer: But rule authors are likely to use `GetMessageTypes`, `GetFields`, etc., and should not have to think about whether this particular item they are using has a `Get-` prefix.

Also, it would require `MessageRule`, `MethodRule`, etc., to adopt different field names in their structs.

Therefore, this should stay as-is.

lint/rule.go
lukesneeringer: URI is much more valuable than description (which can just be a comment above the rule). If we should not have URI, we _definitely_ should not have description.

lint/rule.go
lukesneeringer: Updated, and test added.

lint/rule.go
mingzhi: Maybe we should start to discuss about having rule numbers in an AIP, e.g., AIP-131-R1 for the rule that checks that "A resource name field must be included", and in the AIP text, a reference to the rule. We may want to remove RuleName then, since sometimes it is really hard to give a unique, informative name.

lint/rule.go
mingzhi: Okay, conflict of field and function names is a legit excuse.

cmd/api-linter/cli.go
```diff
@@ -63,19 +59,25 @@ func runCLI(rules lint.Rules, configs lint.Configs, args []string) error {
 			},
 			Action: func(c *cli.Context) error {
 				filenames := c.Args()
+
+				// Sanity check: Were we given any files to parse at all?
+				// If not, abort.
```
mingzhi: Add a TODO to inform the user to include files, or better that we check all proto files in the current directory.

cmd/api-linter/integration_test.go
```diff
@@ -23,27 +23,31 @@ import (
 	"testing"
 )
 
-var protocPath = func() string {
-	return "protoc"
-}
-
 // Each rule must have a testing case stored here.
```
mingzhi: Remove this line of comment, as we don't require that each rule must have a testing case here. Thanks!

cmd/api-linter/integration_test.go
```diff
@@ -94,12 +100,14 @@ func TestRules_DisabledByFileComments(t *testing.T) {
...
 
 	for _, test := range testCases {
-		disableInFile := fmt.Sprintf("// (-- api-linter: %s=disabled --)\n", test.rule)
-		proto := disableInFile + "\n" + test.proto
-		result := runLinter(t, proto, config)
-		if strings.Contains(result, test.rule) {
-			t.Errorf("Rule %q should be disabled by file comments", test.rule)
-		}
+		t.Run(test.testName, func(t *testing.T) {
+			disableInFile := fmt.Sprintf("// (-- api-linter: %s=disabled --)", test.rule)
```
mingzhi: Remove "(--" and "--)".

cmd/api-linter/integration_test.go
```diff
@@ -119,12 +127,14 @@ func TestRules_DisabledByInlineComments(t *testing.T) {
...
 
 	for _, test := range testCases {
-		disableInline := fmt.Sprintf("// (-- api-linter: %s=disabled --)\n", test.rule)
-		proto := strings.Replace(test.proto, "disable-me-here", disableInline, -1)
-		result := runLinter(t, proto, config)
-		if strings.Contains(result, test.rule) {
-			t.Errorf("Rule %q should be disabled by in-line comments", test.rule)
-		}
+		t.Run(test.testName, func(t *testing.T) {
+			disableInline := fmt.Sprintf("(-- api-linter: %s=disabled --)", test.rule)
```
mingzhi: Remove "(--" and "--)".

cmd/api-linter/integration_test.go
lukesneeringer: The common use case will include these tokens.

cmd/api-linter/integration_test.go
lukesneeringer: The common use case will include these tokens.

cmd/api-linter/integration_test.go
lukesneeringer: Done.

cmd/api-linter/integration_test.go
```diff
@@ -165,11 +177,14 @@ func runLinter(t *testing.T, proto, config string) string {
 	if err != nil {
 		t.Fatal(err)
 	}
+	if err := os.Chdir(workdir); err != nil {
```
mingzhi: I am not very sure if changing directory would affect other tests.

cmd/api-linter/cli.go
lukesneeringer: Added an error message.

cmd/api-linter/integration_test.go
```diff
@@ -184,7 +199,6 @@ func runLinter(t *testing.T, proto, config string) string {
 		"--cfg=" + configPath,
 		"--out=" + outPath,
 		"--proto_path=" + workdir,
```
mingzhi: If we change the current directory to workdir, I don't think we need to point proto_path to it again.

lint/rule.go
lukesneeringer: > Maybe we should start to discuss about having rule numbers in an AIP, e.g., AIP-131-R1 for the rule that checks that "A resource name field must be included"

This is what I originally asked for.

That said, at this point, I would like to get this PR in, as it is going to keep getting larger. If we want to refactor rule names, let's do a separate followup.

lint/lint.go
```diff
@@ -76,27 +57,30 @@ func (l *Linter) LintProtosWithRegistry(files []*descriptorpb.FileDescriptorProt
 // It uses the proto file path to determine which rules will
 // be applied to the request, according to the list of Linter
 // configs.
-func (l *Linter) run(req Request) (Response, error) {
+func (l *Linter) lintFileDescriptor(fd *desc.FileDescriptor) (Response, error) {
 	resp := Response{
-		FilePath: req.ProtoFile().Path(),
+		FilePath: fd.GetName(),
+		Problems: []Problem{},
```
mingzhi: I don't think we need to initialize problem array here -- better to remove this line.

lint/lint_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/jhump/protoreflect/desc/builder"
+)
+
+func TestLinter_run(t *testing.T) {
+	fd, err := builder.NewFile("protofile.proto").Build()
+	if err != nil {
+		t.Fatalf("Failed to build a file descriptor.")
+	}
+	defaultConfigs := Configs{
+		{[]string{"**"}, []string{}, map[string]RuleConfig{}},
```
mingzhi: I think it is more informative to write down the field name, e.g., `IncludedPaths: []string{"**"}`, which also prevents issues when we refactor the `Config` struct, e.g., by inserting a new field.

lint/lint_test.go
```diff
@@ -0,0 +1,177 @@
...
+					return test.problems
+				},
+			})
+			if err != nil {
+				t.Fatal(err)
+			}
+			l := New(rules, test.configs)
+
+			// Actually run the linter.
+			resp, _ := l.lintFileDescriptor(fd)
```
mingzhi: I think it is better to test the export function: `LintProtos`.

lint/rule.go
mingzhi: Totally agree! Thanks!

lint/rule.go
```diff
@@ -14,10 +14,304 @@
...
+			problems = append(problems, r.LintEnum(enum)...)
+		}
+	}
+	return problems
+}
+
+// ruleIsEnabled returns true if the rule is enabled (not disabled by the comments
+// for the given descriptor or its file), false otherwise.
+func ruleIsEnabled(rule ProtoRule, d desc.Descriptor) bool {
+	directive := fmt.Sprintf("api-linter: %s=disabled", rule.GetName())
```
mingzhi: If a user wants to disable two rules, he/she needs to write
```
api-linter: a=disabled
api-linter: b=disabled
```
but not
```
api-linter: a,b=disable
```
right?

mingzhi: In users' perspective, I prefer the second way. But we can discuss more and if needed, make the changes in another PR. Thanks!

lint/rule.go
```diff
@@ -14,10 +14,143 @@
...
+		messages = append(messages, getAllNestedMessages(message)...)
+	}
+	return messages
+}
+
+// getAllNestedMessages returns a slice with the given message descriptor as well
+// as all nested message descriptors, traversing to arbitrary depth.
+func getAllNestedMessages(m *desc.MessageDescriptor) (messages []*desc.MessageDescriptor) {
+	for _, nested := range m.GetNestedMessageTypes() {
+		messages = append(messages, nested)
```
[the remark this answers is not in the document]
mingzhi: Nitpick -- I am Okay either way.

lint/rule.go
```diff
@@ -14,10 +14,304 @@
...
+func fileHeader(fd *desc.FileDescriptor) string {
+	var firstLoc *descriptor.SourceCodeInfo_Location
+	var firstSpan int64
+	for _, curr := range fd.AsFileDescriptorProto().GetSourceCodeInfo().GetLocation() {
+		// Skip locations that have no comments.
+		if curr.LeadingComments == nil && len(curr.LeadingDetachedComments) == 0 {
+			continue
+		}
+		// Skip locations that are not syntax, package, option, or import.
+		allowedPaths := map[int32]struct{}{2: {}, 3: {}, 8: {}, 12: {}}
```
mingzhi: Bring it out of the for loop, we don't need to create it in every loop.

lint/rule_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/builder"
+)
+
+func TestFileRule(t *testing.T) {
+	// Create a file descriptor with nothing in it.
+	fd, err := builder.NewFile("test.proto").Build()
+	if err != nil {
+		t.Fatalf("Could not build file descriptor.")
```
mingzhi: Print the `err` as well.

lint/rule_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+func TestMessageRule(t *testing.T) {
+	// Create a file descriptor with two messages in it.
+	fd, err := builder.NewFile("test.proto").AddMessage(
+		builder.NewMessage("Book"),
+	).AddMessage(
+		builder.NewMessage("Author"),
+	).Build()
+	if err != nil {
+		t.Fatalf("Failed to build file descriptor.")
```
mingzhi: Print the `err` as well.

lint/lint_test.go
lukesneeringer: This code is not actually changed (it is only in the diff because I renamed the file).

lint/rule.go
lukesneeringer: This mirrors the previous behavior, and can be done later.

lint/rule_test.go
lukesneeringer: Done, but this really does not matter as there is no user input or other variance, and this will never actually error.

lint/rule_test.go
lukesneeringer: Done, but this really does not matter as there is no user input or other variance, and this will never actually error.

lint/rule.go
lukesneeringer: Done.

lint/lint_test.go
mingzhi: Got it -- it is a nit pick -- we can refactor it later. Thanks!

lint/lint.go
lukesneeringer: In general, I want to make sure to send down `[]Problem{}` and not `nil`, because it makes testing easier (both `reflect.DeepEqual` and `cmp.Diff` consider empty slice to be unequal to nil).

rules/aip0131/message_rules_test.go
```diff
@@ -0,0 +1,103 @@
...
+				builder.NewField(test.nameFieldName, test.nameFieldType),
+			).Build()
+			if err != nil {
+				t.Fatalf("Could not build %s message.", test.messageName)
+			}
+
+			// Run the lint rule, and establish that it returns the correct
+			// number of problems.
+			if problems := standardFields.LintMessage(message); len(problems) != test.problemCount {
+				t.Errorf("%s on rule %s: %#v", test.errPrefix, standardFields.Name, problems)
```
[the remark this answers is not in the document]
alin04: This should be fixed. Merely checking the count isn't sufficient.

rules/aip0131/message_rules_test.go
```diff
@@ -0,0 +1,103 @@
...
+			).AddField(
+				builder.NewField(test.fieldName, test.fieldType),
+			).Build()
+			if err != nil {
+				t.Fatalf("Could not build GetBookRequest message.")
+			}
+
+			// Run the lint rule, and establish that it returns the correct
+			// number of problems.
+			if problems := unknownFields.LintMessage(message); len(problems) != test.problemCount {
```
alin04: Here too and other tests. Merely checking count isn't sufficient.

rules/aip0132/method_rules.go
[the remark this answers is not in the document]
alin04: sg

lukesneeringer:
Submitting this; will address @alin04's concern about testing in a follow-up PR very soon.

PR #159, opened 2019-09-17
Checked if rule name alias is disabled in comments.

mingzhi:


lint/rule.go
```diff
@@ -240,24 +240,46 @@ func (r *EnumRule) Lint(fd *desc.FileDescriptor) []Problem {
...
+		"core::0140::lower-snake": "naming-format",
+	}
+	return aliasMap[name]
+}
 
-	// If the comments above the descriptor disable the rule,
-	// return false.
+// ruleDisabledInComments returns true if the rule is disabled by the comments
+// for the given descriptor or its file.
+func ruleDisabledInComments(ruleName string, d desc.Descriptor) bool {
```
lukesneeringer: Instead of a separate function, just make a slice of strings `[]directive`, with the rule name and the alias, and `range` over it. It will be much simpler.

lint/rule_test.go
```diff
@@ -294,13 +293,16 @@ func TestRuleIsEnabled(t *testing.T) {
...
 		enabled        bool
 	}{
-		{"Enabled", "", "", true},
-		{"FileDisabled", "api-linter: test=disabled", "", false},
-		{"MessageDisabled", "", "api-linter: test=disabled", false},
+		{"Enabled", NewRuleName("test"), "", "", true},
+		{"FileDisabled", NewRuleName("test"), "api-linter: test=disabled", "", false},
+		{"MessageDisabled", NewRuleName("test"), "", "api-linter: test=disabled", false},
+		{"NameNotMatch", NewRuleName("test"), "", "api-linter: other=disabled", true},
+		{"AliasDisabled", NewRuleName("core", "0140", "lower-snake"), "", "api-linter: naming-format=disabled", false},
```
robcapo: I don't think this test should depend on the actual rule implementations that we have in the repo, because those will change (i.e. we will add / remove aliases as we fix the usages of them). Maybe inject the aliasMap to ruleIsEnabled and then make a dummy one for the test. I think there's only one non-test usage of ruleIsEnabled so it should be an easy fix.

robcapo: we can also create an aliases.go that holds a global alias map and then have [the linter usage](https://github.com/googleapis/api-linter/blob/master/lint/lint.go#L82) pass that one in. Then the aliases aren't buried in code.

mingzhi: Done. Thanks!

lint/rule.go
mingzhi: Done. Thanks!

lint/rule.go
```diff
@@ -239,21 +239,31 @@ func (r *EnumRule) Lint(fd *desc.FileDescriptor) []Problem {
...
+		names = append(names, alias)
+	}
 
-	// If the comments above the descriptor disable the rule,
-	// return false.
-	if sourceInfo := d.GetSourceInfo(); sourceInfo != nil {
-		if strings.Contains(sourceInfo.GetLeadingComments(), directive) {
-			return false
-		}
+	var directives []string
```
lukesneeringer:
```suggestion
	directives := []string{}
```

PR #163, opened 2019-09-17
Problem testing

lukesneeringer:
This PR attempts to address some of @alin04's concerns around testing.

It creates a `testutil` package with a `ProblemStub` and `ProblemStubs` object. These are basically subsets of problems that know how to compare themselves with `lint.Problem` objects. The intent is to reduce test boilerplate while testing against the contents of the problems rather than just how many problems come back.

I do not really love that `ProblemStub` and `ProblemStubs` have to be exported, but I see no other choice. (I can not use an `internal` directory because the AIP rule packages would not be able to import from it.)

Note: I only edited two tests for now; if @alin04 is happy with this approach, I will convert the remainder.


testutil/problem_stub.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// ProblemStub describes a subset of a Problem, and is used to verify
+// that problems look correct in tests.
+type ProblemStub struct {
+	Descriptor desc.Descriptor
```
robcapo: Problems don't always have a descriptor (at least not in the original design). The descriptor was actually just used as a convenience to specify the location. I think you should compare Location instead.

lukesneeringer: I have used `Descriptor` for literally every rule I have written. I expect it to be by far the common case.

robcapo: I don't know where the old PackageLocation and SyntaxLocation methods went, but those don't have descriptors (unless this new library has them). We've written corp rules that lint both, so they won't be able to use this utility.

lukesneeringer: I concur that location makes sense in those cases.

robcapo: You can probably still keep the SetDescriptor method, but it can then use the location to verify.

lukesneeringer: So my goal here is to optimize for the common case. I am happy to add location if it makes sense later, but I still intend to leave this as-is for now.

robcapo: Fair enough. And although it's not completely relevant to this PR, can you point me to how rules will actually return a problem that points to package or syntax location? It will be relevant when we try to sync this.

mingzhi: @robcapo for proto rules, returning the descriptor is required, as we use it to check if a rule is disabled or not by looking at its comments. We will need to find a way to enforce it. 

That said, rules involving syntax, package, imports or options are tricky, as they are not descriptors, but attributes belong to FileDescriptor. As a result, we need to 
1. create helpers to extract those location information from the original SourceCodeInfo, as we did before;
2. educate users that they can only disable those rules at the file comments.

mingzhi:
@lukesneeringer could you please point us to Angie's original comments? Thanks!

lukesneeringer:
https://github.com/googleapis/api-linter/pull/154#discussion_r318795908
https://github.com/googleapis/api-linter/pull/154#discussion_r318799345

testutil/problem_stub.go
robcapo: Until we address this we will be blocked from syncing. I believe we used to use file comments to determine if package or syntax rules were disabled.

testutil/problem_stub.go
```diff
@@ -0,0 +1,85 @@
...
+// problem's descriptor match.
+func (ps *ProblemStub) VerifyDescriptor(p lint.Problem, t *testing.T) {
+	if !reflect.DeepEqual(p.Descriptor, ps.Descriptor) {
+		t.Errorf("Expected descriptor %v; got %v", ps.Descriptor, p.Descriptor)
+	}
+}
+
+// VerifyMessage verifies that the stub's message is included in the
+// problem's message.
+func (ps *ProblemStub) VerifyMessage(p lint.Problem, t *testing.T) {
```
mingzhi: It looks like [change detectors](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). I don't think we want to verify the *exact* message, maybe just to check if the message contains some required information. For example, we more likely want to see if the problem returns the right suggestion.

testutil/problem_stub.go
mingzhi: If >1 rules need syntax/package/options location, we will create a common help function, otherwise, I prefer that the rule itself deals with it, e.g., creating a `getSyntaxLocation(f FileDescriptor) *descriptor.SourceCodeInfo_Location`. I think it is not an issue, as @lukesneeringer already knows how to extract the original SourceCodeInfo, see [here](https://github.com/googleapis/api-linter/blob/9693cceee2f082cd894be60f10a88849e5ed3238/lint/rule.go#L287), but I am happy to help.

lukesneeringer: I think @robcapo is right here. I originally changed `protoRule` to say (in the comments) "always use descriptor", but he makes a compelling case on package/syntax that I had not thought about. I went back and looked at my port of the syntax rule in AIP-191, and I used the file descriptor (meaning the error would be on line 1). That is not terrible, but it does not seem correct either.

I will address this in a separate PR. It is orthogonal to this one.

testutil/problem_stub.go
lukesneeringer: Completely agree, I oppose change detector tests. If you notice, it checks for _containment_, not _equivalence_. And if you look at the test for AIP-131, you will notice that I am testing for a single word -- what the message says the right value is.

testutil/problem_stub.go
robcapo: Thanks Luke, and sorry to derail this PR with this conversation. I made #166 for the issue. Resolving this for now.

rules/testutils/problem_stub.go
```diff
@@ -0,0 +1,85 @@
...
+// problem's descriptor match.
+func (ps *ProblemStub) VerifyDescriptor(p lint.Problem, t *testing.T) {
+	if !reflect.DeepEqual(p.Descriptor, ps.Descriptor) {
+		t.Errorf("Expected descriptor %v; got %v", ps.Descriptor, p.Descriptor)
+	}
+}
+
+// VerifyMessage verifies that the stub's message is included in the
+// problem's message.
+func (ps *ProblemStub) VerifyMessage(p lint.Problem, t *testing.T) {
```
alin04: This pattern breaks our internal style guide.

rules/testutils/problem_stub.go
```diff
@@ -0,0 +1,85 @@
...
+// that problems look correct in tests.
+type ProblemStub struct {
+	Descriptor desc.Descriptor
+	Message    string
+}
+
+// VerifyDescriptor verifies that the stub's descriptor and
+// problem's descriptor match.
+func (ps *ProblemStub) VerifyDescriptor(p lint.Problem, t *testing.T) {
+	if !reflect.DeepEqual(p.Descriptor, ps.Descriptor) {
```
alin04: Remind me again why cmp can't be used?

rules/testutils/problem_stub.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// ProblemStubs is a slice of ProblemStub objects.
+type ProblemStubs []ProblemStub
+
+// SetDescriptor sets the descriptor to each ProblemStub and returns a new
+// ProblemStubs object.
+// It returns a new ProblemStubs (for chaining).
+func (pss ProblemStubs) SetDescriptor(d desc.Descriptor) ProblemStubs {
+	answer := ProblemStubs{}
```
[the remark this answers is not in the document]
lukesneeringer: That was what I thought originally, but if I change the code to this...

```go
func (pss ProblemStubs) SetDescriptor(d desc.Descriptor) ProblemStubs {
	for _, stub := range pss {
		stub.Descriptor = d
	}
	return pss
}
```

...then the descriptor is unchanged and the test fails.

rules/testutils/problem_stub.go
lukesneeringer: Descriptor has unexpected properties. I think I can get it to work, though, let me look into it. (I could not elsewhere because of a circular import but that is not an issue here.)

rules/testutils/problem_stub.go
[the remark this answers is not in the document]
lukesneeringer: Changing it to this, which works:

```go
func (pss ProblemStubs) SetDescriptor(d desc.Descriptor) ProblemStubs {
	for i := range pss {
		pss[i].Descriptor = d
	}
	return pss
}
```

mingzhi:
Consider to make `testutils` as an internal package, and we can open it once it is mature and proved.

lukesneeringer:
> Consider to make `testutils` as an internal package, and we can open it once it is mature and proved.

I would love to, but how? Internal packages can only be imported by their parent directories, so it would have to be copied in the package for every AIP.

alin04:
> > Consider to make `testutils` as an internal package, and we can open it once it is mature and proved.
> 
> I would love to, but how? Internal packages can only be imported by their parent directories, so it would have to be copied in the package for every AIP.

Make an internal/ dir that is peer-level. See https://github.com/golang-standards/project-layout
The name "internal" is merely a convention to indicate those are private - not that it's actually enforced.

mingzhi:
> > Consider to make `testutils` as an internal package, and we can open it once it is mature and proved.
> 
> I would love to, but how? Internal packages can only be imported by their parent directories, so it would have to be copied in the package for every AIP.

I believe their parent and its children directories, i.e., any packages rooted at the parent.

lukesneeringer:
> I believe their parent and its children directories, i.e., any packages rooted at the parent.

Confirmed. Making that change.

lukesneeringer:
@alin04 and I decided not to do this after all. We will alter the tests in different ways.

PR #185, opened 2019-09-18
[feat] Helper method to get HTTP rules.

lukesneeringer:
This PR adds a helper method for getting the google.api.http
binding, and flattening out any additional bindings so that
rule authors can range over them easily.


rules/internal/utils/http.go
```diff
@@ -0,0 +1,50 @@
...
+	// Get the "primary" rule (the direct google.api.http annotation).
+	var httpRule *annotations.HttpRule
+	if x, err := proto.GetExtension(opts, annotations.E_Http); err == nil {
+		httpRule = x.(*annotations.HttpRule)
+		rules = append(rules, httpRule)
+	}
+
+	// Iterate over any additional bindings and flatten them into the
+	// `rules` array.
+	for _, binding := range httpRule.GetAdditionalBindings() {
```
robcapo: This looks like it risks a nil pointer dereference if err on line 37 != nil

robcapo: Nevermind, they have [defensive checks](https://github.com/googleapis/go-genproto/blob/master/googleapis/api/annotations/http.pb.go#L527-L532) in place

robcapo: Not blocking, but go has a cleaner way to do this (for once!): https://play.golang.org/p/hSPAbjk-MkC

lukesneeringer: Done.

rules/internal/utils/http.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Get the "primary" rule (the direct google.api.http annotation).
+	var httpRule *annotations.HttpRule
+	if x, err := proto.GetExtension(opts, annotations.E_Http); err == nil {
+		httpRule = x.(*annotations.HttpRule)
+		rules = append(rules, httpRule)
+	}
+
+	// Add any additional bindings and flatten them into `rules`.
+	rules = append(rules, httpRule.GetAdditionalBindings()...)
```
robcapo: I'd suggest moving this into the `if` block, remove line 36, and change 38 to `:=`

You can also optionally change line 30 to `var rules []*annotations.HttpRule` to avoid an unnecessary allocation (I believe this is a matter of preference because I've seen both forms suggested in readability reviews). The zero value of slices is intended to be useful by itself (i.e. you can call `append` and `len` on it, which will perform allocations if needed).

lukesneeringer: Done. I chose to leave line 30 as-is.

PR #186, opened 2019-09-19
[GoLint] Fixed golint errors when syncing to Google3.

mingzhi:


mingzhi:
Thank you Rob!!!

rules/aip0131/message_rules_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/desc/builder"
-	"google.golang.org/genproto/protobuf/field_mask"
+	fpb "google.golang.org/genproto/protobuf/field_mask"
```
lukesneeringer: Why were these lint errors exactly?

lukesneeringer:
Hey, can we talk about this one?
If GitHub is our source of truth, we can not expect (potentially external) rule authors to conform to arbitrary lint rules that are not enforced _here_. This is going to be a consistent problem.

I would argue that:
(1) We should probably enforce linting here (as we already do).
(2) We should probably disable any linting in google3. It should accept whatever code is accepted here, with only the modifications required to get it _running_ in google3.

rules/aip0131/message_rules_test.go
mingzhi: Protocol buffer packages should have short names that end in "pb" to distinguish with other canonical Go packages.

PR #188, opened 2019-09-19
[AIP-131] Add rules to check the annotation.

lukesneeringer:


rules/aip0131/method_rules_test.go
```diff
@@ -92,3 +96,158 @@ func TestResponseMessageName(t *testing.T) {
...
+		{"Invalid", "post", "GetBook", "HTTP GET"},
+		{"Irrelevant", "post", "AcquireBook", ""},
+	}
+
+	// Run each test.
+	for _, test := range tests {
+		t.Run(test.testName, func(t *testing.T) {
+			// Create the appropriate pattern based on the desired verb.
+			var httpRule *annotations.HttpRule
+			if test.verb == "post" {
```
mingzhi: Can you do the set up outside of tests (for loop)? In general, a testing loop does running and comparing only. And from the `tests` cases, one should have a general idea of what are testing without digging into the testing loop.

rules/aip0131/method_rules_test.go
```diff
@@ -92,3 +96,158 @@ func TestResponseMessageName(t *testing.T) {
...
 	}
 }
+
+func TestHttpVerb(t *testing.T) {
+	// Set up testing permutations.
+	tests := []struct {
+		testName   string
+		verb       string
+		methodName string
+		msg        string
```
mingzhi: Use `wantMsg` to be more informative.

rules/aip0131/method_rules_test.go
```diff
@@ -92,3 +96,158 @@ func TestResponseMessageName(t *testing.T) {
...
+			).SetOptions(opts)).Build()
+			if err != nil {
+				t.Fatalf("Could not build %s method.", test.methodName)
+			}
+
+			// Run the method, ensure we get what we expect.
+			problems := httpVerb.Lint(service.GetFile())
+			if test.msg == "" && len(problems) > 0 {
+				t.Errorf("Got %v, expected no problems.", problems)
+			} else if test.msg != "" && !strings.Contains(problems[0].Message, test.msg) {
```
mingzhi: What about `problems` is empty?

lukesneeringer: Then it would panic. I am fine with that.

In general, still not a fan of the test situation. I plan to do more work to make it easier to test problems.

rules/aip0131/method_rules_test.go
lukesneeringer: Leaving this as-is. I would rather do a more thorough pass at comparing problems.

rules/aip0131/method_rules_test.go
lukesneeringer: Done.

PR #189, opened 2019-09-24
[feat] Problem testing redux

lukesneeringer:
This is my (hopefully last ;-)) attempt to make problem testing reasonable. It adds a `Problems` type which is just `[]Problem`, and gives it an `Equal` method so that it can be passed to `cmp.Diff`.

There is also a `SetDescriptor` helper used for tests.

As before, I have migrated only a couple of tests to get feedback before migrating the rest.

And hopefully this is the last major interface change. :-)


lukesneeringer:
Question to @mingzhi and @alin04: Can/should I leave the rule interface as `[]lint.Problem`?

lukesneeringer:
> Question to @mingzhi and @alin04: Can/should I leave the rule interface as `[]lint.Problem`? I would assume and hope that a Problems object also satisfies `[]lint.Problem`.

After playing around with this, it _can_ be done but I expect it is not a good idea. It basically means I have to typecast in every test, which is annoying and not worth the value of being more generic in the interface.

mingzhi:
Hi Luke, I am Okay with a `Problems`, but (strongly) against defining a `Equal` function in the main library. If it is for testing and to satisfy `cmp.Diff`, I suggest you to put them to a test util package.

For any public function, we should assume:
1. people will use it;
2. they will not read the comments;
3. they will not read the implementation.

So the method name should be accurate and informative -- `Equal` means "Equal", nothing else.

mingzhi:
I actually like Java [Truth](https://truth.dev/api/latest/com/google/common/truth/Truth.html) style, e.g.,:
```
assertThat(problems).hasOne().containsMessage("hello, world").containsDescriptor(desc);
```
It is very self-explanatory.

lukesneeringer:
Our Golang docs say that assert libraries are bad, so that is out.
I can bring back `ProblemStubs` to accomplish this, but it requires having `ProblemStubs`.

lint/problem.go
```diff
@@ -46,3 +48,65 @@ type Problem struct {
...
+type Problems []Problem
+
+// Equal determines whether a Problem is sufficiently similar to another
+// to be considered equal.
+//
+// This is intended for unit tests and is intentially generous on what
+// constitutes equality.
+func (problems Problems) Equal(other Problems) bool {
+	// `other` may be nil.
+	// Consider an length-0 slice to be equal to nil.
```
robcapo: Go already considers this to be true, you can remove this condition.

len(nil) == 0

lukesneeringer: It very much does not. :-) I had numerous problems returning `nil` and having `reflect.DeepEqual` complain about lack of equivalence to `[]Problem`.

robcapo: https://play.golang.org/p/NjLpZJVZRTA ?

robcapo: `reflect.DeepEqual` does check for `nil`-ness:
> Slice values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they point to the same initial entry of the same underlying array (that is, &x[0] == &y[0]) or their corresponding elements (up to length) are deeply equal. Note that a non-nil empty slice and a nil slice (for example, []byte{} and []byte(nil)) are not deeply equal.

But that's not what you want in this case, you're assuming if something is nil that it's equivalent to having `len(.) == 0`, which is already true.

lint/problem.go
```diff
@@ -46,3 +48,65 @@ type Problem struct {
...
+
+		// The suggestions, if present, must exactly match.
+		if x.Suggestion != y.Suggestion {
+			return false
+		}
+
+		// When comparing messages, we want to know if the shorter string is a
+		// substring of the longer one, but to preserve the communitive property
+		// of equality, we do not care which is which.
+		short, long := x.Message, y.Message
```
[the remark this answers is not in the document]
lukesneeringer: Fixed with the refactor.

lukesneeringer:
I redid this, and I agree with a lot of the feedback:

  * I moved `Problems` into a testutils package.
  * I stopped calling `cmp.Diff` directly (essentially because it becomes a royal pain when using different types and no common interface) and just added a `Problems.Diff` that returns empty string for equivalence and calls out to `cmp.Diff` otherwise. This basically mimics the `cmp.Diff` interface without a lot of developer pain.
  * I am pretending with all of my might that `problems.Diff` is not an assert library even though it totally is.
  * All use of `Problems` in non-test code is reverted to `[]lint.Problem`

lint/problem.go
lukesneeringer: Confirmed tests still pass with what I have now. Special `nil` behavior excised.

lukesneeringer:
@robcapo / @mingzhi Can you review this? :)

PR #190, opened 2019-09-24
[feat] Add OnlyIf to rules.

lukesneeringer:
There is a really common pattern which looks like...

```go
// Sanity check: Is this a descriptor that this actually
// cares about.
if !relevantMessage(message) {
  return nil
}
```

This formalizes that into an (optional) `OnlyIf` function that
restricts a rule to a scope.

### DIFF
--- a/lint/rule.go
+++ b/lint/rule.go
@@ -74,6 +74,10 @@ type MessageRule struct {
 	// of Problems it finds.
 	LintMessage func(*desc.MessageDescriptor) []Problem
 
+	// OnlyIf accepts a MessageDescriptor and determines whether this rule
+	// is applicable.
+	OnlyIf func(*desc.MessageDescriptor) bool
+
 	noPositional struct{}
 }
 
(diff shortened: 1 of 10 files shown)


mingzhi:
Nitpicking: `Predicate` seems to be a more common name.

lukesneeringer:
I considered `Predicate`, but I think a lot of people do not know what that means. I think `OnlyIf` will be clearer for more people.

lint/rule.go
```diff
@@ -108,6 +114,10 @@ type FieldRule struct {
 	// Problems it finds.
 	LintField func(*desc.FieldDescriptor) []Problem
 
+	// OnlyIf accepts a FieldDescriptor and determines whether this rule
+	// is applicable.
+	OnlyIf func(*desc.FieldDescriptor) bool
```
mingzhi: My feeling is that checking message is more common -- but it is easy to get the message. Since we are not using this function now, maybe we should postpone it.

lukesneeringer: The `OnlyIf` function can easily call `f.GetMessage()` in that case.

mingzhi: Fair enough. Thanks!

rules/aip0131/message_rules.go
```diff
@@ -24,15 +24,10 @@ import (
 
 // The Get standard method should only have expected fields.
 var standardFields = &lint.MessageRule{
-	Name: lint.NewRuleName("core", "0131", "request-message", "name-field"),
-	URI:  "https://aip.dev/131#request-message",
+	Name:   lint.NewRuleName("core", "0131", "request-message", "name-field"),
+	URI:    "https://aip.dev/131#request-message",
+	OnlyIf: isGetRequestMessage,
```
mingzhi: Like it! Very fluent in reading!

PR #193, opened 2019-09-27
[fix] Be stricter about rule names.

lukesneeringer:
We should enforce a stricter standard on rule names.

- They should always be lower-case (no upper-case).
- They should always start with a letter or number.
- They should always use `-` as a word separator, not `_` (kebab-case).
- They should not begin or end with a word separator.

This PR does all of that with the exception of enforcing no
trailing hyphens (I am willing to assume that given the enforcement
of the remaining rules, nobody will try that).

### DIFF
--- a/lint/rule_name_test.go
+++ b/lint/rule_name_test.go
@@ -16,25 +16,49 @@ package lint
 
 import "testing"
 
-func TestRuleName_IsValid(t *testing.T) {
+func TestRuleNameValid(t *testing.T) {
 	tests := []struct {
-		name  RuleName
-		valid bool
+		testName string
+		ruleName RuleName
 	}{
-		{"", false},
-		{"a:::b", false},
-		{"a::::b", false},
-		{"a", true},
-		{"::my_rule", false},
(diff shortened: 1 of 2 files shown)


lint/rule_name_test.go
```diff
@@ -16,26 +16,52 @@ package lint
...
-		{"a", true},
-		{"::my_rule", false},
-		{"my_namespace::", false},
-		{"my_namespace:my_rule", false},
-		{"my_namespace::my_rule", true},
-		{"my_rule", true},
+		{"Lower", "aip"},
+		{"LowerNumber", "aip0121"},
+		{"LowerNumberKebab", "aip-0121"},
+		{"Namespaced", "aip::0121"},
```
mingzhi: Add one more case: "core::aip-0121".

lukesneeringer:
```suggestion
		{"Namespaced", "aip::0121"},
		{"NamespacedHyphen", "core::aip-0121"},
```

PR #194, opened 2019-09-27
[feat] Add parsing from proto string.

lukesneeringer:
This PR adds a `ParseProtoString` function in an internal testutils
directory, to allow easier to read tests, and also permit tests
that are location-sensitive.

This is a hard prerequisite for the syntax and package location work
(which I have more or less done but split into a different PR).


internal/testutils/parse.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/jhump/protoreflect/desc/protoparse"
+	"github.com/lithammer/dedent"
+)
+
+// ParseProtoString parses a string representing a proto file, and returns
+// a FileDescriptor.
+//
+// It dedents the string before parsing.
+// It is unable to handle imports, and panics if there is any error.
+func ParseProtoString(src string) *desc.FileDescriptor {
```
mingzhi: Consider `ParseProtoStringDedented(fileName, src string) *desc.FileDescriptor`:
1. It informs developers that the `src` will be dedented -- note in comments is not enough and they will be surprise when checking the returned locations;
2. `fileName` allows them to specify the wanted file name, instead of the baked `"test.proto"`.

lukesneeringer: 1. At least for me, the way I check returned locations is to "get it wrong" (usually by putting in intentionally bogus numbers) and then letting the test fail, and then modifying the test to match the location I actually got. Also, it will be a surprise once, and then after that it is a convenience.

2. YAGNI.

robcapo: +1 to mingzhi's suggestion, but I don't feel too strongly about it for a test utility.

internal/testutils/parse.go
```diff
@@ -0,0 +1,42 @@
...
+	parser := protoparse.Parser{
+		Accessor: protoparse.FileContentsFromMap(map[string]string{
+			"test.proto": strings.TrimSpace(dedent.Dedent(src)),
+		}),
+		IncludeSourceCodeInfo: true,
+	}
+	fds, err := parser.ParseFiles("test.proto")
+	if err != nil {
+		panic(err)
```
robcapo: Don't do this. panicking is really only intended if the program can't continue running. Either return an error as a second arg, or take in a `testing.T` as the first argument and call `t.Fatal` on it.

lukesneeringer: I really do not want to having this return an error as a second argument, because then _every single rule test_ has to deal with it. This is not user input -- this is a string hard-coded into a file that will _always_ parse or _never_ parse.

I am slightly less opposed to passing a `*testing.T`, but when I did that before, I was told it was an anti-pattern.

robcapo: I'd like to hear more about why it's an anti-pattern. At worst it's still better than panicking IMO. As is, this will currently cause all tests to halt and the binary will exit.

There are many [examples](http://cs/search/?q=testing.T,+lang:go&sq=package:piper+file://depot/google3+-file:google3/experimental&type=cs) of this pattern in google3

lukesneeringer: I am working on changing this, but it is worth noting that `t.Fatalf` calls runtime.goexit anyway...

```
--- FAIL: TestParseProtoStringError (0.00s)
panic: test executed panic(nil) or runtime.Goexit

goroutine 6 [running]:
testing.tRunner.func1(0xc00012c500)
	/usr/lib/google-golang/src/testing/testing.go:874 +0x3a3
runtime.Goexit()
	/usr/lib/google-golang/src/runtime/panic.go:563 +0xec
testing.(*common).FailNow(0xc00012c600)
	/usr/lib/google-golang/src/testing/testing.go:653 +0x39
testing.(*common).Fatalf(0xc00012c600, 0x6f4788, 0x2, 0xc000085ec0, 0x1, 0x1)
	/usr/lib/google-golang/src/testing/testing.go:716 +0x90
github.com/googleapis/api-linter/internal/testutils.ParseProtoString(0xc00012c600, 0x7039a2, 0x59, 0x47e696)
	/usr/local/google/home/lukesneeringer/Code/Google/api-linter/internal/testutils/parse.go:40 +0x23b
github.com/googleapis/api-linter/internal/testutils.TestParseProtoStringError(0xc00012c500)
	/usr/local/google/home/lukesneeringer/Code/Google/api-linter/internal/testutils/parse_test.go:60 +0x59
testing.tRunner(0xc00012c500, 0x704618)
	/usr/lib/google-golang/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/usr/lib/google-golang/src/testing/testing.go:960 +0x350
FAIL	github.com/googleapis/api-linter/internal/testutils	0.037s
FAIL
Error: Tests failed.
```

robcapo: runtime.Goexit only ends the currently running goroutine, so it shouldn't affect other tests. See [T.FailNow](http://go/godoc/testing/#T.FailNow)

lukesneeringer: Done.

internal/testutils/parse.go
lukesneeringer: I would like to leave this one as-is. It exchanges a one-time surprise for long-time convenience.

PR #195, opened 2019-09-27
[feat] Support locations in rules.

lukesneeringer:
~Uses #194 as a base.~ Merged.

This PR refactors how we think about locations.

1. It internalizes the distinction between a `lint.Location`
   and a `dpb.SourceCodeInfo_Location`. As far as rule authors
   are concerned, we only use the latter (as natively returned
   from `GetSourceInfo()`. Coversion happens when it is time
   to marshal to JSON or YAML. This avoids a situation where
   rule authors have to think about two different things with
   the same name.

2. It restores `Location` (but now the `dpb` one) as a first-class
   attribute on `Problem`.

3. It provides helpers for getting the package, syntax, or name.
   All helpers return the correct object to just be passed through.


lukesneeringer:
I need to write more tests (specifically for `Problem`) before this is actually ready. I also need to run it and make sure it works. ;-)

lint/location.go
```diff
@@ -14,43 +14,97 @@
...
 
-// Location describes a location in a source code file.
+// SyntaxLocation returns the location of the syntax definition in a file
+// descriptor.
 //
-// Note: positions are one-based.
-type Location struct {
-	Start Position `json:"start_position" yaml:"start_position"`
-	End   Position `json:"end_position" yaml:"end_position"`
+// If the location can not be found (for example, because there is no syntax
```
mingzhi: Why not `nil`? Especially that we are returning a pointer.

lint/location.go
```diff
@@ -14,43 +14,97 @@
...
-	Column int `json:"column_number" yaml:"column_number"`
+type sourceInfo map[string]*dpb.SourceCodeInfo_Location
+
+// findLocation returns the best location it can find for the given path.
+//
+// If the requested path can not be found, it climbs the ancestry tree until
+// it finds one, and ultimately returns a location corresponding to the
+// beginning of the file if it can not find anything.
+func (si sourceInfo) findLocation(path []int32) *dpb.SourceCodeInfo_Location {
+	// Base case: If we have no path left, send back a fake location that
```
mingzhi: As we return a pointer, returning `nil` is more accurate.

lint/location.go
```diff
@@ -14,43 +14,97 @@
...
+		}
+	}
+
+	// If the path exists in the source info registry, return that object.
+	if loc, ok := si[strPath(path)]; ok {
+		return loc
+	}
+
+	// No dice; try the immediate parent path.
+	return si.findLocation(path[:len(path)-1])
```
mingzhi: If not found, we should return `nil`, and let developers decide what to do.

lint/location.go
```diff
@@ -14,43 +14,97 @@
...
-// String returns the string representation.
-func (p Position) String() string {
-	return fmt.Sprintf("{line: %d, column: %d}", p.Line, p.Column)
+// sourceInfo compiles the source info object for a given file descriptor.
+// It also caches this into a registry, so subsequent calls using the same
+// descriptor will return the same object.
+func (sir sourceInfoRegistryType) sourceInfo(fd *desc.FileDescriptor) sourceInfo {
+	answer, ok := sir[fd]
+	if !ok {
+		answer = sourceInfo{}
```
mingzhi: Should we use `make` to create a `map`?

lint/location.go
```diff
@@ -14,43 +14,97 @@
...
+
+		// This file descriptor does not yet have a source info map.
+		// Compile one.
+		for _, loc := range fd.AsFileDescriptorProto().GetSourceCodeInfo().GetLocation() {
```
mingzhi: We should check if source code infor is available, i.e., `GetSourceCodeInfo()` may return `nil`.

lint/location.go
```diff
@@ -14,43 +14,97 @@
...
+		}
+
+		// Now that we calculated all of this, cache it on the registry so it
+		// does not need to be calculated again.
+		sir[fd] = answer
+	}
+	return answer
 }
+
+var sourceInfoRegistry = sourceInfoRegistryType{}
```
mingzhi: Should we use `make` to create a `map`?

lint/location.go
lukesneeringer: Because this is about usefulness. The idea is that I _always_ want a valid location of some kind.

If someone sends a `Problem` and there is some unforeseen issue, the graceful fallback behavior is "put the error at the top of the file", as opposed to "return nil and probably panic later".

lint/location.go
lukesneeringer: Again, this is intentional, moving up the chain to the closest possible match.

The premise of "return `nil` and let developers decide what to do" does not hold water -- at the point that this does the wrong thing, it is out of the rule author's hands.

lint/location.go
lukesneeringer: Does it matter?

mingzhi: My misunderstanding -- I thought map needs `make` to be initialized -- thanks!

lint/location.go
mingzhi: But in the `Problem`, we already do the fallback to get the location from the `Descriptor` if `Location == nil`, which in this case is the `FileDescriptor`.

lukesneeringer: Right, and this is what we _use_ for that. It is therefore important that we return a valid location.

lint/location.go
mingzhi: Please note that `SourceCodeInfo_Location` has comments information. If, for example, a rule wants to check the comments of a package, it is wrong to return the file's comments (which may be still `nil`). But take one step back, I am also Okay with what you do here -- I don't have strong objection, and feel free to make it as-it and resolve our conservation. Thank you!

lint/location.go
robcapo: Agree that we should return nil. We can still provide convenience in the core linter (for example, as we already do: if locaiton is nil, we set it to the location of the descriptor).

If you truly want to have this fallback to the start location of the file, you can name the function something like `SyntaxLocationOrFileStart`, but I don't see the advantage of doing this.

The user will be using this to attach locations to problems. If we gracefully handle the `nil` case in the core linter code, we shouldn't have to worry about panics.

In what case do you see returning the start location as useful?

lint/problem.go
```diff
@@ -15,34 +15,128 @@
...
+	return json.Marshal(p.marshal())
+}
+
+// MarshalYAML defines how to represent a Problem in YAML.
+func (p Problem) MarshalYAML() ([]byte, error) {
+	return yaml.Marshal(p.marshal())
+}
+
+// Marshal defines how to represent a serialized Problem.
+func (p Problem) marshal() interface{} {
```
robcapo: :+1: much happier with this than previous solution. Now we can also make `ruleId` and `category` private, rather than asking developers not to set them in comments.

lint/location.go
lukesneeringer: Convinced, will return `nil`. Waiting for #194 to be merged before making changes here.

lint/problem.go
```diff
@@ -15,34 +15,128 @@
...
 	// Location provides the location of the problem.
 	// DO NOT SET: Set the descriptor instead.
-	Location Location `json:"location" yaml:"location"`
+	Location *dpb.SourceCodeInfo_Location
 
 	// RuleID provides the ID of the rule that this problem belongs to.
 	// DO NOT SET: this field will be set by the linter based on rule info
 	// and user configs.
-	RuleID RuleName `json:"rule_id" yaml:"rule_id"`
+	RuleID RuleName
```
robcapo: Make private, remove `DO NOT SET` docs

lint/problem.go
```diff
@@ -15,34 +15,128 @@
...
 
 	// DO NOT SET:  this field will be set by the linter based on user configs.
-	Category string `json:"category,omitempty" yaml:"category,omitempty"`
+	Category string
```
robcapo: Make private, remove `DO NOT SET` docs

lint/problem.go
lukesneeringer: Done.

lint/problem.go
lukesneeringer: Agreed, done.

lint/location.go
lukesneeringer: This seems to work regardless.

lint/location.go
```diff
@@ -14,43 +14,90 @@
...
-func (l Location) IsValid() bool {
-	return l.Start.IsValid() &&
-		l.End.IsValid() &&
-		(l.End.Line > l.Start.Line ||
-			l.End.Line == l.Start.Line && l.End.Column >= l.Start.Column)
+// PackageLocation returns the location of the package definition in a file
+// descriptor.
+//
+// If the location can not be found (for example, because there is no package
+// statement), it returns beginning of file.
```
mingzhi: Update this comment. I think it now returns `nil` if not found.

lint/location.go
mingzhi: It seems that we always set `IncludeSourceCodeInfo` in the parser, so it works now. But for panic proof, we should check nil.

lint/location_test.go
```diff
@@ -15,45 +15,113 @@
...
+				syntax = "proto3";
+
+				package google.api.linter;
+
+				message Foo {
+				  string bar = 1;
+				}
+			`)),
+		}),
+		IncludeSourceCodeInfo: true,
```
mingzhi: Add a test by disabling `IncludeSourceCodeInfo`, and we should not panic, but return `nil` for locations.

lint/location.go
mingzhi: From your tests, it seems Okay. Thanks!

lint/location_test.go
mingzhi: Ignore it -- it seems you already have such tests.

lint/location.go
lukesneeringer: Fixed.

PR #212, opened 2019-10-07
[AIP-134] Add Update Method rules

michaelbausor:


rules/aip0134/aip0134.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// AddRules accepts a register function and registers each of
+// this AIP's rules to it.
+func AddRules(r lint.RuleRegistry) {
+	r.Register(
+		//httpBody,
```
lukesneeringer: This seems wrong.

rules/aip0134/message_rules.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/builder"
+	fpb "google.golang.org/genproto/protobuf/field_mask"
+)
+
+func extractResource(reqName string) string {
+	return reqName[6:len(reqName)-7]
```
lukesneeringer:
```suggestion
	// Strips "Update" from the beginning and "Request" from the end.
	return reqName[6:len(reqName)-7]
```

rules/aip0134/message_rules.go
```diff
@@ -0,0 +1,102 @@
...
+	OnlyIf: isUpdateRequestMessage,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		// Get the correct message type for google.protobuf.FieldMask.
+		fieldMask, err := desc.LoadMessageDescriptorForMessage(&fpb.FieldMask{})
+		if err != nil {
+			panic(fmt.Sprintf("Unable to load the field mask message: `%v`", err))
+		}
+
+		r := extractResource(m.GetName());
+		rField := builder.FieldTypeMessage(builder.NewMessage(r))
```
lukesneeringer: Why are you building a proto in the lint rule itself? This seems like something that would do in a test.

rules/aip0134/message_rules.go
```diff
@@ -0,0 +1,102 @@
...
+		fields := []struct {
+			name string
```
lukesneeringer: This is wrong. Update methods do not have a name field.

```proto
message UpdateBookRequest {
  Book book = 1;
  google.protobuf.FieldMask = 2;
}
```

rules/aip0134/message_rules.go
```diff
@@ -0,0 +1,102 @@
...
+			field := m.FindFieldByName(f.name)
+			if field == nil {
+				return []lint.Problem{{
+					Message:    fmt.Sprintf("Method %q has no `%q` field", m.GetName(), f.name),
+					Descriptor: m,
+				}}
+			}
+
+			// Rule check: Ensure that the field is the correct type.
+			if field.GetType() != f.typ.GetType() {
```
lukesneeringer: You can just check what the name of the type is; no need to actually import the type itself.

rules/aip0134/message_rules.go
```diff
@@ -0,0 +1,102 @@
...
+var unknownFields = &lint.MessageRule{
+	Name:   lint.NewRuleName("core", "0134", "request-message", "unknown-fields"),
+	URI:    "https://aip.dev/134#request-message",
+	OnlyIf: isUpdateRequestMessage,
+	LintMessage: func(m *desc.MessageDescriptor) (problems []lint.Problem) {
+		resource := extractResource(m.GetName());
+		// Rule check: Establish that there are no unexpected fields.
+		allowedFields := map[string]struct{}{
+			strings.ToLower(resource): {}, // AIP-134
+			"update_mask":             {}, // AIP-134
```
lukesneeringer: string etag (AIP-154)

rules/aip0134/message_rules_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// Run each test individually.
+	for _, test := range tests {
+		fmt.Println(fmt.Sprintf("standardFieldsTest `%q`", test.testName))
+		t.Run(test.testName, func(t *testing.T) {
+			// Create an appropriate message descriptor.
+			bookMsg := builder.NewMessage("Book")
+			message, err := builder.NewMessage(test.messageName).AddField(
+				builder.NewField(test.fieldName, builder.FieldTypeMessage(bookMsg)),
+			).AddField(
+				builder.NewField("update_mask", builder.FieldTypeImportedMessage(fieldMask)),
```
lukesneeringer: update_mask goes on `UpdateBookRequest`, not `Book`.

rules/aip0134/message_rules_test.go
```diff
@@ -0,0 +1,157 @@
...
+	).Build()
+	if err != nil {
+		t.Fatalf("Could not build descriptor.")
+	}
+
+	// Run the lint rule, and establish that it returns the correct
+	// number of problems.
+	wantProblems := testutils.Problems{{
+		Descriptor: message.GetFields()[0],
+		Message:    "`\"book\"` field on Update RPCs should be of type `\"TYPE_MESSAGE\"`",
```
lukesneeringer: `Parchment book = 1;` would pass this, and should not.

rules/aip0134/method_rules.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Update methods should use the resource as the response message
+var responseMessageName = &lint.MethodRule{
+	Name:   lint.NewRuleName("core", "0134", "response-message", "name"),
+	URI:    "https://aip.dev/134#guidance",
+	OnlyIf: isUpdateMethod,
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		// Rule check: Establish that for methods such as `UpdateFoo`, the response
+		// message is named `Foo`.
```
lukesneeringer: google.longrunning.Operation is also valid here.

rules/aip0134/method_rules.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Update methods should have an HTTP body.
+var httpBody = &lint.MethodRule{
+	Name:   lint.NewRuleName("core", "0134", "http-body"),
+	URI:    "https://aip.dev/134#guidance",
+	OnlyIf: isUpdateMethod,
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		fieldName := strcase.SnakeCase(m.GetName()[6:])
+		// Establish that the RPC has no HTTP body.
```
lukesneeringer: Fix this comment.

rules/aip0134/method_rules.go
michaelbausor: Done.

rules/aip0134/method_rules.go
michaelbausor: Done

rules/aip0134/message_rules.go
lukesneeringer: Right, but what you have is wrong (unless you have fixed it). What you are actually checking is...

```proto
message Book {
  string name = 1;
  google.protobuf.FieldMask = 2;
}
```

...which is very much not the same thing. :-)

lukesneeringer: Ah, yes, I see now.

rules/aip0134/message_rules.go
```diff
@@ -0,0 +1,95 @@
...
+	Name:   lint.NewRuleName("core", "0134", "request-message", "name-field"),
+	URI:    "https://aip.dev/134#request-message",
+	OnlyIf: isUpdateRequestMessage,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		r := extractResource(m.GetName())
+		fields := []struct {
+			name string
+			typ  string
+		}{
+			{strings.ToLower(r), r},
```
lukesneeringer: You need `strings.ToLower(strcase.SnakeCase(r))`.

Otherwise, a resource `FooBar` will ask for field name `foobar` (wrong) rather than `foo_bar` (right).

rules/aip0134/message_rules.go
lukesneeringer: You are right.

rules/aip0134/message_rules_test.go
lukesneeringer: You are right; I misread.

rules/aip0134/message_rules.go
michaelbausor: Very good point, done (and updated tests).

Issue #173, opened 2019-09-18
AIP-203: Field behavior / input only format


robcapo:
Old linter rule: https://g3doc.corp.google.com/google/api/tools/linter/g3doc/rules/input-only-format.md?cl=head

PR #214, opened 2019-10-07
Added field annotation input_only rule.

mingzhi:
This PR will fix issue #173

### DIFF
--- a/rules/aip0203/aip0203.go
+++ b/rules/aip0203/aip0203.go
@@ -1,0 +1,44 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


rules/aip0203/aip0203.go
```diff
@@ -0,0 +1,44 @@
...
+// AddRules adds all of the AIP-203 rules to the provided registry.
+func AddRules(r lint.RuleRegistry) {
+	r.Register(
+		inputOnly,
+	)
+}
+
+// check leading comment of a field and produce a problem
+// if the comment matches the give pattern.
+func checkAnnotation(f *desc.FieldDescriptor, pattern *regexp.Regexp) []lint.Problem {
```
lukesneeringer: This is not really checking an annotation though; it is checking the comment.

PR #216, opened 2019-10-08
[feat] Add common imports to proto string parsing.

lukesneeringer:
I expect several of my decisions here to be controversial. :-)

This adds common imports to proto string parsing, so that tests may import files that they may be expected to need (e.g. google.api annotations, timestamp, etc.).

The path I took is very different than the one @mingzhi did originally, which was to include [api-common-protos](https://github.com/googleapis/api-common-protos) as a submodule. Instead, I am working backwards from the compiled Go protos, meaning that dependency tooling and such will handle keeping them up to date.

It also does the work once at a module level and just does a blanket "make these available everywhere".

Some things I expect to be contentious:

- The internal `decompile` method panics if it encounters problems. This is because it is only called once, effectively on module initialization, and so even if it returned an error, the thing that called it would have to just panic anyway.
- Similarly, two error cases are ignored. This is because I literally have no way to reach them in tests, because we are using trusted input here.


rules/internal/testutils/parse.go
```diff
@@ -16,14 +16,25 @@ package testutils
...
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/desc/protoparse"
+	"github.com/jhump/protoreflect/desc/protoprint"
 	"github.com/lithammer/dedent"
+
+	// These imports cause the common protos to be registered with
+	// the protocol buffer registry, and therefore make the call to
+	// `proto.FileDescriptor` work for the imported files.
+	_ "google.golang.org/genproto/googleapis/api/annotations"
```
mingzhi: Group these common protos into a separate file, e.g., `common_imported_protos.go`, like [this one](https://github.com/googleapis/api-linter/blob/210f02b24ebe301d581cd8560e1010d909911d03/parser/internal/protos/api_common_protos.go).

rules/internal/testutils/parse.go
```diff
@@ -80,3 +102,69 @@ func ParseProto3Tmpl(t *testing.T, src string, data interface{}) *desc.FileDescr
 	// Parse the proto as a string.
 	return ParseProto3String(t, protoBytes.String())
 }
+
+// decompile reads compiled Go protos and returns a map of strings
+// corresponding to the original protos.
+//
+// This is used to handle google.api common protos, which some tests need
+// to use.
+func decompile(filenames ...string) map[string]string {
```
mingzhi: As protoparser does not support compiled descriptors as dependency, the process of descriptor -> source file -> descriptor is a bummer. I filed a [ticket](https://github.com/jhump/protoreflect/issues/279) in protoreflect two weeks ago, but got no response.

rules/internal/testutils/parse.go
```diff
@@ -80,3 +102,69 @@ func ParseProto3Tmpl(t *testing.T, src string, data interface{}) *desc.FileDescr
...
+// of its imports here, and the corresponding compiled Go protos must be
+// imported up top.
+var commonProtos = decompile(
+	"google/api/annotations.proto",
+	"google/api/client.proto",
+	"google/api/field_behavior.proto",
+	"google/api/http.proto",
+	"google/api/resource.proto",
+	"google/longrunning/operations.proto",
+	"google/protobuf/any.proto",
```
pedgeio: These are all already done automatically https://github.com/jhump/protoreflect/blob/master/desc/protoparse/std_imports.go#L44

lukesneeringer: Correct, but they still must be included here, because some of the `google/api/` protos import them.

rules/internal/testutils/parse.go
lukesneeringer: Yep, I had the same thought. I may file a pull request to protoreflect in the future, but expect that to take longer. Nothing stops us from altering the implementation then, though.

rules/internal/testutils/parse.go
lukesneeringer: Why? It seems like it is better to have the import in the file that requires it.

mingzhi: Just nitpicking.

rules/internal/testutils/parse.go
```diff
@@ -80,3 +102,69 @@ func ParseProto3Tmpl(t *testing.T, src string, data interface{}) *desc.FileDescr
...
+// to use.
+func decompile(filenames ...string) map[string]string {
+	var fdps []*dpb.FileDescriptorProto
+	for _, filename := range filenames {
+		// Compiled golang protos include a compressed descriptor, which must be
+		// uncompressed before it can be populated into a FileDescriptorProto.
+		compressedDesc := proto.FileDescriptor(filename)
+		r, err := gzip.NewReader(bytes.NewReader(compressedDesc))
+		if err != nil {
+			panic(fmt.Errorf("Corrupted compressed file %q: %v", filename, err))
```
mingzhi: Since this function is used for testing, may be using `testing.T` to `Fatalf` is better.

rules/internal/testutils/parse.go
```diff
@@ -80,3 +102,69 @@ func ParseProto3Tmpl(t *testing.T, src string, data interface{}) *desc.FileDescr
...
+		}
+		descBytes, _ := ioutil.ReadAll(r) // Error case is unreachable; inputs are trusted.
```
mingzhi: Let's just `Fatal` the error, even it is a unlikeable event.

rules/internal/testutils/parse.go
```diff
@@ -80,3 +102,69 @@ func ParseProto3Tmpl(t *testing.T, src string, data interface{}) *desc.FileDescr
...
+
+		// Unmarshal the byte string into a FileDescriptorProto.
+		fdp := &dpb.FileDescriptorProto{}
+		proto.Unmarshal(descBytes, fdp)
```
mingzhi: Check the error, e.g., `if err := proto.Unmarshal; err != nil { ... }`

rules/internal/testutils/parse.go
```diff
@@ -80,3 +102,69 @@ func ParseProto3Tmpl(t *testing.T, src string, data interface{}) *desc.FileDescr
...
+		`), err))
+	}
+
+	// Iterate over the values in the map and populate the original contents
+	// of the protos.
+	printer := &protoprint.Printer{}
+	answer := make(map[string]string)
+	for filename, fd := range fds {
+		// Error case is unreachable; inputs are trusted.
+		protoStr, _ := printer.PrintProtoToString(fd)
```
mingzhi: Let's check the error, in case any error is carried from the upstream library.

rules/internal/testutils/parse.go
lukesneeringer: The problem is I have no way of reaching it in a test, and I dislike having uncovered code.

mingzhi: Okay...

rules/internal/testutils/parse.go
lukesneeringer: It occurs at module setup. We do not want to do this (somewhat expensive) work on every test.

rules/internal/testutils/parse.go
lukesneeringer: Done.

rules/internal/testutils/parse.go
lukesneeringer: Done.

PR #218, opened 2019-10-08
[feat] Field behavior

lukesneeringer:
This adds a helper method for reading the field_behavior
annotation.

Uses #216 as a base (for the test).


rules/internal/utils/extension.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/golang/protobuf/proto"
+	"github.com/jhump/protoreflect/desc"
+	apb "google.golang.org/genproto/googleapis/api/annotations"
+)
+
+// GetFieldBehavior returns a slice of FieldBehavior annotations for
+// the given field.
+func GetFieldBehavior(f *desc.FieldDescriptor) []apb.FieldBehavior {
```
mingzhi: Consider to use plural for a slice result: `GetFieldBehaviors`.

lukesneeringer: The plural of "behavior" is "behavior", and this matches how the proto is written.

mingzhi: Ahhh, forgive my English knowledge. Thanks!

PR #221, opened 2019-10-09
AIP-203: Field behavior / immutable format

michaelwq:
AIP-203: Field behavior / immutable format

Rename the existing file aip203_test.go to aip0203_test.go


rules/aip0203/immutable.go
```diff
@@ -0,0 +1,60 @@
...
+	OnlyIf: withoutImmutableFieldBehavior,
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		return checkLeadingComments(f, immutableRegexp, "IMMUTABLE")
+	},
+}
+
+var immutableRegexp = regexp.MustCompile("(?i).*immutable.*")
+
+func withoutImmutableFieldBehavior(f *desc.FieldDescriptor) bool {
+	opts := f.GetFieldOptions()
```
mingzhi: Can you use [`GetFieldBehavior`](https://github.com/googleapis/api-linter/blob/2f7e4b96ffe9dbebfbef2e04d360dfe2eacf4783/rules/internal/utils/extension.go#L25) in the `rules/internal/utils` package?

PR #233, opened 2019-10-14
[AIP-141] Initial rule: num_foos -> foo_count

lukesneeringer:


rules/aip0141/count_test.go
```diff
@@ -0,0 +1,56 @@
...
+		testName  string
+		FieldName string
+		want      string
+		err       bool
+	}{
+		{"Valid", "item_count", "item_count", false},
+		{"NumItems", "num_items", "item_count", true},
+		{"NumResults", "num_results", "result_count", true},
+		{"NumErrors", "num_errors", "error_count", true},
+		{"NumMoose", "num_moose", "moose_count", true}, // singular of "moose" is "moose"
```
mingzhi: Add a tricky one: `{"NumTeeth", "num_teeth", "num_tooth", true}`

rules/aip0141/count.go
```diff
@@ -0,0 +1,40 @@
...
+	pluralize "github.com/gertd/go-pluralize"
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var count = &lint.FieldRule{
+	Name: lint.NewRuleName("core", "0141", "count-suffix"),
+	URI:  "https://aip.dev/141#guidance",
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		if n := f.GetName(); strings.HasPrefix(n, "num_") {
```
mingzhi: Do we need to check if the field type is an integer: `int32`, `int64` ...?

lukesneeringer: No, I plan a different role for that.

rules/aip0141/count_test.go
lukesneeringer: Good idea.

lukesneeringer: Done, but I used geese, because I am a jerk. ;-)

rules/aip0141/count.go
```diff
@@ -0,0 +1,40 @@
...
+			want := pluralize.NewClient().Singular(n[4:]) + "_count"
```
mingzhi: Will people use `num_of_books` or `number_of_books` as a quantity field? For `num_of_books`, we will then produce a wired suggestion `of_book_count`.

lukesneeringer: I am not trying to address every possible contingency, and will make a separate rule later prohibiting prepositions.

mingzhi: Thanks!

PR #237, opened 2019-10-17
[AIP-143] Add standardized codes rules.

lukesneeringer:
Uses #236 as a base.

### DIFF
--- a/rules/aip0143/aip0143.go
+++ b/rules/aip0143/aip0143.go
@@ -1,0 +1,39 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


rules/aip0143/aip0143.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, n := range []string{
+		"mime_type",
+		"country_code",
+		"currency_code",
+		"language_code",
+		"time_zone",
+	} {
+		standardFields[n] = struct{}{}
+	}
+}
```
mingzhi: Should we wrap this block to a function `IsStandardField(string)bool`? It would be easy to understand and make the map read only.

lukesneeringer: That would involve building the map each time, rather than just once.

mingzhi: Consider:
```golang
var standardFields = map[string]bool{
  "mime_type": true,
  "country_code": true,
  "currency_code": true,
  "language_code": true,
  "time_zone": true,
}

func isStandardField(fieldName string) bool {
  // this allows us to pre-process the field name here if needed in the future.
  // for example, strings.ToLower(fieldName).
  return standardFields[fieldName]
}
```

lukesneeringer: That is basically just adding a method on top of what is here. The map does not become read-only (something could still modify it).

I guess I do not mind a method, but it is worth pointing out that the read-only aspect of this no longer holds.

mingzhi: I am Okay either way :) I don't think we should spend more time on debating it :) Thanks!

PR #240, opened 2019-10-17
[AIP-140] Add rule prohibiting prepositions.

lukesneeringer:


rules/aip0140/prepositions.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/scylladb/go-set/strset"
+)
+
+var noPrepositions = &lint.FieldRule{
+	Name: lint.NewRuleName("core", "0140", "prepositions"),
+	URI:  "https://aip.dev/140#prepositions",
+	LintField: func(f *desc.FieldDescriptor) (problems []lint.Problem) {
+		banned := strset.New(
```
mingzhi: I am not sure if using `go-set` package would save more time than importing and maintaining the package in Google3. The package is currently not in Google3.

lukesneeringer: Oh, ugh. That makes sense.
Is there an equivalent package that _is_ imported into google3?

mingzhi: `bitbucket.org/creachadair/stringset` is a very popular one (>1000 imports) in Google3.

lukesneeringer: Sold, thanks!

lukesneeringer: Done. Thanks.

PR #242, opened 2019-10-17
AIP-122: Resource names / HTTP url patterns

michaelwq:
AIP-122: Resource names / HTTP url patterns

### DIFF
--- a/rules/aip0122/http_url_pattern.go
+++ b/rules/aip0122/http_url_pattern.go
@@ -24,36 +24,51 @@ import (
 
 // HTTP URL pattern shouldn't include underscore("_")
 var httpURLPattern = &lint.MethodRule{
-	Name: lint.NewRuleName("core", "0122", "http-url-pattern"),
-	URI:  "https://aip.dev/122#guidance",
+	Name:   lint.NewRuleName("core", "0122", "http-url-pattern"),
+	URI:    "https://aip.dev/122#guidance",
+	OnlyIf: isUnderscoreExisted,
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
 		// Establish that the RPC uri shouldn't include the `_`.
-		for _, httpRule := range utils.GetHTTPRules(m) {
-			if uri := httpRule.GetGet(); uri != "" {
-				return getURLProblems(uri, m)
-			} else if uri := httpRule.GetPut(); uri != "" {
-				return getURLProblems(uri, m)
-			} else if uri := httpRule.GetPost(); uri != "" {
-				return getURLProblems(uri, m)
-			} else if uri := httpRule.GetDelete(); uri != "" {
-				return getURLProblems(uri, m)
-			} else if uri := httpRule.GetPatch(); uri != "" {
-				return getURLProblems(uri, m)
-			} else if custom := httpRule.GetCustom(); custom != nil && custom.GetPath() != "" {
-				return getURLProblems(custom.GetPath(), m)
-			}
-		}
-
-		return nil
+		return []lint.Problem{{
+			Message:    "HTTP URL should use camel case, but not snake case.",
+			Descriptor: m,
+		}}
 	},
 }
 
-func getURLProblems(uri string, m *desc.MethodDescriptor) []lint.Problem {
-	if strings.Contains(uri, "_") {
-		return []lint.Problem{{
-			Message:    "HTTP URL pattern shouldn't include underscore(`_`).",
-			Descriptor: m,
-		}}
+func isUnderscoreExisted(m *desc.MethodDescriptor) bool {
+
+	for _, httpRule := range utils.GetHTTPRules(m) {
+		if uri := httpRule.GetGet(); uri != "" {
+			if strings.Contains(uri, "_") {
+				return true
+			}
+		}
+		if uri := httpRule.GetPut(); uri != "" {
+			if strings.Contains(uri, "_") {
+				return true
+			}
+		}
+		if uri := httpRule.GetPost(); uri != "" {
+			if strings.Contains(uri, "_") {
+				return true
+			}
+		}
+		if uri := httpRule.GetDelete(); uri != "" {
+			if strings.Contains(uri, "_") {
+				return true
+			}
+		}
+		if uri := httpRule.GetPatch(); uri != "" {
+			if strings.Contains(uri, "_") {
+				return true
+			}
+		}
+		if custom := httpRule.GetCustom(); custom != nil && custom.GetPath() != "" {
+			if strings.Contains(custom.GetPath(), "_") {
+				return true
+			}
+		}
 	}
-	return nil
+	return false
 }
--- a/rules/aip0122/http_url_pattern.go
+++ b/rules/aip0122/http_url_pattern.go
@@ -1,0 +1,59 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,59 @@
...
+// HTTP URL pattern shouldn't include underscore("_")
+var httpURLPattern = &lint.MethodRule{
+	Name: lint.NewRuleName("core", "0122", "http-url-pattern"),
+	URI:  "https://aip.dev/122#guidance",
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		// Establish that the RPC uri shouldn't include the `_`.
+		for _, httpRule := range utils.GetHTTPRules(m) {
+			if uri := httpRule.GetGet(); uri != "" {
+				return getURLProblems(uri, m)
+			} else if uri := httpRule.GetPut(); uri != "" {
```
mingzhi: Consider to break `else if`, for example:
```golang
if uri := httpRule.GetGet(); uri != "" {
		return getURLProblems(uri, m)
}

if uri := httpRule.GetPut(); uri != "" {
   return getURLProblems(uri, m)
}
...
```

lukesneeringer: This should probably be a helper method.

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,59 @@
...
+			} else if custom := httpRule.GetCustom(); custom != nil && custom.GetPath() != "" {
+				return getURLProblems(custom.GetPath(), m)
+			}
+		}
+
+		return nil
+	},
+}
+
+func getURLProblems(uri string, m *desc.MethodDescriptor) []lint.Problem {
```
lukesneeringer: This is the component that belongs in the rule; the stuff in the rule should be the helper method.

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,59 @@
...
+	if strings.Contains(uri, "_") {
+		return []lint.Problem{{
+			Message:    "HTTP URL pattern shouldn't include underscore(`_`).",
```
lukesneeringer: We should explicitly say "camel case, not snake case".

rules/aip0122/http_url_pattern_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/jhump/protoreflect/desc/builder"
+	"google.golang.org/genproto/googleapis/api/annotations"
+)
+
+func TestHttpUriField(t *testing.T) {
+	tests := []struct {
+		testName   string
+		input      string
+		output     string
+		httpRule   *annotations.HttpRule
```
lukesneeringer: This would probably be a lot less painful if you used `testutils.ParseProto3Tmpl`.

michaelwq: Done.

rules/aip0122/http_url_pattern.go
michaelwq: Done.

rules/aip0122/http_url_pattern.go
michaelwq: Done.

rules/aip0122/http_url_pattern.go
michaelwq: Done.

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"strings"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// HTTP URL pattern shouldn't include underscore("_")
+var httpURLPattern = &lint.MethodRule{
```
lukesneeringer: Name this `camelCase`. There may be other rules we enforce here in the future.

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,74 @@
...
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		// Establish that the RPC uri shouldn't include the `_`.
+		return []lint.Problem{{
+			Message:    "HTTP URL should use camel case, but not snake case.",
+			Descriptor: m,
+		}}
+	},
+}
+
+func isUnderscoreExisted(m *desc.MethodDescriptor) bool {
```
lukesneeringer: Name this `hasUnderscore`

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// HTTP URL pattern shouldn't include underscore("_")
+var httpURLPattern = &lint.MethodRule{
+	Name:   lint.NewRuleName("core", "0122", "http-url-pattern"),
+	URI:    "https://aip.dev/122#guidance",
+	OnlyIf: isUnderscoreExisted,
```
lukesneeringer: This technically works, but is not really the spirit. This really should not be here and have an `if hasUnderscore {` in `LintMethod`.

rules/aip0122/http_url_pattern_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+func TestHttpUriField(t *testing.T) {
+	tests := []struct {
+		testName string
+		src      string
+		problems testutils.Problems
+	}{
+		{
+			testName: "Valid",
+			src: `import "google/api/annotations.proto";
```
lukesneeringer: These should be similar enough that a template can be used.

rules/aip0122/http_url_pattern.go
michaelwq: Done.

rules/aip0122/http_url_pattern.go
michaelwq: Done

rules/aip0122/http_url_pattern.go
michaelwq: Done

rules/aip0122/http_url_pattern_test.go
michaelwq: Keep the old test because double quotes in substituted values will cause the issue when using a template.

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,76 @@
...
+			return []lint.Problem{{
+				Message:    "HTTP URL should use camel case, but not snake case.",
+				Descriptor: m,
+			}}
+		}
+		return nil
+	},
+}
+
+func hasUnderscore(m *desc.MethodDescriptor) bool {
```
mingzhi: Consider to rename it to `hasUnderscoreURI`.

rules/aip0122/http_url_pattern.go
```diff
@@ -0,0 +1,76 @@
...
+
+	for _, httpRule := range utils.GetHTTPRules(m) {
+		if uri := httpRule.GetGet(); uri != "" {
```
mingzhi: Consider to rewrite the condition match into a `for` loop:
```go
uris := []string{httpRule.GetGet(), httpRule.GetPost(), httpRule.GetPut(), httpRule.GetDelete(), httpRule.GetPatch()}
if custom := httpRule.GetCustom(); custom != nil {
  uris = append(uris, custom.GetPath())
}
for _, uri := range uris {
  if strings.Contains(uri, "_") {
    return true
  }
}
```

lukesneeringer: No need, since #252 makes this entire code block unnecessary. :-)

michaelwq: Done.

PR #245, opened 2019-10-18
[AIP-133/134/135] Test LRO responses better.

lukesneeringer:
This adds response type checking when the protobuf response
type is an LRO.

(It also fixes a couple of bugs I found in the process. ;-))

### DIFF
--- a/rules/aip0134/method_rules.go
+++ b/rules/aip0134/method_rules.go
@@ -55,25 +55,31 @@ var responseMessageName = &lint.MethodRule{
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
 		// Rule check: Establish that for methods such as `UpdateFoo`, the response
 		// message is `Foo` or `google.longrunning.Operation`.
-		got := m.GetOutputType().GetFullyQualifiedName()
-		want := []string{
-			m.GetName()[6:],
-			"google.longrunning.Operation",
+		want := m.GetName()[6:]
+		got := m.GetOutputType().GetName()
+
+		// If the return type is an LRO, use the annotated response type instead.
+		if m.GetOutputType().GetFullyQualifiedName() == "google.longrunning.Operation" {
+			got = utils.GetOperationInfo(m).GetResponseType()
 		}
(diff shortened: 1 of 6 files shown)


rules/aip0134/method_rules.go
```diff
@@ -55,25 +55,31 @@ var responseMessageName = &lint.MethodRule{
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
 		// Rule check: Establish that for methods such as `UpdateFoo`, the response
 		// message is `Foo` or `google.longrunning.Operation`.
-		got := m.GetOutputType().GetFullyQualifiedName()
-		want := []string{
-			m.GetName()[6:],
-			"google.longrunning.Operation",
+		want := m.GetName()[6:]
```
mingzhi: Consider use `strings.Replace(m.GetName(), "Update", "", 1)`:
1. It avoids runtime errors about slice out of bound, if somehow we make a mistake in `isUpdateMethod`;
2. It is better documented -- I needed sometime to figure out what is `6` for.

rules/aip0135/method_rules.go
```diff
@@ -52,27 +53,38 @@ var responseMessageName = &lint.MethodRule{
...
-		want := []string{
+		// message is `google.protobuf.Empty` or `Foo`.
+		got := m.GetOutputType().GetName()
+		if stringset.New("Empty", "Operation").Contains(got) {
+			got = m.GetOutputType().GetFullyQualifiedName()
+		}
+		want := stringset.New(
 			"google.protobuf.Empty",
-			"google.longrunning.Operation",
 			m.GetName()[6:],
```
mingzhi: Consider use `strings.Replace(m.GetName(), "Update", "", 1)`.

lukesneeringer: Delete, but done.

rules/aip0134/method_rules.go
lukesneeringer: Good idea. Done.

PR #247, opened 2019-10-19
[AIP-126] Pilot great godoc documentation.

lukesneeringer:
This PR is an attempt to make documentation for a rule awesome.
It uses godoc to do the heavy lifting.

This requires making a function that returns the rule.
If we decide to go with this route, we will also probably need to tell rules what their godoc link is, because they will not be able to infer it as best as I can tell (rats).

image

We should not merge this until we talk about whether this is the strategy we want.
If we do want this strategy, we should probably shore up our docs everywhere since people are likely view them frequently. I have sent #250 as a down payment on this.

### DIFF
--- a/rules/aip0126/aip0126.go
+++ b/rules/aip0126/aip0126.go
@@ -21,6 +21,6 @@ import "github.com/googleapis/api-linter/lint"
 func AddRules(r lint.RuleRegistry) {
 	r.Register(
 		enumValueUpperSnakeCase,
-		unspecified,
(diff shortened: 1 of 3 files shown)


mingzhi:
The godoc looks nice! I think we can remove the `Details` section, as it can be merged into the function's comment. How did you generate the godoc page?

lukesneeringer:
Concur re: Details. I will do that.

I just ran the local `godoc` server (`localhost:6060` and all that).

PR #248, opened 2019-10-20
[AIP-136] Add initial rules for AIP-136.

lukesneeringer:
This also renames `http-verb` in other AIPs to `http-method`, which is the term most folks use.

Uses #247 as a base, because I made a mistake.

### DIFF
--- a/rules/aip0136/verb_noun.go
+++ b/rules/aip0136/verb_noun.go
@@ -1,0 +1,43 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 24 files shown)


rules/aip0136/verb_noun.go
```diff
@@ -0,0 +1,43 @@
...
+	Name: lint.NewRuleName("core", "0136", "verb-noun"),
+	URI:  "https://aip.dev/136#guidance",
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		// We can not detect this precisely without a full dictionary (probably
+		// not worth it), but we can catch some common mistakes.
+
+		// Are there at least two words?
+		if wordCount := len(strings.Split(strcase.SnakeCase(m.GetName()), "_")); wordCount == 1 {
+			return []lint.Problem{{
+				Message:    "Custom methods should be named using a verb followed by a noun.",
```
mingzhi: Did we miss `OnlyIf` it is a custom method?

lukesneeringer: This one should run on all methods. (`Update` _should_ flag this rule.)

PR #249, opened 2019-10-21
[AIP-191] Prohibit proto versions as filenames.

lukesneeringer:


rules/aip0191/filename.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var filename = &lint.FileRule{
+	Name: lint.NewRuleName("core", "0191", "filename"),
+	URI:  "https://aip.dev/191#guidance",
+	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
+		fn := strings.ReplaceAll(f.GetName(), ".proto", "")
+		if segments := strings.Split(fn, "/"); len(segments) > 1 {
```
mingzhi: Consider to use `filepath.Base(fn)`

lukesneeringer: Done, thanks.

PR #250, opened 2019-10-21
[docs] Add documentation.

lukesneeringer:
I am starting to work on making our godoc page awesome.

### DIFF
--- a/lint/problem.go
+++ b/lint/problem.go
@@ -22,21 +22,34 @@ import (
 )
 
 // Problem contains information about a result produced by an API Linter.
+//
+// All rules return []Problem. Most lint rules return 0 or 1 problems, but
+// occasionally there are rules that may return more than one.
 type Problem struct {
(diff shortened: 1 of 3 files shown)


lint/problem.go
```diff
@@ -22,21 +22,34 @@ import (
...
 type Problem struct {
 	// Message provides a short description of the problem.
+	// This should be no more than a single sentence.
 	Message string
 
 	// Suggestion provides a suggested fix, if applicable.
+	//
+	// This integrates with certain IDEs to provide "push-button" fixes,
+	// so these need to be machine-readable, not just human-readable.
+	// Additionally, when setting `Suggestion`, one should almost always set
```
mingzhi: Should we check if `Location` is present for `Suggestion` in the `lint` package?

lukesneeringer: Not yet, that will break some rules in-place. I do intend to shore that up.

PR #252, opened 2019-10-22
Refactor GetHTTPRules.

lukesneeringer:
In writing an upcoming rule for AIP-156, I realized I have cases
where I *only* care about the URI, and not the method. Getting
this is a huge pain.

I have therefore refactored `GetHTTPRules` to return a custom
struct instead of the underlying proto annotation. The custom
struct has `Method` and `URI` fields that correspond to the
values on the annotation.

In the refactor to the standard method rules, the change clearly
does a better job of expressing intent: `if r.Method == "POST"`
is much clearer than `if r.GetPost() != ""`.

### DIFF
--- a/rules/internal/utils/http.go
+++ b/rules/internal/utils/http.go
@@ -26,8 +26,8 @@ import (
 // and then flattens the values in `additional_bindings`.
 // This allows rule authors to simply range over all of the HTTP rules,
 // since the common case is to want to apply the checks to all of them.
-func GetHTTPRules(m *desc.MethodDescriptor) []*apb.HttpRule {
-	rules := []*apb.HttpRule{}
+func GetHTTPRules(m *desc.MethodDescriptor) []*HTTPRule {
+	rules := []*HTTPRule{}
 
 	// Get the method options.
 	opts := m.GetMethodOptions()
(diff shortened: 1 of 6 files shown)


rules/internal/utils/http.go
```diff
@@ -26,21 +26,69 @@ import (
...
 	// Done; return the rules.
 	return rules
 }
+
+func parseRule(rule *apb.HttpRule) *HTTPRule {
+	answer := &HTTPRule{}
+	if uri := rule.GetGet(); uri != "" {
+		answer.Method = "GET"
+		answer.URI = uri
+	} else if uri := rule.GetPost(); uri != "" {
```
mingzhi: Consider to write it into a `for` loop:
```go
methods := []string{"GET", "POST", "PUT", "PATCH", "DELETE", rule.GetKind()}
uris := []string{rule.GetGet(), rule.GetPost(), rule.GetPut(), rule.GetPatch(), rule.GetDelete(), rule.GetPath()}
for i := range methods {
  if uris[i] != "" {
    return &HTTPRule{
      Method: methods[i],
      URI: uris[i],
      Body: rule.GetBody(),
      ResponseBody: rule.GetResponseBody(),
    }
  }
}
return &HTTPRule{
    Body: rule.GetBody(),
    ResponseBody: rule.GetResponseBody(),
}
```

lukesneeringer: I would rather not, for a couple of reasons:

1. It does not handle the `GetCustom()` case well (if you look at your code carefully, you will notice that it is wrong there).
2. The test case to hit the bottom return line would be weird.
3. I do not really like the "iterate over two slices that happen to be the same size" thing, and Golang has no equivalent of Python's [`itertools.zip`](https://docs.python.org/3/library/functions.html#zip).

lukesneeringer: One alternative that is slightly better (IMO) is:

```go
	oneof := map[string]string{
		"GET":    rule.GetGet(),
		"POST":   rule.GetPost(),
		"PUT":    rule.GetPut(),
		"PATCH":  rule.GetPatch(),
		"DELETE": rule.GetDelete(),
	}
	if custom := rule.GetCustom(); custom != nil {
		oneof[custom.GetKind()] = custom.GetPath()
	}
	for method, uri := range oneof {
		if uri != "" {
			return &HTTPRule{
				Method:       method,
				URI:          uri,
				Body:         rule.GetBody(),
				ResponseBody: rule.GetResponseBody(),
			}
		}
	}
	return nil
```

My gripe with that variant is that I literally can not hit the `return nil` clause without calling the unexported method directly.

mingzhi: Fair enough! Thanks!

lukesneeringer: After thinking about it some more, I did the final variant proposed.

PR #253, opened 2019-10-23
[AIP-136] Add URI suffix check for custom methods.

lukesneeringer:


rules/aip0136/uri_suffix.go
```diff
@@ -0,0 +1,68 @@
...
+
+var uriSuffix = &lint.MethodRule{
+	Name:   lint.NewRuleName("core", "0136", "uri-suffix"),
+	URI:    "https://aip.dev/136#guidance",
+	OnlyIf: isCustomMethod,
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		for _, httpRule := range utils.GetHTTPRules(m) {
+			// URIs should end in a `:` character followed by the name of the method.
+			// However, if the noun is the URI's resource, then we only use `:verb`,
+			// not `:verbNoun`.
```
mingzhi: Consider to give some good and bad examples here to help readers understand (I needed to jump to the testing cases to understand this paragraph).

rules/aip0136/uri_suffix.go
```diff
@@ -0,0 +1,68 @@
...
+			// over time, but the following rules should be mostly correct:
+			//   1. If the URI contains `{name=`, expect `:verb`.
+			//   2. Otherwise, expect `:verbNoun`.
+			//
+			// We blindly assume that the verb is always one word (the "noun" may be
+			// any number of words; they often include adjectives).
+			//
+			// N.B. The LowerCamel(Snake(name)) is because strcase does not translate
+			//      from upper camel to lower camel correctly.
+			want := ":" + strcase.LowerCamelCase(strcase.SnakeCase(m.GetName()))
```
mingzhi: Can we create a `toLowerCamelCase` function and test the corner cases? I don't find the mentioned corner cases (upper camel to lower camel) in the test.

rules/aip0136/uri_suffix.go
lukesneeringer: I agree, but I am going to put that off until we decide on a documentation strategy.

rules/aip0136/uri_suffix.go
lukesneeringer: `strcase.CamelCase("FooBar")` will return `foobar` (not `fooBar`).

lukesneeringer: Rather than a private method, I would rather work with the developer to fix it. :)

lukesneeringer: I sent github.com/stoewer/go-strcase/pull/4 to fix this there.

PR #254, opened 2019-10-23
[AIP-136] Check the HTTP `body` clause.

lukesneeringer:


rules/aip0136/http_body.go
```diff
@@ -0,0 +1,51 @@
...
+			noBody := stringset.New("GET", "DELETE")
+			if !noBody.Contains(httpRule.Method) && httpRule.Body != "*" {
+				// FIXME: We intentionally only return one Problem here.
+				// When we can attach problems to the particular annotation, update this
+				// to return multiples.
+				return []lint.Problem{{
+					Message:    "Custom POST methods should set `body: \"*\"`.",
+					Descriptor: m,
+				}}
+			} else if noBody.Contains(httpRule.Method) && httpRule.Body != "" {
```
mingzhi: Consider to break `else if`.

lukesneeringer: Huh?

mingzhi: I tend to not avoid `else if` because the pattern of `if some condition then return` is better for me to understand. It was a nitpicking.

PR #255, opened 2019-10-23
[AIP-122] Keys must be snake case.

lukesneeringer:
Also, make the tests better.


lukesneeringer:
Marking `do not merge`; this still has one bug.

rules/aip0122/http_url_pattern.go
```diff
@@ -23,19 +23,56 @@ import (
...
-					Message:    "HTTP URL should use camel case, but not snake case.",
+			parsedURI := parseURI(httpRule.URI)
+			if strings.Contains(parsedURI.pattern, "_") {
+				problems = append(problems, lint.Problem{
+					Message:    "HTTP URI patterns should use camel case, not snake case.",
 					Descriptor: m,
-				}}
+				})
+			}
+			for _, v := range parsedURI.vars {
```
mingzhi: Shouldn't we create a separate rule for checking variable names?

rules/aip0122/http_url_pattern.go
```diff
@@ -23,19 +23,56 @@ import (
...
-		return nil
+		return
 	},
 }
+
+// extractURI separates variable names from the URI pattern.
+//
+// It does not maintain the relationship between variable names and what
+// part of the pattern they apply to (we do not need that to lint here).
+func parseURI(uri string) (parsed struct {
```
mingzhi: This function contains complicated logics -- should we create a unit test for it?

lukesneeringer: In general, you should not test private (unexported) functions. I ensured that the existing test hits every statement.

rules/aip0122/http_url_pattern.go
lukesneeringer: It is fundamentally the same string, just different pieces of it. I could be convinced either way.

lukesneeringer:
Fixed my bug.

rules/aip0122/http_url_pattern.go
lukesneeringer: I _think_ I very slightly prefer a single rule. These really are part and parcel the same thing.

mingzhi: Okay, thanks!

rules/aip0122/http_url_pattern.go
mingzhi: Okay.

PR #256, opened 2019-10-24
Use package-level README.md to describe rules in a package

mingzhi:
Preview: https://github.com/googleapis/api-linter/tree/github_page/rules/aip0203

### DIFF
--- a/rules/aip0203/README.md
+++ b/rules/aip0203/README.md
@@ -1,3 +1,9 @@
+# Rules
+
+* [input-only](#input-only)
+* [output-only](#output-only)
+* [immutable](#immutable)
+
 # Input only
 
 When a field on a resource is input only, this should be described using the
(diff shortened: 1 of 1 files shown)


lukesneeringer:
Advantages to this approach:

- Markdown support
- Documentation for linter users is separated clearly from documentation for linter contributors.

Disadvantages to this approach:

- Documentation is farther from the code; more likely to go out of date.
- It somehow does not look quite as nice (surprising given that, well, Markdown)

I could accept either of these.

jgeewax:
Can we add a lint rule style guide if we're going to go this way? For example, the *correct* version should probably be before the incorrect one, etc.

mingzhi:
Yes, I think we can create (1) a style guide and example template; (2) a CI to check that each rule must have a description page and the right style.

And I have the same concern about code and markdown being separated (and we need to catch out-of-date issues in code review). But the markdown page is easy to write and highlight some information.

lukesneeringer:
> Can we add a lint rule style guide if we're going to go this way? For example, the _correct_ version should probably be before the incorrect one, etc.

FWIW, that is the opposite of what you did internally.

mingzhi:
In this PR, I basically copy and paste the internal version :)

jgeewax:
> > Can we add a lint rule style guide if we're going to go this way? For example, the _correct_ version should probably be before the incorrect one, etc.
> 
> FWIW, that is the opposite of what you did internally.

Correct -- internally we have red and green colors though -- so the order matters much less. Since we don't have the color coding, we have to be more careful.

PR #257, opened 2019-10-24
Rewrote the linter cmd to offer a simpler interface.

mingzhi:
Usage:

```sh
api-lint proto_file1 proto_file2 ...
```

```
Usage of api-lint:
  -I value
    	The folders to search for proto imports.
  -c string
    	The config file.
  -f string
    	The format of the printed results.
  -o string
    	The output file path.
```

PR #259, opened 2019-10-24
Shortened cmd args and added `--proto_desc` flag.

mingzhi:


lukesneeringer:
I did like getting rid of `checkproto`/`c` in #257. Can we do that here?

mingzhi:
> I did like getting rid of `checkproto`/`c` in #257. Can we do that here?

I will give a try.

PR #263, opened 2019-10-31
[feat] Linter documentation in GitHub Pages.

lukesneeringer:
This (incredibly large) PR adds a `docs/` directory that we can serve with GitHub Pages.

It provides documentation for three linter rules; the remainder will come in follow-up PRs.

Note that as part of this, I intend to enforce (and soon lint for) some consistency on linter rules, including:

- One rule per file.
- The rule's filename must be derived from the rule name
  (converting `-` to `_` but no other changes).

This is to make things like "View Implmenetation" easy.


lukesneeringer:
The front page looks nice, but needs installation instructions (which can also come in a followup, preferably by not-me):

Screen Shot 2019-10-30 at 5 25 40 PM

lukesneeringer:
The base page for rule documentation will eventually have tiles for the groups of rules. Right now only `core` has any rules implemented, so there is only one. This is fine.

Screen Shot 2019-10-30 at 5 25 51 PM

lukesneeringer:
The rule listing has an expandable "zippy" for each AIP:

Screen Shot 2019-10-30 at 5 25 59 PM

lukesneeringer:
Each rule gets a page, like in the internal linter. There is a quick button on the right that takes you to the implementation code in GitHub. Incorrect and correct code get the red or green line on the side, which interacts better with syntax highlighting than does changing the background color:

Screen Shot 2019-10-30 at 5 26 22 PM

lukesneeringer:
The page overall looks good on mobile, although protos that are too wide will cause horizontal scrolling, as seen in the comments. That's probably just life.

Screen Shot 2019-10-30 at 5 27 10 PM

lukesneeringer:
Print mode looks good too, although I need to figure out what is up with the big blank space on the top. I bet that is broken on the AIP site also.

Note that the incorrect and correct protos now start with `// Incorrect.` and `// Correct.` (That's actually the magic that causes the green and red lines in the screen mode; I intentionally throw away those comments on the screen but preserve them in print.

Screen Shot 2019-10-30 at 5 27 59 PM

lukesneeringer:
This is not perfect, but it is _definitely_ good enough. I want to merge it and then make incremental improvements. This will also unblock getting documentation written for other rules.

PR #275, opened 2019-11-01
[AIP-131] Document method rules.

lukesneeringer:


docs/rules/core/0131-http-body.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the method.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+// (-- api-linter: core::0131::http-body=disabled
+//     api-linter: core::0131::http-method=disabled
```
mingzhi: Why do we need to disable `http-method` too?

lukesneeringer: See the `Important` box below.

PR #283, opened 2019-11-02
Fixed links in TOC.

mingzhi:


docs/_layouts/default.html
```diff
@@ -187,7 +187,11 @@
             {% include linter-rule-summary.html %}
           {% endif -%}
           <section id="linter-toc" class="docs-component-sidebar-toc">
-            {% include toc.html html=content h_min=2 h_max=3 baseurl=page.url %}
+              {% if page.index %}
+              {% include toc.html html=content h_min=2 h_max=3 %}
```
lukesneeringer: Why do we _not_ want `baseurl` in this case? (Also, I am not sure off the top of my head what `page.index` means.)

docs/index.md
```diff
@@ -25,7 +26,7 @@ linter should be used as a useful tool, but not as a substitute for reading and
 understanding API guidance.
 
 Each linter rule has its own [rule documentation][], and rules can be
-[configured][configuration] using a config file, or in a proto file itself.
+[configured](configuration.md) using a config file, or in a proto file itself.
```
lukesneeringer: `./configuration.md`, please.

docs/_layouts/default.html
mingzhi: I am not sure either, but it fixed the links. Otherwise, for example, `installation` was linked to `https://googleapis.github.io/#installation`. `page.index` is similar `page.rule`, that I added a metadata in the `index.md`:
```
---
index:
---
```

docs/index.md
mingzhi: Done.

PR #287, opened 2019-11-02
Added -disable_rule and -enable_rule flags.

mingzhi:


mingzhi:
```
Usage of api-linter:
  -config string
        The linter config file.
  -disable_rule value
        Disable a rule with the given name.
        May be specified multiple times.
  -enable_rule value
        Enable a rule with the given name.
        May be specified multiple times.
  -output_format string
        The format of the linting results.
        Supported formats include YAML, JSON and summary text.
        YAML is the default.
  -output_path string
        The output file path.
        If not given, the linting results will be printed out to STDOUT.
  -proto_descriptor_set string
        A delimited (':') list of files each containing a FileDescriptorSet for searching proto imports.
  -proto_path value
        The folder for searching proto imports.
        May be specified multiple times; directories will be searched in order.
        The current working directory is always used.
```

PR #309, opened 2019-11-06
[refactor] Move location functions into their own package.

### LINKED ISSUE #306: Issues running api-linter on googleapis/googleapis
This is using latest commit https://github.com/googleapis/api-linter/commit/b7c8cacd5574838c908c88ad5c472258156aaba7

```
$ git remote -v
origin	https://github.com/googleapis/googleapis (fetch)
origin	https://github.com/googleapis/googleapis (push)

$ git log -n 1
commit 698b05fe2b05893dfe74a156a5665f879de7fceb (HEAD -> master, origin/master, origin/HEAD)
Author: Google APIs <noreply@google.com>
Date:   Wed Oct 23 08:10:26 2019 -0700

    Synchronize new proto/yaml changes.

    PiperOrigin-RevId: 276276483

$ api-linter $(find . -name '*.proto')
2019/11/05 16:41:01 google/ads/googleads/v1/resources/ad.proto:42:1: duplicate symbol google.ads.googleads.v1.resources.Ad: already defined as message in "./google/ads/googleads/v1/resources/ad.proto"

# removing the leading './' from find output
$ api-linter $(find . -name '*.proto' | sed 's/^\.\///')
2019/11/05 16:41:12 runtime error: invalid memory address or nil pointer dereference

# prints something with relative path for one file and exits with code 0
$ api-linter google/type/date.proto
- file_path: google/type/date.proto
  problems: []

# change the above to absolute path
$ api-linter $(pwd)/google/type/date.proto
2019/11/05 16:57:16 /Users/foo/googleapis/google/type/date.proto: open Users/foo/googleapis/google/type/date.proto: no such file or directory
```

lukesneeringer:
This creates a Go package for `locations` to group these helper
functions together. It also adds several more:

  - FieldType
  - MethodRequestType
  - MethodResponseType

This will better group these functions together in godoc, which
will be useful to rule authors.

Conveniently, also fixes #306.

### DIFF
--- a/locations/descriptor_locations.go
+++ b/locations/descriptor_locations.go
@@ -23,7 +23,10 @@ import (
 //
 // This works for any descriptor, regardless of type (message, field, etc.).
 func DescriptorName(d desc.Descriptor) *dpb.SourceCodeInfo_Location {
-	// All descriptors seem to have `string name = 1`, so this conveniently works.
-	path := append(d.GetSourceInfo().Path, 1)
-	return pathLocation(d.GetFile(), path)
+	if sourceInfo := d.GetSourceInfo(); sourceInfo != nil {
+		// All descriptors seem to have `string name = 1`, so this conveniently works.
+		path := append(sourceInfo.Path, 1)
+		return pathLocation(d.GetFile(), path)
+	}
+	return nil
 }
--- a/locations/descriptor_locations.go
+++ b/locations/descriptor_locations.go
@@ -12,28 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package aip0191
+package locations
 
 import (
-	"path/filepath"
-	"strings"
-
-	"github.com/googleapis/api-linter/lint"
-	"github.com/googleapis/api-linter/locations"
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/jhump/protoreflect/desc"
 )
 
-var filename = &lint.FileRule{
-	Name: lint.NewRuleName("core", "0191", "filenames"),
-	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
-		fn := strings.ReplaceAll(filepath.Base(f.GetName()), ".proto", "")
-		if versionRegexp.MatchString(fn) {
-			return []lint.Problem{{
-				Message:    "The proto version must not be used as the filename.",
-				Descriptor: f,
-				Location:   locations.FilePackage(f),
-			}}
-		}
-		return nil
-	},
+// DescriptorName returns the precise location for a descriptor's name.
+//
+// This works for any descriptor, regardless of type (message, field, etc.).
+func DescriptorName(d desc.Descriptor) *dpb.SourceCodeInfo_Location {
+	// All descriptors seem to have `string name = 1`, so this conveniently works.
+	path := append(d.GetSourceInfo().Path, 1)
+	return pathLocation(d.GetFile(), path)
 }
--- a/locations/descriptor_locations_test.go
+++ b/locations/descriptor_locations_test.go
@@ -12,28 +12,35 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package aip0191
+package locations
 
 import (
-	"path/filepath"
-	"strings"
+	"testing"
 
-	"github.com/googleapis/api-linter/lint"
-	"github.com/googleapis/api-linter/locations"
+	"github.com/google/go-cmp/cmp"
 	"github.com/jhump/protoreflect/desc"
 )
 
-var filename = &lint.FileRule{
-	Name: lint.NewRuleName("core", "0191", "filenames"),
-	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
-		fn := strings.ReplaceAll(filepath.Base(f.GetName()), ".proto", "")
-		if versionRegexp.MatchString(fn) {
-			return []lint.Problem{{
-				Message:    "The proto version must not be used as the filename.",
-				Descriptor: f,
-				Location:   locations.FilePackage(f),
-			}}
+func TestDescriptorName(t *testing.T) {
+	f := parse(t, `
+		message Foo {
+		  string bar = 1;
 		}
-		return nil
-	},
+	`)
+
+	tests := []struct {
+		testName string
+		d        desc.Descriptor
+		wantSpan []int32
+	}{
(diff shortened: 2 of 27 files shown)


locations/field_locations.go
```diff
@@ -0,0 +1,31 @@
...
+
+package locations
+
+import (
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// FieldType returns the precise location for a field's type.
+func FieldType(f *desc.FieldDescriptor) *dpb.SourceCodeInfo_Location {
```
mingzhi: As discovered in issue #306, we have trouble getting `SourceCodeInfo` for a map field, for which `GetSourceInfo()` will return `nil`. You can add a map field in the tests.

locations/descriptor_locations.go
```diff
@@ -0,0 +1,29 @@
...
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// DescriptorName returns the precise location for a descriptor's name.
+//
+// This works for any descriptor, regardless of type (message, field, etc.).
+func DescriptorName(d desc.Descriptor) *dpb.SourceCodeInfo_Location {
+	// All descriptors seem to have `string name = 1`, so this conveniently works.
+	path := append(d.GetSourceInfo().Path, 1)
```
mingzhi: If `d` is a map field, `d.GetSourceInfo()` will return `nil`. See issue #306. You can add a map field in the tests.

bufdev: Any reliance on SourceCodeInfo being present from protoreflect is going to have issues, especially with added FileDescriptorSets from the command line and/or well-known types potentially. You should handle them not being present as a general policy

locations/field_locations.go
lukesneeringer: Oh, good to know. Will fix tomorrow morning.
Is there a protoreflect bug tracking the underlying issue?

lukesneeringer: Fixed.

locations/descriptor_locations.go
lukesneeringer: Good to know, thanks. Fixed.

locations/descriptor_locations.go
```diff
@@ -0,0 +1,32 @@
...
+package locations
+
+import (
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// DescriptorName returns the precise location for a descriptor's name.
+//
+// This works for any descriptor, regardless of type (message, field, etc.).
```
mingzhi: Does it work for `FileDescriptor`? I guess not. But I think `FileDescriptor` will return `nil` for `SourceInfo`, so the code should work as normal. Can you add a test for `FileDescriptor` as well? Thanks!

locations/descriptor_locations_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/jhump/protoreflect/desc"
+)
+
+func TestDescriptorName(t *testing.T) {
+	f := parse(t, `
+		message Foo {
+		  string bar = 1;
```
mingzhi: Can you add a map field for testing?

locations/descriptor_locations_test.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+	`)
+
+	tests := []struct {
+		testName string
+		d        desc.Descriptor
+		wantSpan []int32
+	}{
+		{"Message", f.GetMessageTypes()[0], []int32{2, 8, 11}},
```
mingzhi: Can you test the `FileDescriptor` itself, since it is also a `Descriptor`? Thanks!

locations/field_locations_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/jhump/protoreflect/desc"
+)
+
+func TestFieldLocations(t *testing.T) {
+	f := parse(t, `
+		message Book {
+		  string name = 1;
```
mingzhi: Can you add a test for a map field, since it is a special type? Thanks!

locations/descriptor_locations.go
lukesneeringer: Yes, `FileDescriptor` [has a name, and it is number 1](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L63).
I have no clue what the corresponding `Span` is, though.

locations/descriptor_locations_test.go
lukesneeringer: Just did. Incidentally, they work fine (it does not return nil).

locations/descriptor_locations_test.go
lukesneeringer: Done.

locations/field_locations_test.go
lukesneeringer: Done, but it turned out not to be special.

locations/descriptor_locations_test.go
mingzhi: In that case, can you double check the rule `core::0192::has-comments`? Last night, I had some issues running this rule against the `google/cloud/iot/v1/resources.proto`
```
api-linter google/cloud/iot/v1/resources.proto
```
It checked a message `MetadataEntry`, which I believe was generated by the map field `map<string, string> metadata = 17;` in the `Device` message. And since it is generated, there is no comment. So once we fix the `nil` issue here, the rule will complain about the generated message.

locations/descriptor_locations.go
mingzhi: So the path for the name of the `FileDescriptor` will become `[1]`. And I think there is no `SourceInfo` for the path `[1]`, so it should return `nil`. Can you double check it? Thanks!

locations/descriptor_locations_test.go
lukesneeringer: Sure. Let's make that a separate task though. Can you file an issue?

locations/descriptor_locations.go
lukesneeringer: Correct, it returned nil.

locations/descriptor_locations_test.go
mingzhi: Done. Issue #311

mingzhi: Done. Issue #311

PR #310, opened 2019-11-06
[refactor] Make rule names stricter (again).

lukesneeringer:
This commit refactors `NewRuleName` to take just an AIP number
and the final segment.

It creates `NewGenericRuleName` for cases not covered by this
(for example, rules to enforce internal AIPs).

Basically: It makes the common case better and leaves the advanced
case mostly alone.

### DIFF
--- a/lint/rule_name.go
+++ b/lint/rule_name.go
@@ -15,6 +15,7 @@
 package lint
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 )
@@ -29,11 +30,40 @@ const nameSeparator string = "::"
 
 var ruleNameValidator = regexp.MustCompile("^([a-z0-9][a-z0-9-]*(::[a-z0-9][a-z0-9-]*)?)+$")
 
-// NewRuleName creates a RuleName from segments.
+// NewRuleName creates a RuleName from an AIP number and a unique name within
+// that AIP.
+func NewRuleName(aip int, name string) RuleName {
+	// Determine the group.
+	group := ""
+	if aip > 0 && aip < 1000 {
+		group = "core"
+	}
+
+	// Sanity check: If the group does not exist, complain.
+	if group == "" {
+		panic("Invalid AIP; no available group.")
+	}
+
+	// Get the AIP as a four-character string.
+	aipStr := fmt.Sprintf("%d", aip)
+	for len(aipStr) < 4 {
+		aipStr = "0" + aipStr
+	}
+
+	// Return the rule name.
+	return NewGenericRuleName(group, aipStr, name)
+}
+
+// NewGenericRuleName creates a RuleName from segments.
 // It will join the segments with the "::" separator.
-func NewRuleName(segments ...string) RuleName {
+//
+// Use this function for rules not covered in public AIPs.
+func NewGenericRuleName(segments ...string) RuleName {
 	return RuleName(strings.Join(segments, nameSeparator))
 }
+
+// NewAIPRuleName creates a RuleName based on the AIP number and a unique name
+// within that AIP.
 
 // IsValid checks if a RuleName is syntactically valid.
 func (r RuleName) IsValid() bool {
--- a/lint/rule_name.go
+++ b/lint/rule_name.go
@@ -30,9 +30,8 @@ const nameSeparator string = "::"
 
 var ruleNameValidator = regexp.MustCompile("^([a-z0-9][a-z0-9-]*(::[a-z0-9][a-z0-9-]*)?)+$")
 
-// NewRuleName creates a RuleName from an AIP number and a unique name within
-// that AIP.
-func NewRuleName(aip int, name string) RuleName {
+// getRuleGroup takes an AIP number and returns the appropriate group.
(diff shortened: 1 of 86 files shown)


lint/rule_name.go
```diff
@@ -29,12 +30,41 @@ const nameSeparator string = "::"
...
+	// Determine the group.
+	group := ""
+	if aip > 0 && aip < 1000 {
+		group = "core"
+	}
+
+	// Sanity check: If the group does not exist, complain.
+	if group == "" {
+		panic("Invalid AIP; no available group.")
+	}
```
mingzhi: I think it is more what the linter-linter should check. It is a mistake by the rule developer, and should not panic to the users.

lint/rule_name.go
```diff
@@ -29,12 +30,41 @@ const nameSeparator string = "::"
...
 
-// NewRuleName creates a RuleName from segments.
+// NewRuleName creates a RuleName from an AIP number and a unique name within
+// that AIP.
+func NewRuleName(aip int, name string) RuleName {
+	// Determine the group.
+	group := ""
+	if aip > 0 && aip < 1000 {
+		group = "core"
+	}
```
mingzhi: Instead of burying into the code, can we create a function or global variable to determine the group? It will help us find the right place to add other groups, e.g., "cloud", "youtube" and so on.

lint/rule_name.go
```diff
@@ -29,12 +30,41 @@ const nameSeparator string = "::"
...
+
+	// Sanity check: If the group does not exist, complain.
+	if group == "" {
+		panic("Invalid AIP; no available group.")
+	}
+
+	// Get the AIP as a four-character string.
+	aipStr := fmt.Sprintf("%d", aip)
+	for len(aipStr) < 4 {
```
mingzhi: Consider use `fmt.Sprintf("%04d", api)`

lint/rule_name.go
```diff
@@ -29,12 +30,41 @@ const nameSeparator string = "::"
...
+	// Return the rule name.
+	return NewGenericRuleName(group, aipStr, name)
+}
+
+// NewGenericRuleName creates a RuleName from segments.
 // It will join the segments with the "::" separator.
-func NewRuleName(segments ...string) RuleName {
+//
+// Use this function for rules not covered in public AIPs.
+func NewGenericRuleName(segments ...string) RuleName {
```
mingzhi: I assumed that this will only be used in the tests. Can we add a "test" group? Otherwise, this function is public, there is no way to prevent misusing it other than code review.

lint/rule_name.go
```diff
@@ -29,12 +30,41 @@ const nameSeparator string = "::"
...
 	return RuleName(strings.Join(segments, nameSeparator))
 }
 
+// NewAIPRuleName creates a RuleName based on the AIP number and a unique name
```
mingzhi: Is this comment in the wrong place?

lint/rule_name.go
lukesneeringer: This is a compile time check though. The linter will fail to _build_. It will never fail to execute once built.

lint/rule_name.go
lukesneeringer: Oh, thanks. I checked the docs for that but it was not listed.

lint/rule_name.go
lukesneeringer: Actually there is a valid use case for this -- yours. :-)

When someone writes "private" rules (like what your team will do), their private AIPs will not be in the public index, and `NewGenericRuleName` will be needed.

lint/rule_name.go
lukesneeringer: That was vestigal. Thanks.

lint/rule_name.go
lukesneeringer: Done.

lint/rule_name.go
mingzhi: I see. Can you print the invalid AIP and the name to help us locate the error? Thanks!

lint/rule_name.go
mingzhi: I saw it before -- so I Google it :)

lint/rule_name.go
mingzhi: I think this function can be created anywhere, like
```
package myrules
func NewMyRuleName(segments ...string) lint.RuleName { ... }
```
Indeed, I guess people can even do
```
testRuleName := lint.RuleName("test::test-rule-name")

```
So we don't need to provide it now, and we can wait until people yell at us for it.

lint/rule_name.go
robcapo: I'm confused how this could be a compile time check?

lint/rule_name.go
lukesneeringer: Fair enough.

lukesneeringer: Done.

lint/rule_name.go
mingzhi: I think it would be caught/panic when running tests.

PR #318, opened 2019-11-08
[feat] Add a meta-linter.

### LINKED ISSUE #321: Meta-lint needed: One rule per file.

lukesneeringer:
Right now, this meta-linter ensures that each rule defined in
the linter is actually registered.

We can (and likely will) add more checks in the future, such as
ensuring the presence of documentation.

EDIT: Added a filename to rule name check, which caught ~6-7 errant rules. Fixes #321.


lukesneeringer:
Note: The CI failure at the moment is the meta-linter doing its job. Sending a separate PR to fix the bugs it caught. :-)

PR #325, opened 2019-11-08
Rewrite lint configuration.

mingzhi:
The configurations work in two steps:
First, we check if a configuration is matched against the given path.
1. If none is matched, the rule is enabled by default;
2. Otherwise, we apply the matched configuration against the rule:
(a) if the given rule is in the disabled list, the rule is disabled;
(b) if the given rule is in the enabled list, the rule is enabled;
(c) if the given rule is in neither list, the rule remains the state determined by the previous matched configuration.

### DIFF
--- a/cmd/api-linter/main.go
+++ b/cmd/api-linter/main.go
@@ -38,10 +38,10 @@ func runCLI(args []string) error {
 	return c.lint(globalRules, globalConfigs)
 }
 
+// Enable all rules by default.
+// Once we have rules other than core,
+// we will determine if we want to
+// change this policy.
 func defaultConfigs() lint.Configs {
-	return lint.Configs{
-		lint.Config{
-			IncludedPaths: []string{"**"},
-		},
-	}
+	return lint.Configs{}
 }
--- a/cmd/api-linter/main.go
+++ b/cmd/api-linter/main.go
@@ -41,10 +41,7 @@ func runCLI(args []string) error {
 func defaultConfigs() lint.Configs {
 	return lint.Configs{
 		lint.Config{
-			IncludedPaths: []string{"**/*.proto"},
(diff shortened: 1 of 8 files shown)


cmd/api-linter/main.go
```diff
@@ -38,13 +38,10 @@ func runCLI(args []string) error {
 	return c.lint(globalRules, globalConfigs)
 }
 
+// Enable all rules by default.
+// Once we have rules other than core,
+// we will determine if we want to
```
lukesneeringer: We 100% most certainly will, but this is still fine for now.

lint/config.go
```diff
@@ -109,55 +89,42 @@ func ReadConfigsYAML(f io.Reader) (Configs, error) {
...
 		}
 	}
-	for _, pattern := range c.IncludedPaths {
-		if matched, err := doublestar.Match(pattern, path); matched && err == nil {
+	return false
+}
+
+func matchRule(rule string, rulePrefixes ...string) bool {
+	for _, prefix := range rulePrefixes {
+		if prefix == "all" || prefix == rule || strings.HasPrefix(rule, prefix+"::") {
```
lukesneeringer: It need not block this PR, but can we have the logic be "matches any segment" rather than "matches prefix"?

For example, I would like someone to be able to write `(-- api-linter: http-uri=disabled --)` at the top of a file and disable `http-uri` checks from _any_ AIP.

mingzhi: I added segment matching in the configs, but not in the file comments, which we will address it in another PR. Created a issue #326. Thanks!

cmd/api-linter/main.go
mingzhi: Acknowledged.

PR #329, opened 2019-11-09
[refactor] Miscellaneous cleanup.

mingzhi:
Removed unused functions and trimmed codes.


lint/rule_name.go
```diff
@@ -31,19 +31,12 @@ const nameSeparator string = "::"
...
+		return "core"
 	}
-
-	// Sanity check: If the group does not exist, complain.
-	if group == "" {
-		panic("Invalid AIP; no available group.")
-	}
-
-	return group
+	panic(fmt.Sprintf("Invalid AIP number %d: no available group.", aip))
```
lukesneeringer: Is this okay? I usually get lint warnings when the last statement of a function is not a `return` statement.

mingzhi: Return or panic seems very reasonable to me. I run `golangci-lint` and `golint`, and didn't find any warning about this change.

PR #346, opened 2019-11-15
[feat] Add location method for the HTTP rule.

lukesneeringer:
Sadly, it appears this is the most specific we are able to get
at this time.

### DIFF
--- a/locations/method_locations_test.go
+++ b/locations/method_locations_test.go
@@ -29,6 +29,7 @@ func TestMethodRequestType(t *testing.T) {
 		message Book {}
 	`)
 	loc := MethodRequestType(f.GetServices()[0].GetMethods()[0])
+	// Three character span: line, start column, end column.
 	if diff := cmp.Diff(loc.GetSpan(), []int32{3, 14, 28}); diff != "" {
 		t.Errorf(diff)
 	}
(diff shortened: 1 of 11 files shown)


locations/method_locations_test.go
```diff
@@ -47,3 +47,22 @@ func TestMethodResponseType(t *testing.T) {
...
+		    option (google.api.http) = {
+		      get: "/v1/{name=publishers/*/books/*}"
+		    };
+		  }
+		}
+		message GetBookRequest{}
+		message Book {}
+	`)
+	loc := MethodHTTPRule(f.GetServices()[0].GetMethods()[0])
+	if diff := cmp.Diff(loc.GetSpan(), []int32{5, 4, 7, 6}); diff != "" {
```
jgeewax: Can we clarify what the 5,4,7,6 is here? I suspect someone reading this might go "OK you check that the output is a magic number..."

mingzhi: +1, maybe something like:
```golang
wantTextSpan := []int{
  5, // start line
  4, // start column
  7, // end line
  6, // end column
}
```

lukesneeringer: Done.

PR #357, opened 2019-11-19
[refactor] Meta-linter redesign.

lukesneeringer:
This PR makes the meta-linter significantly more modular
and extensible, as I find I want to add several more checks to it
and extending it now is repetitive.

We now walk the tree **once** and call each individual checker
with the AIP number and the rule leaf name. The checker does its
single check and complains if needed.

The meta-linter now groups the output by rule (rather than by
checker) and provides a brief summary at the end.

Eventual todo: Add ANSI colors. Outside the scope of this PR. :-)


lukesneeringer:
```
98 rules scanned: 98 passed, 0 failed.
```

.circleci/meta-linter/main.go
```diff
@@ -18,21 +18,104 @@ import (
...
+			func(p string, _ os.FileInfo) bool { return strings.HasSuffix(p, "_test.go") },
+			func(p string, _ os.FileInfo) bool { return aipIndex.MatchString(p) },
+		} {
+			if exempt(path, info) {
+				return nil
+			}
+		}
+
+		// This represents a rule. Get the AIP and rule name from the path.
+		match := ruleFile.FindStringSubmatch(path)
```
mingzhi: Check if `len(match) == 3`.

.circleci/meta-linter/main.go
```diff
@@ -18,21 +18,104 @@ import (
...
+			return nil
+		}
+		name := strings.ReplaceAll(match[2], "_", "-")
+		token := fmt.Sprintf("%04d-%s", aip, name)
+
+		// Run each checker and run up the list of errors.
+		for _, checker := range []func(aip int, name string) []error{
+			checkRuleDocumented,
+			checkRuleName,
+			checkRuleRegistered,
```
mingzhi: Consider to create a variable outside the for loop or main function to increase its visibility.

.circleci/meta-linter/main.go
```diff
@@ -18,21 +18,104 @@ import (
...
 
 func main() {
 	errors := []error{}
 
-	// Run each checker and run up the list of errors.
-	errors = append(errors, checkRulesRegistered()...)
-	errors = append(errors, ruleFilename()...)
-	errors = append(errors, checkRulesDocumented()...)
+	// Keep track of rules processed, and which ones have one or more failures.
+	pass := stringset.New()
```
mingzhi: Consider to give `pass` and `fail` more informative names, like `passedRules` and `failedRules`.

.circleci/meta-linter/main.go
```diff
@@ -18,21 +18,104 @@ import (
...
+		aip, err := strconv.Atoi(match[1])
+		if err != nil {
+			errors = append(errors, err)
+			return nil
+		}
+		name := strings.ReplaceAll(match[2], "_", "-")
+		token := fmt.Sprintf("%04d-%s", aip, name)
+
+		// Run each checker and run up the list of errors.
+		for _, checker := range []func(aip int, name string) []error{
```
mingzhi: Can we return an error immediately if found in a checking, to increase code readability? Otherwise, we need to keep tracking the error array mentally.

.circleci/meta-linter/rule_registered.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	if !strings.Contains(string(contents), ruleVar) {
+		errata = append(errata, fmt.Errorf("rule %q for AIP-%d not registered in the AIP's AllRules", name, aip))
+	}
+
+	// Ensure that the AIP itself is registered in `rules/rules.go`.
+	contents, err = ioutil.ReadFile("rules/rules.go")
+	if err != nil {
+		errata = append(errata, err)
+		return errata
```
mingzhi: I think we should return an error immediately if one is found. Otherwise, it is hard to keep tracking when we must return and when not.

.circleci/meta-linter/rule_registered.go
```diff
@@ -0,0 +1,68 @@
...
+	ruleMatch := ruleRegexp.FindStringSubmatch(string(contents))
+	if len(ruleMatch) == 0 {
+		return []error{fmt.Errorf("no rule found: %s", path)}
+	}
+
+	// Some errors are now non-fatal; start a running tab.
+	errata := []error{}
+
+	// Ensure this rule is registered within its AIP module file.
+	ruleVar := ruleMatch[1]
```
mingzhi: Check if `len(ruleMatch) > 1`.

.circleci/meta-linter/main.go
lukesneeringer: That feels like a change detector condition. The length of the match is guaranteed to be 3 if there is a match at all.

mingzhi: Fair enough. Thanks!

.circleci/meta-linter/main.go
lukesneeringer: Done.

.circleci/meta-linter/main.go
lukesneeringer: Sure. I do not think it matters much, but I have no reason not to do it.

lukesneeringer: Actually, it turns out I do not want to do this. The current code handles the boundary conditions (e.g. unexpected filenames) better. I _can_ do what you suggest, by using some kind of global tracking failure bit and other scaffolding, but it turns out not to be an improvement. :(

.circleci/meta-linter/rule_registered.go
lukesneeringer: I disagree. I am going with the Golang testing philosophy -- provide all the errors we are able to at once, so people do not have to thrash.

.circleci/meta-linter/rule_registered.go
lukesneeringer: This is guaranteed by the regex engine.

.circleci/meta-linter/main.go
lukesneeringer: Done.

.circleci/meta-linter/main.go
mingzhi: Fair enough!

.circleci/meta-linter/rule_registered.go
mingzhi: Okay, just a nitpicking. Thanks!

.circleci/meta-linter/rule_registered.go
mingzhi: I see. It slipped my mind :) Thanks!

PR #358, opened 2019-11-19
[AIP-191] Require `option java_multiple_files = true;`.

lukesneeringer:
~This should not be merged until googleapis/aip#360 is.~ That PR is merged.

### DIFF
--- a/rules/aip0191/java_multiple_files_test.go
+++ b/rules/aip0191/java_multiple_files_test.go
@@ -1,0 +1,42 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


rules/aip0191/java_multiple_files_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+package aip0191
+
+import (
+	"testing"
+
+	"github.com/googleapis/api-linter/rules/internal/testutils"
+)
+
+func TestJavaMultipleFiles(t *testing.T) {
```
mingzhi: Add a test with no package, a.k.a `f.GetPackage() == ""`.

lukesneeringer: That caught a bug -- thanks.

lukesneeringer: Done.

PR #366, opened 2019-11-20
[AIP-156] Prohibit List, Create, and Delete on singletons.

lukesneeringer:


rules/aip0156/forbidden_methods.go
```diff
@@ -0,0 +1,56 @@
...
+		//   publishers/*/books/* -- not a singleton, many books
+		//   publishers/*/settings -- a singleton; one settings object per book
+		for _, http := range utils.GetHTTPRules(m) {
+			if name, ok := http.GetVariables()["name"]; ok {
+				if !strings.HasSuffix(name, "*") {
+					return true
+				}
+			}
+		}
+		return false
```
mingzhi: Can you create a function for it? I assume that we are going to add more rules related to singleton resources.

lukesneeringer: I do not intend to add more rules at this time, actually. If that changes, I would be happy to refactor it into a function.

PR #368, opened 2019-11-20
[AIP-191] File option consistency.

lukesneeringer:
This rule checks that options like `csharp_namespace`, `ruby_package`,
and so on are applied consistently throughout an API.

~Uses #365 as a base (for `hasPackage`).~

### DIFF
--- a/docs/rules/0191/file-option-consistency.md
+++ b/docs/rules/0191/file-option-consistency.md
@@ -23,6 +23,7 @@ The following annotations are included:
 
 - `csharp_namespace`
 - `go_package`
+- `java_multiple_files`
 - `java_package`
 - `php_class_prefix`
 - `php_metadata_namespace`
--- a/docs/rules/0191/file-option-consistency.md
+++ b/docs/rules/0191/file-option-consistency.md
@@ -8,7 +8,7 @@ redirect_from:
   - /0191/file-option-consistency
 ---
 
-# Java package annotation
+# File option consistency
 
 This rule enforces that every proto file for a public API surface sets file
 packaging options consistently, as mandated in [AIP-191][].
--- a/docs/rules/0191/file-option-consistency.md
+++ b/docs/rules/0191/file-option-consistency.md
@@ -1,0 +1,110 @@
+---
+rule:
+  aip: 191
+  name: [core, '0191', file-option-consistency]
+  summary: All proto files must set file packaging options consistently.
+permalink: /191/file-option-consistency
+redirect_from:
+  - /0191/file-option-consistency
+---
+
+# Java package annotation
+
+This rule enforces that every proto file for a public API surface sets file
+packaging options consistently, as mandated in [AIP-191][].
+
+## Details
+
+This rule looks at each proto file, and reads any files that it imports that
+are in the same proto package. It iterates over the file packaging options in
+each one and complains if they are inconsistent.
+
+The following annotations are included:
+
+- `csharp_namespace`
+- `go_package`
+- `java_package`
+- `php_class_prefix`
+- `php_metadata_namespace`
+- `php_namespace`
+- `objc_class_prefix`
+- `ruby_package`
+- `swift_prefix`
+
+## Examples
+
+**Incorrect** code for this rule:
+
+In `foo.proto`:
+
+```proto
+// Incorrect.
+syntax = "proto3";
+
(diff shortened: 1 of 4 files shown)


docs/rules/0191/file-option-consistency.md
```diff
@@ -0,0 +1,110 @@
...
+
+The following annotations are included:
+
+- `csharp_namespace`
+- `go_package`
+- `java_package`
+- `php_class_prefix`
+- `php_metadata_namespace`
+- `php_namespace`
+- `objc_class_prefix`
```
bufdev: Are you 100% sure on `objc_class_prefix` and `php_metadata_namespace`? I think I concluded these should be treated more like `java_outer_classname` but I'm not Objective-C or PHP expert.

Also you might want to include `java_multiple_files` here just for completeness, even though API-191 specifies it should always be true anyways.

lukesneeringer: > Are you 100% sure on objc_class_prefix and php_metadata_namespace? I think I concluded these should be treated more like java_outer_classname but I'm not Objective-C or PHP expert.

I am pretty sure on `objc_class_prefix`. Objective-C has a custom of including a 3-4 letter prefix on everything (2-letter prefixes are reserved by Apple) because Objective-C's namespacing is virtually non-existent. I can not think of a good reason for an API to not use the same prefix throughout.

I am much less sure about `php_metadata_namespace`, but I can always remove it if it turns out that I am wrong. It _is_ consistent in our internal code though (unlike, say, `java_outer_classname`, which is different per-proto).

> Also you might want to include `java_multiple_files` here just for completeness, even though API-191 specifies it should always be true anyways.

That seems duplicative with #358.

bufdev: `objc_class_prefix` is probably right, I think I decided against it once and then forgot to re-add.

Even if it's very technically duplicative, I'd expect it to be in the list, just as a reader - I actually wish each of these were a separate configurable option, I've found consumers want that, but I don't think that's in line with the AIPs.

lukesneeringer: I considered making these separate rules. I could probably be talked into it (it would avoid the need to use the `reflect.ValueOf` stuff).

That said, I am about to add rules looking at the specifics of each one (e.g. `csharp_namespace` should use dot-separated PascalCase), and this one really is just for "these need to match". I find that this is almost always just an accidental omission -- but the accidental omission happens all the time. I have never seen anyone _want_ inconsistency within a package.

bufdev: Really my goal here is to match up https://buf.build/docs/lint-checkers#package_affinity and this (which probably requires Buf adding `objc_class_prefix`.

Also for reference on specifics for each one, this was what Prototool has https://github.com/uber/prototool/tree/dev/style#file-options in Buf, I haven't added anything yet, but this might come down the line, it was mostly a timing issue

lukesneeringer: Okay. I can add `java_multiple_files`. I actually do see a benefit now that I think about it -- someone might disable `java-multiple-files` inconsistently, and this will tell them there is a problem.

lukesneeringer: Actually, adding `java_multiple_files` is somewhat annoying with the approach I took, since it is a bool.

bufdev: This is how I got around that annoying problem https://github.com/bufbuild/buf/blob/master/internal/buf/bufcheck/buflint/internal/internal.go#L350

lukesneeringer: Okay, I have added `java_multiple_files` now. :-)

docs/rules/0191/file-option-consistency.md
```diff
@@ -0,0 +1,110 @@
...
+rule:
+  aip: 191
+  name: [core, '0191', file-option-consistency]
+  summary: All proto files must set file packaging options consistently.
+permalink: /191/file-option-consistency
+redirect_from:
+  - /0191/file-option-consistency
+---
+
+# Java package annotation
```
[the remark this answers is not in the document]
lukesneeringer: Yes.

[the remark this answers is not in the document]
lukesneeringer: Done.

PR #369, opened 2019-11-21
[AIP-191] Check csharp_namespace for idiomaticity.

lukesneeringer:


lukesneeringer:
This will need a slight change after #370 goes in. Fine to approve both (tests will just start failing after it is merged).

docs/rules/0191/csharp-namespace.md
```diff
@@ -0,0 +1,72 @@
...
+`option csharp_namespace`, that it uses language idiomatic conventions, as
+mandated in [AIP-191][].
+
+## Details
+
+This rule looks at each proto file, and complains if the `csharp_namespace`
+file annotation uses anything other than upper camel case, or includes
+characters other than letters, numbers, and `.`.
+
+## Examples
```
jskeet: Could we have another example with a beta? This is a common error - using V1beta instead of V1Beta.

rules/aip0191/csharp_namespace.go
```diff
@@ -0,0 +1,67 @@
...
+					Descriptor: f,
+					Location:   locations.FileCsharpNamespace(f),
+				}}
+			}
+
+			// Check that upper camel case is used.
+			upperCamel := []string{}
+			for _, segment := range strings.Split(ns, ".") {
+				wantSegment := csharpVersionRegexp.ReplaceAllStringFunc(
+					strcase.UpperCamelCase(segment),
```
jskeet: Will this be okay with Google.Cloud.AutoML.V1 for example?

rules/aip0191/csharp_namespace_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{"Valid", "Google.Example.V1", testutils.Problems{}},
+		{"ValidBeta", "Google.Example.V1Beta1", testutils.Problems{}},
+		{"InvalidBadChars", "Google:Example:V1", testutils.Problems{{Message: "Invalid characters"}}},
+		{"Invalid", "google.example.v1", testutils.Problems{{
+			Suggestion: fmt.Sprintf("option csharp_namespace = %q;", "Google.Example.V1"),
+		}}},
+		{"InvalidVersion", "Google.Example.v1", testutils.Problems{{
+			Suggestion: fmt.Sprintf("option csharp_namespace = %q;", "Google.Example.V1"),
+		}}},
+		{"InvalidBeta", "Google.Example.V1beta1", testutils.Problems{{
```
jskeet: Aha! Hooray :)

docs/rules/0191/csharp-namespace.md
lukesneeringer: Done.

rules/aip0191/csharp_namespace.go
lukesneeringer: Discussed in video chat. Sadly no, it wants `AutoMl`. Most likely we merge this as-is and they have to disable it, because I can not think of a good way around it.

PR #370, opened 2019-11-21
[refactor] Make locations.pathLocation better.

lukesneeringer:
This refactors `locations.pathLocation` to take the baseline
descriptor and the path suffix. This significantly reduces
the amount of work being done in the public methods in the
package.

### DIFF
--- a/locations/locations.go
+++ b/locations/locations.go
@@ -27,9 +27,14 @@ import (
 	"github.com/jhump/protoreflect/desc"
 )
 
-// pathLocation returns the precise location within a file for a given path.
-func pathLocation(f *desc.FileDescriptor, path []int32) *dpb.SourceCodeInfo_Location {
-	return sourceInfoRegistry.sourceInfo(f).findLocation(path)
+// pathLocation returns the precise location for a given descriptor and path.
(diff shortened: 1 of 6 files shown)


locations/locations.go
```diff
@@ -27,9 +27,14 @@ import (
 	"github.com/jhump/protoreflect/desc"
 )
 
-// pathLocation returns the precise location within a file for a given path.
-func pathLocation(f *desc.FileDescriptor, path []int32) *dpb.SourceCodeInfo_Location {
-	return sourceInfoRegistry.sourceInfo(f).findLocation(path)
+// pathLocation returns the precise location for a given descriptor and path.
+// It combines the path of the descriptor itself with any path provided appended.
+func pathLocation(d desc.Descriptor, path ...int) *dpb.SourceCodeInfo_Location {
+	fullPath := d.GetSourceInfo().GetPath()
```
mingzhi: Check if `d.GetSourceInfo() == nil`

lukesneeringer: Believe it or not, unnecessary! :-) All of the `FileLocation` tests (`FileDescriptor` always returns nil) pass and do the correct thing.

PR #375, opened 2019-11-21
[AIP-191] Enforce php_namespace idiomaticity.

lukesneeringer:


locations/file_locations.go
```diff
@@ -43,3 +43,12 @@ func FilePackage(f *desc.FileDescriptor) *dpb.SourceCodeInfo_Location {
 func FileCsharpNamespace(f *desc.FileDescriptor) *dpb.SourceCodeInfo_Location {
 	return pathLocation(f, 8, 37) // 8 == options, 37 == csharp_namespace
 }
+
+// FilePhpNamespace returns the location of the php_namespace file option
+// in a file descriptor.
+//
+// If the location can not be found (for example, because there is no
+// csharp_namespace option), it returns nil.
```
dwsupplee:
```suggestion
// php_namespace option), it returns nil.
```

docs/rules/0191/php-namespace.md
```diff
@@ -0,0 +1,107 @@
...
+package google.example.v1beta1;
+
+option php_namespace = "Google\\Example\\V1beta1";
+```
+
+## Known issues
+
+This rule will improperly complain if it encounters an acronym. For example, it
+will complain about `Google\Cloud\AutoML\V1`, preferring `AutoMl`. This lint
+rule **should** be disabled in this case.
```
bufdev: I would personally disagree with this. I think special casing for abbreviations introduces a human component into deciding these values, and in an optimal world, users don't have to set these values at all and the tooling does it automatically for them. I think making a human decide what is and isn't an abbreviation introduces issues, and it would be better just to have the abbreviation seem a little out of place with the option value.

docs/rules/0191/php-namespace.md
```diff
@@ -0,0 +1,107 @@
...
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+syntax = "proto3";
+
+package google.example.v1;
+
+option php_namespace = "google\\example\\v1";  // Should be UpperCamelCase.
```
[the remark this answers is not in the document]
lukesneeringer: I am less concerned about this because linter.aip.dev (the docs site) does a really good job of showing it as a wrong example.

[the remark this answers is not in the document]
bufdev: Yea it's a nit, but if the docs are going to be here, might as well make it easy to read right? Ha :-)

docs/rules/0191/php-namespace.md
lukesneeringer: This might be true. Certainly the rule **may** be disabled, though (we have existing APIs that use this case). Not going to change it here since there are two equivalent files, but in principle I am open to rounding down to may.

bufdev: I would round down to may at the least, although I'd go even further and say that abbreviations should not be an exception at all. Users want to trust the yes/no decision a linter makes and not have to make manual interventions, especially at organization-scale.

lukesneeringer: Will follow-up in another PR.

PR #377, opened 2019-11-21
[AIP-191] Enforce common file layout mistakes.

lukesneeringer:


rules/aip0191/file_layout.go
```diff
@@ -0,0 +1,70 @@
...
+		// Verify that services precede messages.
+		if len(f.GetMessageTypes()) > 0 {
+			firstMessage := f.GetMessageTypes()[0]
+			for _, service := range f.GetServices() {
+				if isAfter(firstMessage, service) {
+					problems = append(problems, lint.Problem{
+						Message:    "Services should precede all messages.",
+						Descriptor: service,
+					})
+				}
```
mingzhi: Do we want to break here, like the case below?

lukesneeringer: I considered it and decided against it in this case. The number of services is smaller than the number of messages, and it really does make sense to warn on them all.

The fundamental difference is that with this warning, the _service_ is in the wrong place. With the other one, it might be that the enum is in the wrong place, and it might be messages, depending on the situation, and flagging every message as a way to indicate the need to move an enum (or vice versa) is annoying.

mingzhi: Thanks!

PR #380, opened 2019-11-22
[AIP-132] Check unknown fields on list responses.

lukesneeringer:


docs/rules/0132/response-unknown-fields.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# List methods: Unknown fields (Response)
+
+This rule enforces that all `List` standard methods do not have unexpected
+fields, as mandated in [AIP-132][].
+
+## Details
+
+This rule looks at any message matching `List*Request` and complains if it
```
mingzhi: Is "List*Request" a typo?

rules/aip0132/response_unknown_fields.go
```diff
@@ -0,0 +1,54 @@
...
+// The resource itself is not included here, but also permitted.
+// This is covered in code in the rule itself.
+var respAllowedFields = stringset.New(
+	"next_page_token",       // AIP-158
+	"total_size",            // AIP-132
+	"unreachable",           // AIP-217
+	"unreachable_locations", // Wrong, but a separate AIP-217 rule catches it.
+)
+
+var responseUnknownFields = &lint.MessageRule{
```
mingzhi: Can you use a `FieldRule`? You can add a `OnlyIf` function like:
```golang
func (f *desc.FieldDescriptor) bool {
  return isListResponseMessage(f.GetOwner())
}
```

docs/rules/0132/response-unknown-fields.md
lukesneeringer: Yes. Thanks.

```suggestion
This rule looks at any message matching `List*Response` and complains if it
```

rules/aip0132/response_unknown_fields.go
lukesneeringer: I can, but does it really matter?

lukesneeringer: I asked this before reading your refactor PR. In the context of _that_ PR, it obviously does matter.

mingzhi: We can leave it as-is. Will refactor it later.

lukesneeringer: Actually I am happy to change it.

lukesneeringer: Done.

PR #381, opened 2019-11-22
[AIP-217] Enforce the unreachable field.

lukesneeringer:


rules/aip0217/synonyms_test.go
```diff
@@ -0,0 +1,47 @@
...
+		problems  testutils.Problems
+	}{
+		{"unreachable", testutils.Problems{}},
+		{"unreachable_locations", testutils.Problems{{Suggestion: "unreachable"}}},
+	} {
+		t.Run(strcase.UpperCamelCase(test.FieldName), func(t *testing.T) {
+			f := testutils.ParseProto3Tmpl(t, `
+				message ListBooksResponse {
+					repeated Book books = 1;
+					string next_page_token = 2;
```
mingzhi: Should we test the `OnlyIf` to gain 100% coverage?

docs/rules/0217/unreachable-field-type.md
```diff
@@ -0,0 +1,73 @@
...
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message ListBooksResponse {
+  repeated Book books = 1;
+  string next_page_token = 2;
+  string unreachable_locations = 3;  // Should be repeated.
```
mingzhi: Should it be "unreachable" field name?

rules/aip0217/unreachable_field_type.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var unreachableFieldType = &lint.FieldRule{
+	Name: lint.NewRuleName(217, "unreachable-field-type"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		return f.GetName() == "unreachable"
```
mingzhi: Do we need to ensure that the message contains"next_page_token"?

rules/aip0217/synonyms.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var synSet = stringset.New(
+	"unreachable_locations",
+)
+
+var synonyms = &lint.MessageRule{
```
mingzhi: Can you use a `FieldRule`? We add a `OnlyIf` function like
```
func(f *desc.FieldDescriptor) bool {
  return f.GetOwner().FindFieldByName("next_page_token") != nil
}
```

rules/aip0217/unreachable_field_type.go
lukesneeringer: Nah.

docs/rules/0217/unreachable-field-type.md
lukesneeringer: Good catch, thanks.

lukesneeringer: Done.

rules/aip0217/synonyms_test.go
lukesneeringer: Done.

rules/aip0217/synonyms.go
lukesneeringer: Done.

PR #383, opened 2019-11-24
[Refactor] Split 0131::request-name-field into two rules.

mingzhi:
It separates 0131::request-name-field to two rules:
1. Check if a Get request message has a name field -- users disable it at the message level;
2. Check if the name field type is string -- users disable it at the field level.

There are some code and documentation duplication, but it is much clearer about how to disable the rules. And each rule is smaller and easier to understand and test.


docs/rules/0131/request-has-name-field.md
```diff
@@ -1,11 +1,11 @@
 ---
 rule:
   aip: 131
-  name: [core, '0131', request-name-field]
+  name: [core, '0131', request-has-name-field]
```
lukesneeringer: How about `request-required-fields` for the presence check, and `request-name-field` for the type check?
This will need to be done in AIPs 132 through 135 as well for consistency.

mingzhi: Done. PTAL. Thank you!

PR #386, opened 2019-11-24
ProtoRule accepts a Descriptor instead of a FileDescriptor.

mingzhi:


mingzhi:
Changed `ProtoRule` to accept a `Descriptor` instead of a `FileDescriptor`. It has following benefits:
1. It simplifies the rule logic. An implemented rule just needs to check if a `Descriptor` is the type it wants to check, and applies the checking function on it -- it doesn't need to walk over the file.
2. It allows us to pre-filter descriptors before invoking a rule. For example, we can now check if a rule is disabled in a descriptor's comments before invoking it.

lint/lint.go
```diff
@@ -89,7 +93,7 @@ func (l *Linter) lintFileDescriptor(fd *desc.FileDescriptor) (Response, error) {
 	return resp, err
 }
 
-func (l *Linter) runAndRecoverFromPanics(rule ProtoRule, fd *desc.FileDescriptor) (probs []Problem, err error) {
```
lukesneeringer: Why are we making this change? It seems like it _should_ only accept a FileDescriptor.

lint/rule.go
```diff
@@ -55,11 +55,13 @@ func (r *FileRule) GetName() RuleName {
...
 
-// Lint forwards the FileDescriptor to the LintFile method defined on the
-// FileRule.
-func (r *FileRule) Lint(fd *desc.FileDescriptor) []Problem {
-	if r.OnlyIf == nil || r.OnlyIf(fd) {
-		return r.LintFile(fd)
+// Lint accepts a Descriptor and applies LintFile on it
+// only if the descriptor is a FileDescriptor.
+func (r *FileRule) Lint(d desc.Descriptor) []Problem {
+	if f, ok := d.(*desc.FileDescriptor); ok {
```
lukesneeringer: We do not want to do this. This is a great way to get silent failures.

lukesneeringer: Nevermind. I understand the PR better now.

lint/lint.go
lukesneeringer: Nevermind. I understand the PR better now.

mingzhi:
> Okay, reading this in more detail, I have changed my opinion, and this seems fine.

I am also eyeing on linting OpenAPI -- I guess other companies would be interested in adding AIPs for OpenAPI. So we want to decouple jhump's descriptors from the `lint` library -- this PR is the first step. 

In other words, the `lint` will accepts rules written on jhump's descriptors and rules written on OpenAPI descriptors -- the rule itself needs to assert if the type of descriptor is what it wants to check.

More PRs are coming.

lukesneeringer:
My own opinion is that linting OpenAPI _in this linter_ should probably be out of scope. It would be much less complex to simply write another linter than to try to use this one to serve both purposes.

mingzhi:
> My own opinion is that linting OpenAPI _in this linter_ should probably be out of scope. It would be much less complex to simply write another linter than to try to use this one to serve both purposes.

Nevertheless, I would still like to decouple jhump's desc library from the `lint` package -- almost done. It makes `lint` more lightweight and easier for testing.

mingzhi:
Close it. It is kind of error-prone.

PR #387, opened 2019-11-25
Add DoNotImplement to the ProtoRule interface.

mingzhi:
Restrict implementation of `ProtoRule` to this repo until we are ready to release linter version 1.


mingzhi:
> I would prefer not do this.

Fair enough -- I am on the fence :)

PR #388, opened 2019-11-25
Decouple descriptor rules from the lint package.

mingzhi:


lukesneeringer:
We discussed this internally and decided not to do it.

PR #392, opened 2019-11-26
[AIP-140] Prohibit reserved words.

lukesneeringer:
This rule currently checks Java, JavaScript, and Python 3.
I will add a couple additional languages in a follow-up PR.

### DIFF
--- a/docs/rules/0140/reserved-words.md
+++ b/docs/rules/0140/reserved-words.md
@@ -8,7 +8,7 @@ redirect_from:
   - /0140/reserved-words
 ---
 
-# Field names: Prepositions
+# Field names: Reserved words
 
 This rule enforces that field names are not reserved words, as mandated in
 [AIP-140][].
--- a/docs/rules/0140/reserved-words.md
+++ b/docs/rules/0140/reserved-words.md
@@ -1,0 +1,70 @@
+---
+rule:
+  aip: 140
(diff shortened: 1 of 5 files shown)


docs/rules/0140/reserved-words.md
```diff
@@ -0,0 +1,70 @@
...
+rule:
+  aip: 140
+  name: [core, '0140', reserved-words]
+  summary: Field names must not be reserved words.
+permalink: /140/reserved-words
+redirect_from:
+  - /0140/reserved-words
+---
+
+# Field names: Prepositions
```
mingzhi: Copy-paste typos?

lukesneeringer:
```suggestion
# Field names: Reserved words
```

bufdev:
This gets to be a mess (to cover all languages, you'll have to hit like 60 reserved words, and you also should have automated testing heh), but I support it for what it's worth.

A couple that are usually missed:

- `internal` - If a package name includes a component named `internal`, this effectively makes it so that Golang generated types are unreachable, ie `foo.internal.bar`.
- `std` - If a package name includes a component named `std`, this effectively makes it so that C++ generated types are unreachable, i.e. `std.bar`. This could be fixed in `protoc`, however (maybe it already has been).

PR #393, opened 2019-11-26
[refactor] Improve utils.GetTypeName

lukesneeringer:
Having different methods is duplicative. One is better.

### DIFF
--- a/rules/internal/utils/type_name.go
+++ b/rules/internal/utils/type_name.go
@@ -22,6 +22,8 @@ import (
 
 // GetTypeName returns the name of the type of the field, as a string,
 // regardless
+//
+// TODO: Add support for map types.
 func GetTypeName(f *desc.FieldDescriptor) string {
 	if m := f.GetMessageType(); m != nil {
 		return m.GetFullyQualifiedName()
--- a/rules/internal/utils/type_name.go
+++ b/rules/internal/utils/type_name.go
@@ -17,23 +17,17 @@ package utils
 import (
 	"strings"
 
-	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/jhump/protoreflect/desc"
 )
 
-type hasGetType interface {
-	GetType() dpb.FieldDescriptorProto_Type
-}
-
-// GetScalarTypeName returns the name of the type of the field, as a string.
-func GetScalarTypeName(t hasGetType) string {
-	return strings.ToLower(t.GetType().String()[len("TYPE_"):])
-}
-
-// GetMessageTypeName returns the name of the type of the field, as a string.
-func GetMessageTypeName(f *desc.FieldDescriptor) string {
-	if messageType := f.GetMessageType(); messageType != nil {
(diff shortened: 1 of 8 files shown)


rules/internal/utils/type_name.go
```diff
@@ -17,23 +17,17 @@ package utils
...
-	return strings.ToLower(t.GetType().String()[len("TYPE_"):])
-}
-
-// GetMessageTypeName returns the name of the type of the field, as a string.
-func GetMessageTypeName(f *desc.FieldDescriptor) string {
-	if messageType := f.GetMessageType(); messageType != nil {
-		return messageType.GetFullyQualifiedName()
+// GetTypeName returns the name of the type of the field, as a string,
+// regardless
+func GetTypeName(f *desc.FieldDescriptor) string {
```
mingzhi: Add a support for map type too.

https://godoc.org/github.com/jhump/protoreflect/desc#FieldDescriptor.GetMapKeyType
https://godoc.org/github.com/jhump/protoreflect/desc#FieldDescriptor.GetMapValueType

rules/internal/utils/type_name.go
```diff
@@ -17,23 +17,17 @@ package utils
...
+	if m := f.GetMessageType(); m != nil {
+		return m.GetFullyQualifiedName()
```
mingzhi: Are you sure to return the fully qualified name?
For example:
```
package test;
message Book {
  Author author = 1;
}
message Author{}
```
Will it return "test.Author" if we get the type name of `author`? But we really want "Author", right?

rules/internal/utils/type_name_test.go
```diff
@@ -20,43 +20,22 @@ import (
 	"github.com/googleapis/api-linter/rules/internal/testutils"
 )
 
-func TestGetScalarTypeName(t *testing.T) {
-	for _, ty := range []string{"int32", "int64", "string", "bytes", "float", "double"} {
+func TestGetTypeName(t *testing.T) {
+	for _, ty := range []string{"int32", "int64", "string", "bytes", "google.protobuf.Timestamp", "Format"} {
 		t.Run(ty, func(t *testing.T) {
 			file := testutils.ParseProto3Tmpl(t, `
+				import "google/protobuf/timestamp.proto";
```
mingzhi: Add "package test.v1;".

rules/internal/utils/type_name.go
lukesneeringer: I always want the fully qualified name in this case.

rules/internal/utils/type_name.go
lukesneeringer: Reasonable, but can I do it in a follow-up?

mingzhi: Sure. Can you add a TODO?

lukesneeringer:
```suggestion
//
// TODO: Add support for map types.
func GetTypeName(f *desc.FieldDescriptor) string {
```

PR #394, opened 2019-11-26
[AIP-214] Enforce `ttl` type.

lukesneeringer:
Uses #393 as a base.

### DIFF
--- a/rules/aip0214/ttl_type.go
+++ b/rules/aip0214/ttl_type.go
@@ -31,7 +31,7 @@ var ttlType = &lint.FieldRule{
 	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
 		if fieldType := utils.GetTypeName(f); fieldType != "google.protobuf.Duration" {
 			return []lint.Problem{{
-				Message:    fmt.Sprintf("ttl fields should be google.protobuf.Duration type, not %s.", fieldType),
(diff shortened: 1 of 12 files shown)


rules/aip0214/ttl_type.go
```diff
@@ -0,0 +1,42 @@
...
+
+var ttlType = &lint.FieldRule{
+	Name: lint.NewRuleName(214, "ttl-type"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		return f.GetName() == "ttl"
+	},
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		if fieldType := utils.GetTypeName(f); fieldType != "google.protobuf.Duration" {
+			return []lint.Problem{{
+				Message:    fmt.Sprintf("ttl fields should be google.protobuf.Duration type, not %s.", fieldType),
```
mingzhi: Should we add quotes around "google.protobuf.Duration"? And should `%s` be `%q`?

lukesneeringer: I have tended to prefer backtick over quotes. Doing that here.

lukesneeringer:
```suggestion
				Message:    fmt.Sprintf("ttl fields should be `google.protobuf.Duration` type, not `%s`.", fieldType),
```

PR #395, opened 2019-11-28
[chore] CI refresh

lukesneeringer:
This cleans up our CI configuration slightly.

Importantly, it also adds `-v` to the unit tests. The original
rationale for touching this file is that there is a unit test that
is flaky (but only in CI) and I want more information about the
failure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
           command: go test -mod=readonly -v ./...
   lint:
     docker:
-      - image: circleci/golang:1.13
+      - image: golang:1.13
     steps:
       - checkout
       - run:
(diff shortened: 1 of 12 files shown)


lukesneeringer:
Some notes for posterity:

- Moving the unit test run off of the weird CircleCI golang Docker image seems to have solved the flakiness problem. (In general, we should always use official golang images.)
- The previous go fmt check was just flatly wrong and also relied on goofy behavior of that specific Docker image. It also did not actually perform the check, which I know because it turns out that there was a file that failed it. The correct command is `test -z (gofmt -l .)`, which I added to the lint section.

.circleci/config.yml
```diff
@@ -35,6 +26,12 @@ jobs:
       - run:
           name: Run golangci-lint
           command: "! $(go env GOPATH)/bin/golangci-lint run ./... 2>&1 | read"
```
mingzhi: Should we change this to use `test -z` too? If `go fmt` was not working, `golangci-lint` most likely wasn't either. Thanks!

lukesneeringer: Sure enough, it was definitely not working, although `test -z` is the wrong answer in this case. Also, turns out golangci-lint has a thing to check formatting, so we no longer need the other separate step.

.golangci.yml
```diff
@@ -0,0 +1,8 @@
+---
+linters:
+  enable:
+    - goimports
+    - misspell
+  disable:
```
mingzhi: Should we disable them in the individual files? I understand that we need `noPositional` in those rules, but we want those checks in other places.

lukesneeringer: Fair point. Done.

PR #399, opened 2019-12-03
Add a installation note to Googlers.

mingzhi:
Due to some differences about protobuf dependencies, we have an internal, compiled version in Google. In other words, the Github version installed using `go get` does not work in Google3. In this PR, we advice googlers to use the internal, released binary instead.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,6 +39,8 @@ It will install `api-linter` into your local Go binary directory
 `$HOME/go/bin`. Ensure that your operating system's `PATH` contains the Go
 binary directory.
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -39,6 +39,8 @@ It will install `api-linter` into your local Go binary directory
 `$HOME/go/bin`. Ensure that your operating system's `PATH` contains the Go
 binary directory.
 
+**Note:** In Google, you should use the released binary `/google/bin/releases/api-linter/api-linter`.
```
lukesneeringer:
```suggestion
**Note:** For working in Google-internal source control, you should
use the released binary `/google/bin/releases/api-linter/api-linter`.
```

PR #401, opened 2019-12-03
[fix] Make java_outer_classname ignore directories.

lukesneeringer:
Previously it was taking the directory into account
when determining the `java_outer_classname` value,
which was wrong.

### DIFF
--- a/rules/aip0191/java_outer_classname.go
+++ b/rules/aip0191/java_outer_classname.go
@@ -29,7 +29,11 @@ var javaOuterClassname = &lint.FileRule{
 	OnlyIf: hasPackage,
 	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
 		if f.GetFileOptions().GetJavaOuterClassname() == "" {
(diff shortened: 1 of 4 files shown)


rules/aip0191/java_outer_classname.go
```diff
@@ -29,7 +29,11 @@ var javaOuterClassname = &lint.FileRule{
 	OnlyIf: hasPackage,
 	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
 		if f.GetFileOptions().GetJavaOuterClassname() == "" {
-			want := strcase.UpperCamelCase(strings.ReplaceAll(f.GetName(), ".", "_"))
+			filename := f.GetName()
+			if ix := strings.LastIndex(filename, "/"); ix != -1 {
```
mingzhi: Should we use `filepath.Base` instead?

lukesneeringer: I did not know that was a thing. Thanks!

PR #402, opened 2019-12-03
[fix] Fix Create parent false positive.

lukesneeringer:
The linter was providing a false positive for requiring
a parent field on Create methods.

### DIFF
--- a/rules/aip0133/http_uri_parent_test.go
+++ b/rules/aip0133/http_uri_parent_test.go
@@ -15,59 +15,41 @@
 package aip0133
 
 import (
-	"strings"
 	"testing"
 
-	"github.com/golang/protobuf/proto"
-	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
-	"github.com/jhump/protoreflect/desc/builder"
-	"google.golang.org/genproto/googleapis/api/annotations"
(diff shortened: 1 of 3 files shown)


rules/aip0133/http_uri_parent_test.go
```diff
@@ -15,59 +15,41 @@
...
-		msg        string
+		URI        string
+		MethodName string
+		problems   testutils.Problems
 	}{
-		{"Valid", "/v1/{parent=publishers/*/books/*}", "CreateBook", ""},
-		{"InvalidVarParent", "/v1/{book=publishers/*/books/*}", "CreateBook", "`parent` field"},
-		{"NoVarParent", "/v1/publishers/*/books/*", "CreateBook", "`parent` field"},
-		{"Irrelevant", "/v1/{book=publishers/*/books/*}", "BuildBook", ""},
+		{"Valid", "/v1/{parent=publishers/*/books/*}", "CreateBook", testutils.Problems{}},
```
mingzhi: Can you add a false positive case?

lukesneeringer: The Valid case was the false positive case. The rule and test were both broken (because the regex was messed up).

PR #403, opened 2019-12-03
[fix] Fix Get name false positive.

lukesneeringer:


rules/aip0131/http_uri_name_test.go
```diff
@@ -15,59 +15,41 @@
...
-		msg        string
+		URI        string
+		MethodName string
+		problems   testutils.Problems
 	}{
-		{"Valid", "/v1/{name=publishers/*/books/*}", "GetBook", ""},
-		{"InvalidVarName", "/v1/{book=publishers/*/books/*}", "GetBook", "`name` field"},
-		{"NoVarName", "/v1/publishers/*/books/*", "GetBook", "`name` field"},
-		{"Irrelevant", "/v1/{book=publishers/*/books/*}", "AcquireBook", ""},
+		{"Valid", "/v1/{name=publishers/*/books/*}", "GetBook", testutils.Problems{}},
```
mingzhi: Can you add a false positive case here?

lukesneeringer: The Valid case was the false positive case. The rule and test were both broken (because the regex was messed up).

PR #413, opened 2019-12-04
[AIP-215] Proto packages should end in a version.

lukesneeringer:


rules/aip0215/versioned_packages.go
```diff
@@ -0,0 +1,43 @@
...
+
+var versionedPackages = &lint.FileRule{
+	Name: lint.NewRuleName(215, "versioned-packages"),
+	OnlyIf: func(f *desc.FileDescriptor) bool {
+		return f.GetPackage() != ""
+	},
+	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
+		if p := f.GetPackage(); !version.MatchString(p) && !strings.HasSuffix(p, ".type") {
+			return []lint.Problem{{
+				Message:    "API components should be in versioned packages.",
```
mingzhi: Should we give a suggested example? Like `f.GetPackage()+".v1"`.

lukesneeringer: I would rather not, because I have no way of knowing what version is intended, and I do not want to encourage people to mark their alphas as GA (which I see as a likely outcome).

PR #418, opened 2019-12-06
[AIP-123] Add resource name pattern enforcement.

lukesneeringer:
~Uses #417 as a base.~ #417 is merged.

### DIFF
--- a/rules/aip0123/resource_pattern.go
+++ b/rules/aip0123/resource_pattern.go
@@ -15,7 +15,10 @@
 package aip0123
 
 import (
+	"fmt"
 	"strings"
+
+	"github.com/stoewer/go-strcase"
 
 	"github.com/googleapis/api-linter/locations"
 
@@ -42,8 +45,19 @@ var resourcePattern = &lint.MessageRule{
 		// not snake case.
 		for _, pattern := range resource.GetPattern() {
 			if strings.Contains(getPlainPattern(pattern), "_") {
+				want := []string{}
+				for _, token := range strings.Split(pattern, "/") {
+					if strings.HasPrefix(token, "{") && strings.HasSuffix(token, "}") {
+						want = append(want, token)
+					} else {
+						want = append(want, strcase.LowerCamelCase(token))
+					}
+				}
 				return []lint.Problem{{
-					Message:    "Resource patterns should use camel case (apart from the variable names).",
+					Message: fmt.Sprintf(
+						"Resource patterns should use camel case (apart from the variable names), such as %q.",
+						strings.Join(want, "/"),
+					),
 					Descriptor: m,
 					Location:   locations.MessageResource(m),
 				}}
--- a/rules/aip0123/resource_pattern.go
+++ b/rules/aip0123/resource_pattern.go
@@ -1,0 +1,54 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aip0123
+
+import (
+	"strings"
+
+	"github.com/googleapis/api-linter/locations"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var resourcePattern = &lint.MessageRule{
+	Name:   lint.NewRuleName(123, "resource-pattern"),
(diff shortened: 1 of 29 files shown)


rules/aip0123/resource_pattern_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/googleapis/api-linter/rules/internal/testutils"
+)
+
+func TestResourcePattern(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		Pattern  string
+		problems testutils.Problems
+	}{
+		{"Valid", `pattern: "publishers/{publisher}/books/{book}"`, testutils.Problems{}},
```
mingzhi: Should we add a valid "camelCase"?

rules/aip0123/resource_variables.go
```diff
@@ -0,0 +1,51 @@
...
+var resourceVariables = &lint.MessageRule{
+	Name:   lint.NewRuleName(123, "resource-variables"),
+	OnlyIf: hasResourceAnnotation,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		resource := utils.GetResource(m)
+		for _, pattern := range resource.GetPattern() {
+			for _, variable := range getVariables(pattern) {
+				if strings.ToLower(variable) != variable {
+					return []lint.Problem{{
+						Message:    "Variable names in patterns should use snake case.",
```
mingzhi: Similarly, can we point out the invalid variable name and give a suggestion in this message?

rules/aip0123/resource_variables.go
```diff
@@ -0,0 +1,51 @@
...
+						Descriptor: m,
+						Location:   locations.MessageResource(m),
+					}}
+				}
+				if strings.HasSuffix(variable, "_id") {
+					return []lint.Problem{{
+						Message:    "Variable names should omit the `_id` suffix.",
```
mingzhi: Similarly, can we point out the invalid variable name?

rules/aip0123/resource_pattern.go
```diff
@@ -0,0 +1,54 @@
...
+				Location:   locations.MessageResource(m),
+			}}
+		}
+
+		// Ensure that the constant segments of the pattern uses camel case,
+		// not snake case.
+		for _, pattern := range resource.GetPattern() {
+			if strings.Contains(getPlainPattern(pattern), "_") {
+				return []lint.Problem{{
+					Message:    "Resource patterns should use camel case (apart from the variable names).",
```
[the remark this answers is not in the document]
lukesneeringer: Sadly, location can not get that specific, but this is something that I do want to improve later.

rules/aip0123/resource_variables.go
lukesneeringer: Giving suggestions on annotations is really hard, unfortunately. I would like to do it eventually, but it is hard to pinpoint the location precisely enough, and hard to control for the full rule.

I considered pointing out the invalid variable name, but I (intentionally) maxed this rule out at a single problem (because of the location precision issue), and multiple offenders is possible. I _think_ the right thing to do therefore is to leave this as-is, and assume that humans will be smart enough to be able to find their underscore. ;-)

rules/aip0123/resource_pattern_test.go
lukesneeringer: Good call. Yes. Done.

rules/aip0123/resource_pattern.go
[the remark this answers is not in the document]
mingzhi: I meant -- just in the message -- we don't need to provide a `Suggestion`, like "Resource pattern 'some_pattern' should use camel case, like 'somePattern'".

rules/aip0123/resource_pattern.go
```diff
@@ -0,0 +1,68 @@
...
+		for _, pattern := range resource.GetPattern() {
+			if strings.Contains(getPlainPattern(pattern), "_") {
+				want := []string{}
+				for _, token := range strings.Split(pattern, "/") {
+					if strings.HasPrefix(token, "{") && strings.HasSuffix(token, "}") {
+						want = append(want, token)
+					} else {
+						want = append(want, strcase.LowerCamelCase(token))
+					}
+				}
```
mingzhi: Ah, I did expect this complex and difficulty -- sorry! I think we should go back to your original version -- it is better for no suggestion than an incorrect suggestion, and I think, it is no worth to go extra miles now. We can wait until we can give a specific location. Thanks!

mingzhi: did not expect

rules/aip0123/resource_variables.go
mingzhi: Given the complex and difficulty I saw above. I agree and prefer this version. Thank you!

rules/aip0123/resource_variables.go
lukesneeringer: Done.

rules/aip0123/resource_pattern.go
lukesneeringer: Actually, I changed my mind in your favor. I refactored this out into a common method that all three spots can use, that should always give the correct pattern as a suggestion.

mingzhi: Nice. Thank you!

PR #419, opened 2019-12-06
[AIP-131] Require google.api.resource_reference on Get methods.

lukesneeringer:


docs/rules/0131/request-name-reference.md
```diff
@@ -0,0 +1,64 @@
...
+## Disabling
+
+If you need to violate this rule, use a leading comment above the field.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+// (-- api-linter: core::0131::request-name-reference=disabled
+//     aip.dev/not-precedent: This is named "book" for historical reasons. --)
+message GetBookRequest {
+  string book = 1 [(google.api.field_behavior) = REQUIRED];
```
mingzhi: Wrong example?

docs/rules/0131/request-name-reference.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message GetBookRequest {
+  // The `google.api.resource_reference` annotation should also be included.
+  string book = 1 [(google.api.field_behavior) = REQUIRED];
```
mingzhi: Wrong example?

rules/aip0131/request_name_reference.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var requestNameReference = &lint.FieldRule{
+	Name: lint.NewRuleName(131, "request-name-reference"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		return isGetRequestMessage(f.GetOwner())
```
mingzhi: Do we need to check if the field name is "name"?

docs/rules/0131/request-name-reference.md
lukesneeringer: Correct example. Both annotations are expected, and it seems more user-friendly to lead toward both in the docs.

rules/aip0131/request_name_reference.go
lukesneeringer: Yes, we do. Thanks. Fixed.

docs/rules/0131/request-name-reference.md
lukesneeringer: Same as above.

rules/aip0131/request_name_reference_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"testing"
+
+	"github.com/googleapis/api-linter/rules/internal/testutils"
+)
+
+func TestRequestNameReference(t *testing.T) {
+	t.Run("Present", func(t *testing.T) {
+		f := testutils.ParseProto3String(t, `
```
mingzhi: Can you rewrite these two into table tests, like what you did in the other PR? And also can you add a test for a negative case with field name not being "name"? Thank you!

lukesneeringer: I added the not-name check (as a table test with absent). I am declining to write the "present" test as a table test because it gets sufficiently awkward to deal with the template with a multi-line annotation, and I find the existing test more readable.

mingzhi: Fair enough! Thanks!

PR #421, opened 2019-12-06
[chore] Tweak CI a bit more to see if it helps with flakes.

lukesneeringer:


lukesneeringer:
Okay I think I found the issue. We are going OOM. Sending `-p 1` makes tests slower but should avoid overflowing the RAM.

Now, why on earth building our linter overflows _four gigabytes_ is another matter...

PR #432, opened 2019-12-06
[AIP-131] Enforce a consistent method signature.

lukesneeringer:
Uses #431 as a base.

### DIFF
--- a/rules/aip0131/method_signature.go
+++ b/rules/aip0131/method_signature.go
@@ -17,7 +17,6 @@ package aip0131
 import (
 	"fmt"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/googleapis/api-linter/lint"
 	"github.com/googleapis/api-linter/locations"
 	"github.com/googleapis/api-linter/rules/internal/utils"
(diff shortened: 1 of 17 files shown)


rules/aip0131/method_signature.go
```diff
@@ -0,0 +1,55 @@
...
+				Message: fmt.Sprintf(
+					"Get methods should include `(google.api.method_signature) = %q`",
+					"name",
+				),
+				Descriptor: m,
+			}}
+		}
+
+		// Check if the signature is wrong.
+		if diff := cmp.Diff(signatures[0], []string{"name"}); diff != "" {
```
mingzhi: Should we just write `len(signature[0]) == !1 && signature[0][0] == "name"`? Using `cmp` seems to be heavy here.

lukesneeringer: I think `cmp.Diff` is the standard/preferred way to do this sort of thing.

mingzhi: I would prefer `cmp` in testing. But using it probably is overkilled here.

lukesneeringer: Okay, using `reflect.DeepEqual` instead.

PR #434, opened 2019-12-06
[AIP-133] Enforce a consistent method signature.

lukesneeringer:
Uses #431 as a base.

### DIFF
--- a/docs/rules/0133/method-signature.md
+++ b/docs/rules/0133/method-signature.md
@@ -1,0 +1,84 @@
+---
+rule:
+  aip: 133
+  name: [core, '0133', method-signature]
+  summary: |
+    Create RPCs should annotate an appropriate method signature.
+permalink: /133/method-signature
+redirect_from:
+  - /0133/method-signature
+---
+
+# Create methods: Method signature
+
+This rule enforces that all `Create` standard methods have a
+`google.api.method_signature` annotation with an appropirate value, as mandated
+in [AIP-133][].
+
+## Details
+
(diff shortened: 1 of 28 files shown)


docs/rules/0133/method-signature.md
```diff
@@ -0,0 +1,84 @@
...
+rpc CreateBook(CreateBookRequest) returns (Book) {
+  // Should be "parent,book" or "parent,book,book_id", depending on whether
+  // a "book_id" field exists.
+  option (google.api.method_signature) = "publisher,book";
+}
+```
+
+**Correct** code for this rule:
+
+If the `book_id` field does not exist:
```
mingzhi: Maybe it is better to give an example of the `Book` message without the `book_id` field.

docs/rules/0133/method-signature.md
```diff
@@ -0,0 +1,84 @@
...
+
+```proto
+// Correct.
+rpc CreateBook(CreateBookRequest) returns (Book) {
+  option (google.api.method_signature) = "parent,book";
+}
+```
+
+If the `book_id` field exists:
```
mingzhi: Maybe it is better to give an example of the `Book` message with the `book_id` field.

docs/rules/0133/method-signature.md
lukesneeringer: I thought about that, but it adds a lot of overhead. I have to define the service around the RPC, and then the message beneath it. It ended up being distracting. (Also, the field is on `CreateBookRequest`, not `Book`.)

PR #435, opened 2019-12-06
[AIP-134] Enforce a consistent method signature.

lukesneeringer:


docs/rules/0134/method-signature.md
```diff
@@ -0,0 +1,70 @@
...
+rpc UpdateBook(UpdateBookRequest) returns (Book) {
+  // A google.api.method_signature annotation should be present.
+}
+```
+
+```proto
+// Incorrect.
+rpc UpdateBook(UpdateBookRequest) returns (Book) {
+  // Should be "book,update_mask".
+  option (google.api.method_signature) = "book,update_mask";
```
mingzhi: Why is it incorrect?

lukesneeringer:
```suggestion
  option (google.api.method_signature) = "book";
```

PR #446, opened 2019-12-11
[fix] Fix resource annotation noisyness, hopefully.

lukesneeringer:
I can not actually reproduce the bug (showing up in internal source control) that caused me to do this, but I am hoping this fixes it. We shall find out.


lukesneeringer:
> Approve, but I think it is the same.

Same. Holding off on merging to see if it was a weird flake or something.

lukesneeringer:
I have not heard any more about this, so I am going to close this PR.

PR #455, opened 2019-12-20
[Refactor] Allow AIP grouping function injection.

mingzhi:


mingzhi:
> This is fine with me, although it is not entirely obvious how it solves the problem internally. I assume that we add new files to the `lint` package that are internal only?

Exactly.

PR #457, opened 2019-12-20
[AIP-136] Add custom method rule for the name variable.

lukesneeringer:


rules/aip0136/http_name_variable.go
```diff
@@ -0,0 +1,55 @@
...
+		for _, http := range utils.GetHTTPRules(m) {
+			vars := http.GetVariables()
+
+			// If there is a "name" variable, the noun should be present
+			// in the RPC name.
+			if name, ok := vars["name"]; ok {
+				// Determine the resource.
+				name = strings.TrimSuffix(name, "/*")
+				segs := strings.Split(name, "/")
+				resource := p.Singular(segs[len(segs)-1])
```
mingzhi: Is `resource` always single-word?

lukesneeringer: Yes.

PR #471, opened 2020-01-28
[AIP-216] support alias

### LINKED ISSUE #466: API Lint rule 126, Should not check for alias enum type
if the first enum type is also an alias, then an API lint error will occur. Say enum Type { option allow_alias = true; TYPE_UNSPECIFIED = 0; DEFAULT_TYPE = 0; TYPE_ONE = 1; } This type enum alias, in my understanding should be allowed.

Dos-Ph:
Resolve #466
- check if at least one enum with <name>_UNSPECIFIED
- check if <name>_UNSPECIFIED = 0
- added alias test case.

### DIFF
--- a/rules/aip0126/unspecified.go
+++ b/rules/aip0126/unspecified.go
@@ -24,25 +24,21 @@ import (
 	"github.com/stoewer/go-strcase"
 )
 
-var unspecified = &lint.EnumValueRule{
-	Name:   lint.NewRuleName(126, "unspecified"),
-	OnlyIf: isFirstEnumValue,
-	LintEnumValue: func(v *desc.EnumValueDescriptor) []lint.Problem {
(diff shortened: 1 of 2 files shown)


lukesneeringer:
I have mixed feelings about this. In particular, the use case in the test (`HARDBACK` and `FORMAT_UNSPECIFIED` being set to 0) should definitely be something the linter complains about.

I am assuming the real reason for this PR is something more restrained (e.g. an existing API used `UNKNOWN` or something and wants to add an enum value to be in compliance). Couldn't the API just put the new value as the top one in the enum, though?

lukesneeringer:
I just realized that #466 says that it happens if there is _any_ alias on `0` at all regardless of order, so this PR is definitely needed.

PR #478, opened 2020-02-03
[fix] Fix two-word noun false positives.

lukesneeringer:
The `http-name-variable` and `http-parent-variable` rules each
have a false positive if the noun in question is two words.
This PR fixes that.


rules/aip0136/http_name_variable_test.go
```diff
@@ -30,6 +30,7 @@ func TestHTTPVariables(t *testing.T) {
 		{"Valid", "WriteBook", "/v1/{name=publishers/*/books/*}:write", testutils.Problems{}},
 		{"Invalid", "WritePage", "/v1/{name=publishers/*/books/*}:writePage", testutils.Problems{{Message: "name variable"}}},
 		{"ValidBookVar", "WritePage", "/v1/{book=publishers/*/books/*}:writePage", testutils.Problems{}},
+		{"ValidTwoWordNoun", "WriteAudioBook", "/v1/{name=publishers/*/audioBooks/*}:write", testutils.Problems{}},
```
mingzhi: Is "/v1/{name=publishers/*/books/*}:write" Okay for "WriteAudioBook"?

lukesneeringer: No, that would and should error.

lukesneeringer: Actually, I take that back. It _should_ return a problem, but will pass. I would prefer fix that in a separate PR if that is okay.

lukesneeringer: Actually, I prefer not to fix it at all, now that I think about it more, because there is an important valid case (when a resource has the prefix of its parent) that we permit.

So...working as intended. :)

PR #480, opened 2020-02-03
[fix] AIP-162 tag/list/delete methods should not false positive.

lukesneeringer:


rules/aip0136/aip0136.go
```diff
@@ -38,6 +39,11 @@ func AddRules(r lint.RuleRegistry) error {
 }
 
 func isCustomMethod(m *desc.MethodDescriptor) bool {
+	for _, httpRule := range utils.GetHTTPRules(m) {
+		if strings.Contains(httpRule.GetPlainURI(), ":") {
+			return true
+		}
+	}
```
mingzhi: Consider to add a comment here. Do you have a test for it?

lukesneeringer: Yes, tests in this module have full coverage. I will add a comment though.

lukesneeringer: Done.

PR #492, opened 2020-02-13
Fix disabling rule typo for versioned-packages

sagiyin:


docs/rules/0215/versioned-packages.md
```diff
@@ -58,8 +58,5 @@ Remember to also include an [aip.dev/not-precedent][] comment explaining why.
 package foo.bar;
 ```
 
-If you need to violate this rule for an entire file, place the comment at the
```
jgeewax: OK - we want this gone because "above the package" *is* the top of the file!

PR #495, opened 2020-02-14
[AIP-192] Add rule for trademarked names.

lukesneeringer:
Implements this AIP PR: googleapis/aip#420
~Do not merge until the AIP PR is merged.~ EDIT: It now is.

### DIFF
--- a/docs/rules/0192/trademarked-names.md
+++ b/docs/rules/0192/trademarked-names.md
@@ -1,0 +1,86 @@
+---
+rule:
+  aip: 192
+  name: [core, '0192', trademarked-names]
+  summary: Trademarked names should be used correctly.
+permalink: /192/trademarked-names
+redirect_from:
+  - /0192/trademarked-names
+---
+
+# Trademarked names
+
+This rule enforces trademarked names in public comments are not abbreviated and
+follow the trademark owner's branding style, as mandated in [AIP-192][].
+
+## Details
+
+This rule looks at each descriptor in each proto file (exempting oneofs and the
+file itself) and complains if it catches a common trademark mistake. It only
+checks against a known problem list, and does not make guesses beyond that.
+
+It currently catches common mistaken variants for:
+
+- App Engine
+- BigQuery
+- Bigtable
+- Bitbucket
+- Cloud Storage
+- Compute Engine
+- Dataflow
+- Dataprep
+- Dialogflow
+- GitHub
+- GitLab
+- G Suite
+- Pub/Sub
+- Stack Overflow
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message Book {
+  string name = 1;
+
+  // A repository containing Markdown files for each chapter of
+  // the book on Github.
+  // (--           ^ Should be GitHub. --)
+  string github_repo = 2;
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+message Book {
+  string name = 1;
+
+  // A repository containing Markdown files for each chapter of
+  // the book on GitHub.
+  string github_repo = 2;
+}
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the descriptor.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+message Book {
+  string name = 1;
+
+  // (-- api-linter: core::0192::trademarked-names=disabled
+  //     aip.dev/not-precedent: We need to do this because reasons. --)
+  // A repository containing Markdown files for each chapter of
+  // the book on Github.
+  string github_repo = 2;
+}
+```
+
+[aip-192]: https://aip.dev/192
+[aip.dev/not-precedent]: https://aip.dev/not-precedent
--- a/docs/rules/0192/trademarked-names.md
+++ b/docs/rules/0192/trademarked-names.md
@@ -30,10 +30,12 @@ It currently catches common mistaken variants for:
 - Dataflow
 - Dataprep
 - Dialogflow
+- Directory Sync
 - GitHub
 - GitLab
 - G Suite
 - Pub/Sub
+- Service Mesh
 - Stack Overflow
 
 ## Examples
--- a/rules/aip0192/trademarked_names.go
+++ b/rules/aip0192/trademarked_names.go
@@ -26,6 +26,7 @@ import (
 var trademarks = map[string][]string{
 	"App Engine":     []string{"GAE", "AppEngine", "App engine"},
 	"BigQuery":       []string{"Bigquery", "Big Query", "BQ"},
+	"BigQuery ML":    []string{"BQML"},
 	"Bigtable":       []string{"BigTable", "Big Table", "Big table"},
 	"Bitbucket":      []string{"BitBucket", "Bit Bucket"},
 	"Cloud Storage":  []string{"GCS"},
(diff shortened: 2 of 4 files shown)


rules/aip0192/trademarked_names.go
```diff
@@ -0,0 +1,80 @@
...
+	"Directory Sync": []string{"GCDS", "CDS", "DirectorySync"},
+	"GitHub":         []string{"Github", "Git Hub"},
+	"GitLab":         []string{"Gitlab", "Git Lab"},
+	"G Suite":        []string{"GSuite", "G-Suite", "gSuite"},
+	"Pub/Sub":        []string{"PubSub", "Pubsub", "Cloud Pub/Sub"},
+	"Service Mesh":   []string{"ASM", "CSM", "GCSM"},
+	"Stack Overflow": []string{"StackOverflow"},
+}
+
+// We actually want regexes so we do not accidentally false-positive acronyms
```
mingzhi: Can we just try to search substring with spaces before and after the word, like `" BQ "`?

rules/aip0192/trademarked_names.go
```diff
@@ -0,0 +1,80 @@
...
+	"Dataflow":       []string{"Data Flow", "Data flow", "DataFlow"},
+	"Dataprep":       []string{"Data Prep", "Data prep", "DataPrep"},
+	"Dialogflow":     []string{"DialogFlow", "Dialog Flow", "Dialog flow"},
+	"Directory Sync": []string{"GCDS", "CDS", "DirectorySync"},
+	"GitHub":         []string{"Github", "Git Hub"},
+	"GitLab":         []string{"Gitlab", "Git Lab"},
+	"G Suite":        []string{"GSuite", "G-Suite", "gSuite"},
+	"Pub/Sub":        []string{"PubSub", "Pubsub", "Cloud Pub/Sub"},
+	"Service Mesh":   []string{"ASM", "CSM", "GCSM"},
+	"Stack Overflow": []string{"StackOverflow"},
```
piac58: Please check for lower case also: gcs, gce, etc. Many times that is what gets into the docs.

rules/aip0192/trademarked_names.go
lukesneeringer: I thought about it, but it gets a lot of cases wrong (e.g. the use of the substring at the end of the sentence). `\b` is really the correct thing for this purpose.

rules/aip0192/trademarked_names.go
lukesneeringer: Done, except for `bq` (which would false positive `bq://...`).

docs/rules/0192/trademarked-names.md
```diff
@@ -0,0 +1,88 @@
...
+
+```proto
+// Incorrect.
+message Book {
+  string name = 1;
+
+  // A repository containing Markdown files for each chapter of
+  // the book on Github.
+  // (--           ^ Should be GitHub. --)
+  string github_repo = 2;
```
jgeewax: Important to note that `github` is correct in code format (not `git_hub`). What's the correct camel-case for this? `gitHubUsername`? or `githubUsername`? (I think the former...)

rules/aip0192/trademarked_names.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var trademarks = map[string][]string{
```
jgeewax: Isn't this really a list of trademark aliases? or common mistakes?

docs/rules/0192/trademarked-names.md
lukesneeringer: This _only_ checks comments, not code format.

rules/aip0192/trademarked_names.go
lukesneeringer: Done.

rules/aip0192/trademarked_names.go
```diff
@@ -0,0 +1,80 @@
...
+
+func init() {
+	for k, tms := range trademarkAliases {
+		tmReg := []*regexp.Regexp{}
+		for _, tm := range tms {
+			tmReg = append(tmReg, regexp.MustCompile(`\b`+strings.ReplaceAll(tm, " ", `\s+`)+`\b`))
+		}
+		tmRegexes[k] = tmReg
+	}
+}
```
mingzhi: Can you create a function for it, and give a better name? I image that we (or others) will add more trademarks. Like:
```golang
var trademarkTyposRegexs = defaultTrademarkTypos()
func defaultTrademarkTypos() {
...
}
```
Then, if someone wants to customize the list, they can modify the map in their `init` function in an additional go file, e.g., "google_internal_trademark_types.go"

rules/aip0192/trademarked_names.go
mingzhi: Got it. Thanks!

docs/rules/0192/trademarked-names.md
jgeewax: OK. I'll create an issue to check that we don't do things like `git_hub_repo`, etc.

docs/rules/0192/trademarked-names.md
```diff
@@ -0,0 +1,88 @@
...
+---
+
+# Trademarked names
+
+This rule enforces trademarked names in public comments are not abbreviated and
+follow the trademark owner's branding style, as mandated in [AIP-192][].
+
+## Details
+
+This rule looks at each descriptor in each proto file (exempting oneofs and the
```
jgeewax: Why do we exempt oneofs ?

rules/aip0192/trademarked_names.go
jgeewax: +1 to this. At first glance it's unclear why we're doing a bunch of init() work, but if it were called `defaultTrademarkTypos` (or something similar), that'd make this much much clearer.

docs/rules/0192/trademarked-names.md
lukesneeringer: We have not needed a oneof descriptor rule yet, and oneof documentation is unused to the best of my knowledge.

rules/aip0192/trademarked_names.go
lukesneeringer: Done.

mingzhi: Luke, can you initialize the map to the variable, rather than in the `init` function? Global variable initialization precedes `init` function. And the `init` functions in different files are invoked according the file names in alphabet order. So it is better to use variable initialization for default values, and then allow `init` function to modified the variable. Like:
```golang
var tmRegexes = defaultTrademarkTypes()
```

In general, we would like to be careful about `init` functions, as they are invoked in order by file names.

lukesneeringer: Done.

rules/aip0192/trademarked_names.go
```diff
@@ -0,0 +1,82 @@
...
+	"Stack Overflow": []string{"StackOverflow"},
+}
+
+// We actually want regexes so we do not accidentally false-positive acronyms
+// that *contain* our matches. (For example, "BQD" should not match and tell us
+// to change to BigQuery.)
+func defaultTrademarkTypos() map[string][]*regexp.Regexp {
+	tmRegexes := map[string][]*regexp.Regexp{}
+	for k, tms := range trademarkAliases {
+		tmReg := []*regexp.Regexp{}
```
mingzhi: Nitpicking: var declarations are preferred for slices: to create a nil slice instead of a length 0 slice. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

lukesneeringer: I want to declare a length 0 slice because otherwise it gives me a nil error when I assign to it.

Issue #490, opened 2020-02-12
[AIP-158] Be stricter about pagination.


lukesneeringer:
We need a lint rule (and amendment to AIP-158 itself) that will enforce the following:

For all messages that:
* end in `Response`
* Contain a `next_page_token` field.

Throw a linter warning if:

* The _first_ field in the message (by both position and field number) is not repeated
* The _first_ field in the message is a primitive
* The _first_ field in the message is not named according to `plural(message_name.to_snake().split('_')[1:-1])`

These should probably all be separate rules.

Dos-Ph:
I will send PR for this.

lukesneeringer:
This is done now.

PR #496, opened 2020-02-17
[AIP-158] Response's first field must be plural

Dos-Ph:
Issue #490 

For all messages that:
- end in Response
- Contain a next_page_token field.

Throw a linter warning if:
The first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1])


rules/aip0158/response_first_field_must_be_plural.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+
+	"github.com/gertd/go-pluralize"
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var responseFirstFieldMustBePlural = &lint.MessageRule{
+	Name:   lint.NewRuleName(158, "response-first-field-must-be-plural"),
```
lukesneeringer: Let's use `response-plural-first-field`

rules/aip0158/response_first_field_must_be_plural.go
```diff
@@ -0,0 +1,52 @@
...
+	OnlyIf: isPaginatedResponseMessage,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		// Rule check: for all messages that end in Response and contain a next_page_token field.
+		// Throw a linter warning if, the first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1]).
+		nextPageToken := m.FindFieldByName("next_page_token")
+		if nextPageToken == nil {
+			return nil
+		}
+
+		firstField := m.FindFieldByNumber(1)
```
lukesneeringer: We probably care just as much (if not more) about it being _positionally_ first in the file: `m.GetFields()[0]`

rules/aip0158/response_first_field_must_be_plural.go
```diff
@@ -0,0 +1,52 @@
...
+		pluralize := pluralize.NewClient()
+		if !pluralize.IsPlural(firstField.GetName()) {
```
lukesneeringer: This may give unhelpful feedback in some situations.
For example, let's say I did...

```proto
message ListBooksResponse {
  string next_page_token = 1;
  repeated Book books = 2;
}
```

The thing the linter should tell me to do is to reverse the order of the fields. What it will actually do is still me to rename the first field to `string next_page_tokens`.

**Note:** This is probably a situation where there are several rules and at most one of them should complain to avoid giving users misleading output.

docs/rules/0158/response-plural-first-field.md
```diff
@@ -0,0 +1,57 @@
...
+  aip: 158
+  name: [core, '0158', response-plural-first-field]
+  summary: First field of Paginated RPCs' response should be plural.
+permalink: /158/response-plural-first-field
+redirect_from:
+  - /0158/response-plural-first-field
+---
+
+# Paginated methods: Page token field
+This rule enforces that all `List` and `Search` methods have a plural name repeatable field as a first field in the response message, as mandated in [AIP-158][].
```
lukesneeringer: Before merging, make sure to run prettier on this. (If you are using VS Code, there is a great plugin that will just do it every time you save.)

rules/aip0158/response_plural_first_field.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+var responsePluralFirstField = &lint.MessageRule{
+	Name:   lint.NewRuleName(158, "response-plural-first-field"),
+	OnlyIf: isPaginatedResponseMessage,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		// Rule check: for all messages that end in Response and contain a next_page_token field.
+		// Throw a linter warning if, the first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1]).
+		nextPageToken := m.FindFieldByName("next_page_token")
+		if nextPageToken == nil {
```
lukesneeringer: Does line 28 not cover this?

rules/aip0158/response_plural_first_field.go
```diff
@@ -0,0 +1,52 @@
...
+			return nil
+		}
+
+		firstField := m.GetFields()[0]
+		pluralize := pluralize.NewClient()
+		if !pluralize.IsPlural(firstField.GetName()) {
+			want := pluralize.Plural(firstField.GetName())
```
lukesneeringer: I still think there are some cases where this linter rule should be quiet (and defer to others to be loud).

I _think_ we should probably only grouse about the name if the field is a repeated non-primitive, and a _different_ rule will complain about those errors.

rules/aip0158/response_first_field_must_be_plural.go
lukesneeringer: I still think there are some cases where this linter rule should be quiet (and defer to others to be loud).

I think we should probably only grouse about the name if the field is a repeated non-primitive, and a different rule will complain about those errors.

rules/aip0158/response_plural_first_field.go
Dos-Ph: No, it doesn't cover. line 28 just check that pagination response matches with this > regexp.MustCompile("^(List|Search)[A-Za-z0-9]*Response$").

docs/rules/0158/response-plural-first-field.md
Dos-Ph: Done

docs/rules/0158/response-plural-first-field.md
```diff
@@ -0,0 +1,64 @@
...
+    string next_page_token = 2;
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+message ListStudentProfilesResponse {
+  repeated Profile studentProfiles = 1;
```
jgeewax:
```suggestion
  repeated Profile student_profiles = 1;
```

docs/rules/0158/response-plural-first-field.md
```diff
@@ -0,0 +1,64 @@
...
+If you need to violate this rule, use a leading comment above the message or
+above the field. Remember to also include an [aip.dev/not-precedent][] comment
+explaining why.
+
+```proto
+// (-- api-linter: core::0158::response-plural-first-field=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+message ListStudentProfilesResponse {
+    repeated Profile studentProfile = 1; // Field name should be `studentProfiles`.
+    string next_page_token = 2;
```
jgeewax:
```suggestion
  repeated Profile student_profile = 1; // Field name should be `student_profiles`.
  string next_page_token = 2;

File: docs/rules/0158/response-plural-first-field.md [hunk 9]
```diff
@@ -0,0 +1,64 @@
...
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect
+message ListStudentProfilesResponse {
+    repeated Profile studentProfile = 1; // Field name should be `studentProfiles`.
+    string next_page_token = 2;
```
jgeewax:
```suggestion
  repeated Profile student_profile = 1; // Field name should be `student_profiles`.
  string next_page_token = 2;

File: rules/aip0158/response_plural_first_field.go [hunk 10]
```diff
@@ -0,0 +1,52 @@
...
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		// Rule check: for all messages that end in Response and contain a next_page_token field.
+		// Throw a linter warning if, the first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1]).
+		nextPageToken := m.FindFieldByName("next_page_token")
+		if nextPageToken == nil {
+			return nil
+		}
+
+		firstField := m.GetFields()[0]
+		pluralize := pluralize.NewClient()
```
jgeewax: This leads us to a new question about pluralization and where English meets Engineering. If the resource is weird (e.g., person or cactus), which is correct?

Option A:
  - `GetCactus`, `ListCacti` (with `repeated Cactus cacti = 1`)
  - `GetPerson`, `ListPeople` (with `repeated Person people = 1`)

or Option B:
  - `GetCactus`, `ListCactuses` (with `repeated Cactus cactuses = 1`)
  - `GetPerson`, `ListPersons` (with `repeated Person persons = 1`)

Do we want to express an opinion on this? The code here says "Cactus (singular) -> Cacti (plural)" even if folks wanted the other way ("Cactuses").

rules/aip0158/response_plural_first_field.go
jgeewax: Does it matter whether we include pagination or not? Just because we don't have a paginated response doesn't mean we shouldn't check whether the field here is a repeated list of resources with a pluralized name, right?

rules/aip0158/response_plural_first_field_test.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func TestResponsePluralFirstField(t *testing.T) {
+	tests := []struct {
+		testName    string
+		MessageName string
+		problems    testutils.Problems
+	}{
+		{"Valid", "student_profiles", testutils.Problems{}},
+		{"InvalidWrongSuffix", "student_profile", testutils.Problems{{Suggestion: "student_profiles"}}},
```
jgeewax:
```suggestion
		{"Valid", "student_profiles", testutils.Problems{}},
		{"InvalidWrongSuffix", "student_profile", testutils.Problems{{Suggestion: "student_profiles"}}},
    {"ValidLatin", "cacti", testutils.Problems{}},
    {"InvalidLatin", "cactuses", testutils.Problems{{Suggestion: "cacti"}}},
    {"ValidNonstandard", "people", testutils.Problems{}},
    {"InvalidNonstandard", "persons", testutils.Problems{{Suggestion: "people"}}},

File: rules/aip0158/response_plural_first_field.go (hunk 10 above)
Comment (Dos-Ph) [reply rules/aip0158/response_plural_first_field.go]: I prefer option A. However, the current library doesn't support it. I will find a new lib.

Comment (jgeewax) [reply rules/aip0158/response_plural_first_field.go]: I believe it does based on the example here: https://github.com/gertd/go-pluralize#word-make-plural

```
pluralize -word Cactus -cmd Plural

Plural(Cactus)     => Cacti
```

If we do go this route, it's important to have tests for these Latin- and Greek-origin cases (Cactus -> Cacti, Octopus -> Octopuses? Octopi? Octopedes?). As well as the weird English ones (e.g., Person -> People, Goose -> Geese, etc).

Comment (Dos-Ph) [reply rules/aip0158/response_plural_first_field.go]: Done

File: rules/aip0158/response_plural_first_field_test.go (hunk 11 above)
Comment (Dos-Ph) [reply rules/aip0158/response_plural_first_field_test.go]: Done

File: rules/aip0158/response_plural_first_field.go (hunk 5 above)
Comment (lukesneeringer) [reply rules/aip0158/response_plural_first_field.go]: Well, my intent was to attach this rule to the _pagination_ AIP.
I think that is still what I want to do -- it is pagination that triggers the need for this check, and if pagination is implemented, it does not matter whether the request is a list, a search, or custom.

File: docs/rules/0158/response-plural-first-field.md [hunk 12]
```diff
@@ -0,0 +1,64 @@
...
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect
+message ListStudentProfilesResponse {
+    repeated Profile student_profile = 1; // Field name should be `student_profiles`.
+    string next_page_token = 2;
+}
```
lukesneeringer: Let's stick with the canonical example, and two-space indenting.

```suggestion
message ListBooksResponse {
  // Field name should be `books`.
  repeated Book book = 1;
  string next_page_token = 2;
}
```

docs/rules/0158/response-plural-first-field.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+message ListStudentProfilesResponse {
+  repeated Profile student_profiles = 1;
+  string next_page_token = 2;
+}
```
lukesneeringer:
```suggestion
message ListBooksResponse {
  repeated Book book = 1;
  string next_page_token = 2;
}
```

docs/rules/0158/response-plural-first-field.md
```diff
@@ -0,0 +1,64 @@
...
+above the field. Remember to also include an [aip.dev/not-precedent][] comment
+explaining why.
+
+```proto
+// (-- api-linter: core::0158::response-plural-first-field=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+message ListStudentProfilesResponse {
+    repeated Profile student_profile = 1; // Field name should be `student_profiles`.
+    string next_page_token = 2;
+}
```
lukesneeringer:
```suggestion
message ListBooksResponse {
  repeated Book book = 1;
  string next_page_token = 2;
}
```

rules/aip0158/response_plural_first_field.go
```diff
@@ -0,0 +1,52 @@
...
+
+var responsePluralFirstField = &lint.MessageRule{
+	Name:   lint.NewRuleName(158, "response-plural-first-field"),
+	OnlyIf: isPaginatedResponseMessage,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		// Rule check: for all messages that end in Response and contain a next_page_token field.
+		// Throw a linter warning if, the first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1]).
+		if m.FindFieldByName("next_page_token") == nil {
+			return nil
+		}
```
lukesneeringer: This should be part of the `OnlyIf` function.

rules/aip0158/response_plural_first_field.go
Dos-Ph: Got it. So, looks like now we have 2 ways to defined pagination message.
1. Message that match with ("^(List|Search)[A-Za-z0-9]*Response$").
2. Message that match with  ("[A-Za-z0-9]*Response$") and has next_page_token field.

jgeewax: @lukesneeringer Why wouldn't we attach it to the List method AIP ? (https://aip.dev/132#response-message) The List AIP includes a pointer to requirements stemming from the Pagination AIP, and this particular field that we care about here is relevant even in cases where an API decides not to implement pagination (for whatever reason).

Dos-Ph: We don't have a rule for the Search method. Should we have a new rule for this? or should I update these functions?
1. [isPaginatedResponseMessage](https://github.com/googleapis/api-linter/blob/master/rules/aip0158/aip0158.go#L43-L46) function: the message is pagination message only when the message contains the `next_page_token_field` field, or it matches with ("^(List|Search)[A-Za-z0-9]*Response$") pattern. 
2. [isPaginatedRequestMessage](https://github.com/googleapis/api-linter/blob/master/rules/aip0158/aip0158.go#L38-L41) function: the message is pagination message only when the Request message contains `page_size` field or `page_token` field, or it matches with ("^(List|Search)[A-Za-z0-9]*Request$") pattern.

rules/aip0158/response_plural_first_field.go
```diff
@@ -0,0 +1,52 @@
...
+		firstField := m.GetFields()[0]
+		pluralize := pluralize.NewClient()
+
+		// Need to convert name to singular first to support none standard case such as persons, cactuses.
```
mingzhi: I suggest to create a util function for the conversion and hide the implementation details, like:
```golang
func toPlural(s string) string {
  ...
}

File: rules/aip0158/response_plural_first_field.go [hunk 17]
```diff
@@ -0,0 +1,52 @@
...
+var responsePluralFirstField = &lint.MessageRule{
+	Name:   lint.NewRuleName(158, "response-plural-first-field"),
+	OnlyIf: isPaginatedResponseMessage,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		// Rule check: for all messages that end in Response and contain a next_page_token field.
+		// Throw a linter warning if, the first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1]).
+		if m.FindFieldByName("next_page_token") == nil {
+			return nil
+		}
+		firstField := m.GetFields()[0]
```
mingzhi: Could you check if the message is empty, which will return nil for `firstField` and the linter will panic in this case if unchecked? Thanks!

rules/aip0158/response_plural_first_field.go
lukesneeringer: Right, I agree with @Dos-Ph here. The reason we attach it to the pagination AIP is because otherwise we have to duplicate the rule for the various different types of methods that might include pagination, which is not what we want.

In fact, I would argue that the non-list methods are actually the most important because they have less defined structure. (Anecdotally, the RPC that made us realize we needed this rule was a custom method.)

rules/aip0158/response_plural_first_field.go
lukesneeringer: I thought that too at first, but we already know that there is a `next_page_token` field because we checked for that earlier.

mingzhi: I missed that. Thanks!

rules/aip0158/response_plural_first_field.go
jgeewax: > The reason we attach it to the pagination AIP is because otherwise we have to duplicate the rule for the various different types of methods that might include pagination, which is not what we want.

I understand that -- and I get that it's easier to say "Paginated stuff should look like this", however we're using pagination as the "easy way to filter things that look like lists and searches". I still argue that we should have a rule saying that even if you didn't do pagination, List and Search responses should have the first item being a repeated field. In other words, I want a rule that will throw an error for:

```
message ListBooksResponse {
  Book books = 1;
}
```

Right now, this rule will *not* do this (correct me if I'm wrong). I suggest we have a simple function that checks:

1. Is this a paginated response (check for next_page_token)? If so, then this message is in play.
2. Is this a standard method that should follow the rule (check for Search*Response and List*Response)? If so, then this message is in play.

If neither of those matches, then continue on. Perhaps this is "IsCollectionResponse" or something similar.

rules/aip0158/response_plural_first_field.go
Dos-Ph: I created a separate file for this since we can use it with other rules ex. https://aip.dev/122 > Collection identifiers- Collection identifiers must be plural. 

However, I found out the lib consider `infos` as a plural form of `info`.
https://play.golang.org/p/l4FrJuok5Rh

rules/aip0158/response_plural_first_field.go
Dos-Ph: I think this should cover your case now.
```
var paginatedReq = regexp.MustCompile("^(List|Search)[A-Za-z0-9]*Request$")
var paginatedRes = regexp.MustCompile("^(List|Search)[A-Za-z0-9]*Response$")

// Return true if this is an AIP-158 List request message, false otherwise.
func isPaginatedRequestMessage(m *desc.MessageDescriptor) bool {
	return paginatedReq.MatchString(m.GetName()) || m.FindFieldByName("page_size") != nil || m.FindFieldByName("page_token") != nil
}

// Return true if this is an AIP-158 List response message, false otherwise.
func isPaginatedResponseMessage(m *desc.MessageDescriptor) bool {
	return paginatedRes.MatchString(m.GetName()) || m.FindFieldByName("next_page_token") != nil
}
```

rules/aip0158/response_plural_first_field.go
mingzhi: Thanks! Yes, I remember there are couples of places that we can use your function. And it will be easier for us to switch to another library if desired, or make some manual corrections.

Maybe we should move your function into `rules/internal/utils`. What do you think?

rules/aip0158/response_plural_first_field.go
lukesneeringer: I think we might be mixing two concerns here. There are two things we need:

1. On methods with pagination, we need fields to come in the proper order. That is the domain of this rule.

2. On List and Search methods, we specifically need the first field to be _the resource_, and be plural and repeated. In other words, we need the response message equivalent of [this rule for create](https://linter.aip.dev/133/request-resource-field).

Let's keep these concerns separate. For _this_ rule, let's focus just on the pagination aspect, and I will write a separate rule around making sure the appropriate fields (both resource and pagination) are included in list responses. We have those for most of the rest of the standard methods but this one slipped through the cracks.

Dos-Ph: > On methods with pagination, we need fields to come in the proper order. 
= The first field should be repeated and plural
> On List and Search methods, we specifically need the first field to be the resource, and be plural and repeated
= The first field should be repeated and plural
It looks the same to me.

[Question1] For what you said, you mean that for List and Search response we will have a lint rule for `next_page_token`, right?

[Question2] And for all the response that has `next_page_token` will be considered as Pagination Method. And, we will have lint rules for the following condition, right?
```
The first field in the message (by both position and field number) is not repeated
The first field in the message is a primitive
The first field in the message is not named according to plural(message_name.to_snake().split('_')[1:-1])
```
[Question3] What's about Search Method?

rules/aip0158/response_plural_first_field.go
Dos-Ph: Done

rules/aip0158/response_plural_first_field.go
lukesneeringer: > = The first field should be repeated and plural
> Both look the same to me.

There are subtle differences.

- List and Search methods care that the field being repeated over _is the resource being listed_. But a pagination rule does not care what is being paginated.
- The pagination rule cares that the repeated field _lives in a particular spot_ (field 1, at the top of the message), because we do auto-sniffing in some of our tooling to figure out what field to paginate. This is mostly ancillary to list/search; it is specifically a pagination issue.

There is definitely overlap (a field should be repeated and plural) but they are not the same.

> [Question1] For what you said, you mean that for List and Search response we will have a lint rule for `next_page_token`, right? For the List method, we will add the check under AIP-132. [Question2] What's about the Search Method?

Yeah, we need two rules actually, one for each required field. We have them for the request, but not the response -- we just missed them somehow.

We do not have a Search AIP yet, and we need to figure out how strict we want to be on Search. Given the firm rule on having AIP guidance before a lint rule, most likely we would wait on that one.

> [Question3] And for all the response that has `next_page_token` will be considered as Pagination Method. And, we will have lint rules for the following condition defined under AIP-158, right?

Yeah.

Dos-Ph: So, what's about I submit this PR first. And, then create another PR to add next_page_token to List Method lint and update `isPaginatedResponseMessage ` Pagination lint rules?

Dos-Ph:
Could someone merge it? I don't have write access.

PR #501, opened 2020-02-19
Only detect closing tags in no-html lint

### LINKED ISSUE #500: AIP-192: HTML checker picking up example syntax with angle brackets
(From msuozzo@)

Imagine the following comment:

```
// The following format is required:
// spiffe://<domain>/ns/<namespace>/sa/<serviceaccount>
// e.g. spiffe://example.com/ns/test-ns/sa/default
```

These `<example>` annotations are not HTML. I think the linter should really do tag matching if it's going to claim to detect HTML. If something lighter-touch is desirable, it might also filter to a set of know tags like `<code>`, `<div>`, etc.

---

We should either clarify a different recommended example format (e.g., `{foo}` or something?) or be more specific about what we think looks like HTML.

msuozzo:
Fixes #500

There are plenty of places where angle brackets are used to delimit e.g. a placeholder value.
For example: `projects/<proj-id>/...` is a common example format.

This change detects only closing tags which should be a less expansive means of detecting actual
HTML usage in comments.

### DIFF
--- a/rules/aip0192/no_html.go
+++ b/rules/aip0192/no_html.go
@@ -42,10 +42,10 @@ var noHTML = &lint.DescriptorRule{
 // Yes, yes, I know: https://stackoverflow.com/questions/1732348/
 // Even Jon Skeet cannot parse HTML using regular expressions.
 //
-// That said, we really only want to pick up "basic HTML smell", and are
+// That said, we really only want to pick up "basic HTML smell" and are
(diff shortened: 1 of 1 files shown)


jgeewax:
Awesome ! Thanks -- this is a great idea!

I added some tests and such in #502 if you could take a look? If that looks OK, we can close this one out.

mingzhi:
Thank you so much! Your PR has been incorporated in PR #502

PR #506, opened 2020-02-25
[feat] Ignore deprecated descriptors.

lukesneeringer:
@wora had a really good point: running lint rules on deprecated
descriptors almost never makes sense, and they contribute to noise
for API producers.

This PR issues a blanket disable on any descriptor with a
`deprecated=true` annotation.

### DIFF
--- a/lint/rule.go
+++ b/lint/rule.go
@@ -372,6 +372,28 @@ func extractDisabledRuleName(commentLine string) string {
 // ruleIsEnabled returns true if the rule is enabled (not disabled by the comments
 // for the given descriptor or its file), false otherwise.
 func ruleIsEnabled(rule ProtoRule, d desc.Descriptor, aliasMap map[string]string) bool {
+	// Sanity check: All rules are disabled on a deprecated descriptor.
+	deprecated := false
+	switch v := d.(type) {
+	case *desc.EnumDescriptor:
+		deprecated = v.GetEnumOptions().GetDeprecated()
+	case *desc.EnumValueDescriptor:
+		deprecated = v.GetEnumValueOptions().GetDeprecated()
+	case *desc.FieldDescriptor:
+		deprecated = v.GetFieldOptions().GetDeprecated()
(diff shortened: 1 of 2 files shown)


lint/rule.go
```diff
@@ -372,6 +372,28 @@ func extractDisabledRuleName(commentLine string) string {
...
+		deprecated = v.GetMessageOptions().GetDeprecated()
+	case *desc.MethodDescriptor:
+		deprecated = v.GetMethodOptions().GetDeprecated()
+	case *desc.ServiceDescriptor:
+		deprecated = v.GetServiceOptions().GetDeprecated()
+	}
+	if deprecated {
+		return false
+	}
+
```
mingzhi: Can you create a function for it? And we can return in the `case`, like:
```golang
func isDescriptorDeprecated(d desc.Descriptor) bool {
  switch v := d.(type) {
  case *desc.EnumDescriptor:
    return v.GetEnumOptions().GetDeprecated()
  case *desc.EnumValueDescriptor:
    return v.GetEnumValueOptions().GetDeprecated()
  }
  return false
}
```

lint/rule_test.go
```diff
@@ -469,6 +470,45 @@ func TestRuleIsEnabledParent(t *testing.T) {
 	}
 }
 
+func TestRuleIsEnabledDeprecated(t *testing.T) {
```
mingzhi: And we can just test the newly created function.

lint/rule.go
lukesneeringer: I am going to add it to [protoreflect instead](https://github.com/jhump/protoreflect/issues/300), and I will back this out once that lands.

mingzhi: Good idea! Thanks!

lukesneeringer: Can we merge this in the meantime? :)

mingzhi: Approved

PR #512, opened 2020-03-04
Support source code location overrides for ruleEnabled.

alexoneill:
This fixes an issue where the linter does not respect a disable comment
on "package" statements, which themselves do not have an associated
Descriptor (natural file location). Instead these lint rules can specify
a Location, and this location will now be respected when checking for
rule disable comments.


mingzhi:
Added Luke for another look.

PR #513, opened 2020-03-05
Fix a logical typo: below -> above

adsouza:
https://aip.dev/191#file-layout says enum defs should be at the end so we should not complain if they are after messages.

### DIFF
--- a/docs/rules/0191/file-layout.md
+++ b/docs/rules/0191/file-layout.md
@@ -19,7 +19,7 @@ This rule checks for common file layout mistakes, but does not currently check
 the exhaustive file layout in AIP-191. This rule currently complains if:
 
 - Services appear below messages.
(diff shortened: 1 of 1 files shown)


docs/rules/0191/file-layout.md
```diff
@@ -19,7 +19,7 @@ This rule checks for common file layout mistakes, but does not currently check
 the exhaustive file layout in AIP-191. This rule currently complains if:
 
 - Services appear below messages.
-- Top-level enums appear below messages.
+- Top-level enums appear above messages.
```
jgeewax: OK, so the rules (aip.dev/191) say that enums should be LAST. So yes, this rule should complain if enums appear above messages. Took me a second to parse all of this, but this looks perfectly correct.

PR #520, opened 2020-03-12
[AIP-136] Be more permissive on collection suffixes.

lukesneeringer:
The lint rule contradicts [AIP-136](https://aip.dev/136) on the SortBooks example.
This fixes that.

### DIFF
--- a/rules/aip0136/http_uri_suffix.go
+++ b/rules/aip0136/http_uri_suffix.go
@@ -33,6 +33,8 @@ var uriSuffix = &lint.MethodRule{
 	},
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
 		for _, httpRule := range utils.GetHTTPRules(m) {
+			var want string
+
 			// URIs should end in a `:` character followed by the name of the method.
 			// However, if the noun is the URI's resource, then we only use `:verb`,
 			// not `:verbNoun`.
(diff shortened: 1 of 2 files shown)


rules/aip0136/http_uri_suffix.go
```diff
@@ -73,6 +76,24 @@ var uriSuffix = &lint.MethodRule{
...
+			// component of the RPC name, then assume this matches `nouns:verb`.
+			//
+			// Note that we do not do any pluralization here -- both the RPC and
+			// the URI must be plural already. If the RPC is singular, step 4 should
+			// apply.
+			plainURI := httpRule.GetPlainURI()
+			segs := strings.Split(strings.Split(plainURI, ":")[0], "/")
+			segment := segs[len(segs)-1]
+			if len(want) == 0 && strings.HasSuffix(strings.ToLower(n), segment) {
+				want = segment + ":" + strings.Split(strcase.SnakeCase(n), "_")[0]
```
mingzhi: Nitpicking: how about creating a helper function for splitting words? Like:
```go
func splitWords(s string) []string {
  return strings.Split(strcase.SnakeCase(n),  "_")
}
```
It has been used in many places. By creating a function, we can have more proper unit tests for it and make it easier for future reflector.

rules/aip0136/http_uri_suffix.go
```diff
@@ -73,6 +76,24 @@ var uriSuffix = &lint.MethodRule{
...
+			}
+
+			// Nothing else applied; expect `:verbNoun`.
+			if len(want) == 0 {
+				want = ":" + strcase.LowerCamelCase(strcase.SnakeCase(m.GetName()))
```
mingzhi: Why do we need to convert the name to snake case and then to lower camel case? And I think a helper function, like `toLowerCamelCase`, would help understanding and testing.

lukesneeringer:
I will address the nits separately.

PR #531, opened 2020-04-06
[refactor] Make rule enablement extensible.

lukesneeringer:
This PR adds an extension hook for disabling rules.
This will allow us to disable rules with visibility labels internally.


lukesneeringer:
@mingzhi Ping. :-)

lint/rule_enabled.go
```diff
@@ -0,0 +1,112 @@
...
+//
+// This pattern is a hook for internal extension, by creating an additional
+// file in this package that can add an additional check:
+//
+//    func disableForInternalReason(d desc.Descriptor) bool { ... }
+//
+//    func init() {
+//      descriptorDisableChecks = append(descriptorDisableChecks, disableForInternalReason)
+//    }
+var descriptorDisableChecks = []func(d desc.Descriptor) bool{
```
mingzhi, line 55: Do you want to use a map with a key(name) to describe the disabling check function? So if we find a disabling check is not applicable internally, we can remove the check based on the key(name).

lukesneeringer: I'd be up for making this change the first time it comes up, but for now I would prefer match [rule_urls.go](https://github.com/googleapis/api-linter/blob/master/lint/rule_urls.go) since that already exists and uses a slice.

Issue #522, opened 2020-03-13
Exit Status Code = 0 when linting problems are found


ialcor:
Right now, wether problems are found during linting process or are not, the program finishes with status code 0. This could be understood as the right behavior. 

However,  making the api-linter fail when problems are found could be helpful, it could help integrating api-linter into CI stages for example.  What do you think?

This capability to fail after problems are found could be activated conditionally, via input parameters to the linter.  This is a standard practice for some linter. 

example: 

>  api-linter someproto.proto -I some path --set_exit_status (print problems and then exit with code 1)
>  api-linter someproto.proto -I some path (print problems and exit correctly)

What do you think? I have some implementation and I could open a PR if you agree.

mingzhi:
+1. Can you open a PR? Thanks!

lukesneeringer:
My vote is for an exit status of 1 if there are problems; I agree with a switch to make it 0 if needed.

lukesneeringer:
Fixed in #534.

PR #534, opened 2020-04-23
[feat] return non-zero exit status when lint problems found

RoryQ:
This change introduces a new flag `--set-exit-status` which will return
a non-zero exit status when lint problems are found. This is a non
breaking change that allows api-linter to be used in CI environments.

This addresses issue #522

### DIFF
--- a/cmd/api-linter/cli.go
+++ b/cmd/api-linter/cli.go
@@ -33,14 +33,15 @@ import (
 )
 
 type cli struct {
-	ConfigPath       string
-	FormatType       string
-	OutputPath       string
-	ProtoImportPaths []string
-	ProtoFiles       []string
-	ProtoDescPath    []string
-	EnabledRules     []string
-	DisabledRules    []string
+	ConfigPath              string
+	FormatType              string
+	OutputPath              string
+	ExitStatusOnLintFailure bool
+	ProtoImportPaths        []string
+	ProtoFiles              []string
+	ProtoDescPath           []string
+	EnabledRules            []string
+	DisabledRules           []string
 }
 
 func newCli(args []string) *cli {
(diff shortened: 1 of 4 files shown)


cmd/api-linter/cli.go
```diff
@@ -132,29 +136,33 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
...
 	// Create a linter to lint the file descriptors.
 	l := lint.New(rules, configs)
 	results, err := l.LintProtos(fd...)
 	if err != nil {
-		return err
+		return lintFailure, err
+	}
+
+	if c.ExitStatusOnLintFailure && len(results) > 0 {
+		lintFailure = true
```
mingzhi, line 165: I think we can return an error, like
```go
if c.ExitStatusOnLintFailure && len(results) > 0 {
  return fmt.Errorf("Found %d linting results", len(results))
}
```
And in the `main` function, `Fatalf` is equivalent to `Printf()` followed by a call to `os.Exit(1)`. And you can remove `lintFailure`.

RoryQ: Thanks for the suggestion, it looks a lot cleaner now.

cmd/api-linter/cli.go
```diff
@@ -181,6 +187,13 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
 	if _, err = w.Write(b); err != nil {
 		return err
 	}
+
+	// Return error on lint failure which subsequently
+	// exits with a non-zero status code
+	if c.ExitStatusOnLintFailure && len(results) > 0 {
```
AndreasBieber, line 193: It should be checked if there are any problems in `results`:

```
results[*].Problems
```

Currently the linter returns a non-zero status code (false-positive).

RoryQ: Thanks, my test cases weren't doing what I thought they were. I'll submit a fix for this

PR #536, opened 2020-05-01
fix detection of lint problems for exit status

RoryQ:
This PR fixes a false positive reading of the lint results by checking `result[*].Problems` for lint results per file

see https://github.com/googleapis/api-linter/pull/534#discussion_r416489984


lukesneeringer:
Thank you @RoryQ!

PR #538, opened 2020-05-19
[AIP-215] Exempt known prefixes from versioning expectation.

lukesneeringer:


rules/aip0215/versioned_packages.go
```diff
@@ -27,16 +27,24 @@ import (
...
-	LintFile: func(f *desc.FileDescriptor) []lint.Problem {
 		p := f.GetPackage()
-		for _, exception := range []string{"master", "type"} {
-			if strings.HasSuffix(p, fmt.Sprintf(".%s", exception)) {
-				return nil
+		if p == "" {
+			return false
+		}
+		for _, exemptSuffix := range []string{"master", "type"} {
+			if strings.HasSuffix(p, fmt.Sprintf(".%s", exemptSuffix)) {
```
noahdietz, line 35: Nit: Why use `fmt.Sprintf` here to prefix the suffix with a `.`? Can we just make the string literals in the slice start with a `.`? The slice isn't being used anywhere else that would require the `.` prefix not be present.

lukesneeringer: Good point. Done.

PR #539, opened 2020-05-19
[chore] Fix Renovate updates.

lukesneeringer:
Ignore `go-pluralize`, which has been broken for a month, and `genproto`, which is just irritating.

Also convert to JSON5, which Renovate supports, so we can have comments.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,15 @@
+{
+  extends: ["config:base"],
+  commitMessagePrefix: "[chore] ",
+  groupName: "all",
+  ignoreDeps: [
+    // The current version is broken:
+    // https://github.com/gertd/go-pluralize/issues/8
+    "github.com/gertd/go-pluralize",
+
+    // Excessively annoying because any update to any API published by
+    // Google triggers an update to this.
+    "google.golang.org/genproto",
+  ],
+  postUpdateOptions: ["gomodTidy"],
+}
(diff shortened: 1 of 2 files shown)


.github/renovate.json5
```diff
@@ -0,0 +1,15 @@
...
+  commitMessagePrefix: "[chore] ",
+  groupName: "all",
+  ignoreDeps: [
+    // The current version is broken:
+    // https://github.com/gertd/go-pluralize/issues/8
+    "github.com/gertd/go-pluralize",
+
+    // Excessively annoying because any update to any API published by
+    // Google triggers an update to this.
+    "google.golang.org/genproto",
```
noahdietz, line 12: If it helps, we can schedule this one dependency? I do this for all of our Go tools, and the Cloud golang samples repo does it too. For example: https://github.com/googleapis/gapic-generator-go/blob/master/renovate.json#L19-L24

However, the config uses UTC time I think, so I need to adjust it to later, because I get PRs very early Monday morning and I can't get to them before merge-on-green starts complaining.

lukesneeringer: How would this work? Something like "update it once a month" or similar?
That does seem better than just not updating it at all. The problem I have currently is that it is almost-daily. Something like 10-15% of the total PRs are genproto updates from Renovate. :-p

lukesneeringer: Oh -- yeah, this is awesome. I will do that (but in a follow-up PR).

PR #545, opened 2020-05-29
[AIP-140] Prohibit url in field names.

lukesneeringer:
Do not merge until googleapis/aip#519 is merged.

### DIFF
--- a/rules/aip0140/uri_test.go
+++ b/rules/aip0140/uri_test.go
@@ -1,0 +1,46 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


rules/aip0140/uri_test.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func TestURI(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		Field    string
+		problems testutils.Problems
+	}{
+		{"Valid", "uri", nil},
+		{"ValidPrefix", "foo_uri", nil},
```
noahdietz, line 30: Nit: should we add a `ValidIntermixed` test that uses a field name like `curl_uri` with "url" inside a word to ensure it _doesn't_ emit an error? I understand that the implementation splits on `_` and does a `Contains` check on the slice contents, but implementations change, and the behavior shouldn't necessarily.

```suggestion
		{"ValidPrefix", "foo_uri", nil},
		{"ValidIntermixed", "curl_uri", nil},
```

lukesneeringer: Good idea.

lukesneeringer: Done. Still not merging until @jskeet approves the AIP change (probably Monday).

PR #550, opened 2020-06-02
[docs] Fix output-only docs.

### LINKED ISSUE #547: Extraneous line in documentation code example
https://github.com/googleapis/api-linter/blob/master/docs/rules/0203/output-only.md

The code snippet in the "Disabling" category has the comment line "// Immutable. The title of the book." which does not apply to that example. My guess is that this code example was originally copied from 203/immutable and this line was missed. 

This is also the only disabling example in 203 where there is a comment between the not-precedent comment and the field in question.

lukesneeringer:
Fixes #547.

### DIFF
--- a/docs/rules/0203/output-only.md
+++ b/docs/rules/0203/output-only.md
@@ -67,10 +67,10 @@ Remember to also include an [aip.dev/not-precedent][] comment explaining why.
 message Book {
   string name = 1;
 
-  // Immutable. The title of the book.
+  // @OutputOnly
+  // When the book was published.
   // (-- api-linter: core::0203::output-only=disabled
   //     aip.dev/not-precedent: We need to do this because reasons. --)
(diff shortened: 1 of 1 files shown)


docs/rules/0203/output-only.md
```diff
@@ -67,10 +67,10 @@ Remember to also include an [aip.dev/not-precedent][] comment explaining why.
 message Book {
   string name = 1;
 
-  // Immutable. The title of the book.
+  // @OutputOnly
```
noahdietz, line 70: s/@OutputOnly/Output only. ? Is `@OutputOnly` a new form of documenting "Output only."?

lukesneeringer: No, it is a very old form, that we want to discourage. That's why it is in the "if you are suppressing the linter" section.

PR #553, opened 2020-06-03
[AIP-146] Discourage Any.

lukesneeringer:
Note: Do not merge this until AIP-146 itself is merged.

This also refactors the common proto check in AIP-215 into a
utility function so it can be reused here.

### DIFF
--- a/docs/rules/0146/any.md
+++ b/docs/rules/0146/any.md
@@ -1,0 +1,55 @@
+---
+rule:
+  aip: 146
+  name: [core, '0146', any]
+  summary: Avoid `google.protobuf.Any` fields.
+permalink: /146/standardized-codes
+redirect_from:
+  - /0146/standardized-codes
+---
+
+# Any
+
+This rule discourages the use of `google.protobuf.Any`, as encouraged in
+[AIP-146][].
+
+## Details
+
+This rule complains if it sees a `google.protobuf.Any` field. Common packages
+(such as `google.api` or `google.longrunning`) are excluded.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message Book {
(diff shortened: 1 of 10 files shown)


docs/rules/0146/any.md
```diff
@@ -0,0 +1,55 @@
...
+  name: [core, '0146', any]
+  summary: Avoid `google.protobuf.Any` fields.
+permalink: /146/standardized-codes
+redirect_from:
+  - /0146/standardized-codes
+---
+
+# Any
+
+This rule discourages the use of `google.protobuf.Any`, as encouraged in
```
noahdietz, line 13: Nit: I'd probably choose the words used differently - "discourages X..., as encouraged in Y" sounds like this rule contradicts Y regarding X.

How about:
```
This rules discourages the use of `google.protobuf.Any`, as described in [AIP-146][].
```

rules/aip0146/any_test.go
```diff
@@ -0,0 +1,50 @@
...
+func TestAny(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		Package  string
+		Type     string
+		problems testutils.Problems
+	}{
+		{"Valid", "foo", "string", nil},
+		{"IgnoredCommon", "google.longrunning", "google.protobuf.Any", nil},
+		{"InvalidAny", "foo", "google.protobuf.Any", testutils.Problems{{Message: "Any"}}},
```
noahdietz, line 32: Will the message really just be `"Any"`? I thought it would be `"Avoid using google.protobuf.Any fields in public APIs."`?

lukesneeringer: The way `testutils.Problems` works is it uses `strings.Contains` logic. This avoids having tests that are too precise and veer toward change detection.

PR #559, opened 2020-06-15
[AIP-216] Add lint rule for common state value synonyms.

lukesneeringer:
This rule catches common State mistakes like `SUCCESSFUL` and
`SUCCESS` and corrects them.

It also catches `CANCELED` and asks for the two-L spelling:
`CANCELLED`. Both are correct, but we should use one consistently
and the AIP uses the two-L version.

### DIFF
--- a/rules/aip0216/value_synonyms_test.go
+++ b/rules/aip0216/value_synonyms_test.go
@@ -1,0 +1,52 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


rules/aip0216/value_synonyms_test.go
```diff
@@ -0,0 +1,52 @@
...
+		problems  testutils.Problems
+	}{
+		{"ValidSucceeded", "State", "SUCCEEDED", nil},
+		{"ValidDeleted", "State", "DELETED", nil},
+		{"InvalidSuccess", "State", "SUCCESS", testutils.Problems{{Suggestion: "SUCCEEDED"}}},
+		{"InvalidSuccessful", "State", "SUCCESSFUL", testutils.Problems{{Suggestion: "SUCCEEDED"}}},
+		{"InvalidCanceled", "State", "CANCELED", testutils.Problems{{Suggestion: "CANCELLED"}}},
+		{"InvalidFail", "State", "FAIL", testutils.Problems{{Suggestion: "FAILED"}}},
+		{"InvalidFailure", "State", "FAILURE", testutils.Problems{{Suggestion: "FAILED"}}},
+		{"Ireelevant", "Foo", "SUCCESSFUL", nil},
```
noahdietz, line 37: Typo: s/Ireelevant/Irrelevant

lukesneeringer:
```suggestion
		{"Irrelevant", "Foo", "SUCCESSFUL", nil},
```

PR #567, opened 2020-06-18
chore: Binary releases.

### LINKED ISSUE #53: How do we deal with versioning of the linter binary?

lukesneeringer:
This sets up CI/CD for binary releases when we make release tags.

Fixes #53.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,7 +46,7 @@ jobs:
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_path: ./api-linter-linux-amd64.tar.gz
-          asset_name: aip-linter-${{ steps.create_release.outputs.name }}-linux-amd64.tar.gz
+          asset_name: api-linter-${{ steps.create_release.outputs.tag_name }}-linux-amd64.tar.gz
           asset_content_type: application/tar+gzip
       - name: Upload the linux/arm release.
         uses: actions/upload-release-asset@v1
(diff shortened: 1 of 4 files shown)


lukesneeringer:
Okay, @noahdietz, I have this working now. It is not perfect but it is good enough. You can look at the GitHub Actions log to verify. (I deleted all of the releases I made while thrashing.)

Stuff to review:

- Please review the versioning disclaimer in README.md and make sure you think it is reasonable.
- Note how I am versioning the binary itself -- basically hot-swapping in `cmd/api-linter/version.go`. Someone who uses `go get` and `go run` will get the file in the repo; someone who downloads the binary will get the version number if they use `api-linter --version`.
- Filenames end up being `api-linter-{version}-{os}-{arch}.tar.gz`.
- The GitHub Workflow itself arguably has a flaw since it creates a release before doing anything else, meaning it will leave cleanup if a subsequent step fails. Given that a human needs to make the tag, I am okay with this _for now_.

Follow-on steps:
- Automatic release notes. Doing this in the workflow is easy but I need to produce the notes. Once we have this, I will make releases not start in draft.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Install the cross-platform build tool.
+        run: |
+          go get github.com/mitchellh/gox
+          go get github.com/inconshreveable/mousetrap
+      - name: Set the version number in the binary (for `api-linter --version`).
+        run: |
+          cat > cmd/api-linter/version.go <<EOF
+          package main
+          const version = "${{ needs.release.outputs.version }}"
+          EOF
```
noahdietz, line 61: Instead of rewriting a file on the fly, you might consider using Build-time Variable Injection. `gox` should support this compiler flag.

For example: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

lukesneeringer: Oh, nice. That does seem like a better way to do it. I will merge this first and then swap to that as a follow-up.

PR #573, opened 2020-06-20
chore: Add release-on-a-schedule.

lukesneeringer:
This makes us auto-release every Wednesday at 20:00 UTC (noon Pacific when in standard time), and is tolerant of weeks with no changes.

You can see the testing/thrashing in GitHub Actions. As in #567, I cleaned up after myself.

### DIFF
--- a/.github/release-tool/changelog.go
+++ b/.github/release-tool/changelog.go
@@ -1,0 +1,115 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


lukesneeringer:
@noahdietz Ping, but at your convenience. (I am just aware that I sent this on Friday night and it might have gotten lost.)

lukesneeringer:
Sure, seems reasonable. As usual for stuff like this, will handle in a followup.
The weekly schedule thing may not end up being the right idea either,  but it will work as a start.

PR #576, opened 2020-06-26
Allow plural `_times` suffix or singular `_time` suffix for repeated Timestamps.

klmilam:


rules/aip0142/time_field_names.go
```diff
@@ -44,13 +44,21 @@ var fieldNames = &lint.FieldRule{
...
+					Descriptor: f,
+					Location:   locations.DescriptorName(f),
+				}}
+			} else if !strings.HasSuffix(f.GetName(), "_times") {
+				return []lint.Problem{{
+					Message:    "Repeated Timestamp fields should end in `_time` or `_times`.",
+					Descriptor: f,
+					Location:   locations.DescriptorName(f),
+				}}
+			}
```
noahdietz, line 61: Nit, small refactoring: If we move the (repeated && non-plural-suffix) case to a leading check, it reads a bit more explicitly and leaves the non-repeated case as the "default" behavior. This also reduces indentation of the basic case.

```suggestion
			if f.IsRepeated() && !strings.HasSuffix(f.GetName(), "_times") {
				return []lint.Problem{{
					Message:    "Repeated Timestamp fields should end in `_time` or `_times`.",
					Descriptor: f,
					Location:   locations.DescriptorName(f),
				}}
			} 
			
			return []lint.Problem{{
					Message:    "Timestamp fields should end in `_time`.",
					Descriptor: f,
					Location:   locations.DescriptorName(f),
			}}
```

rules/aip0142/time_field_names_test.go
```diff
@@ -31,6 +31,8 @@ func TestFieldName(t *testing.T) {
 		{"InvalidIsMistake", "google.protobuf.Timestamp", "created", testutils.Problems{{Suggestion: "create_time"}}},
 		{"InvalidContainsMistake", "google.protobuf.Timestamp", "last_modified", testutils.Problems{{Suggestion: "update_time"}}},
 		{"InvalidNoSuffix", "google.protobuf.Timestamp", "create", testutils.Problems{{Message: "should end"}}},
+		{"InvalidNoSuffixRepeatedPlural", "repeated google.protobuf.Timestamp", "sample_times", testutils.Problems{}},
+		{"InvalidNoSuffixRepeated", "repeated google.protobuf.Timestamp", "sample_time", testutils.Problems{}},
```
noahdietz, line 35: These look like `Valid` test cases (there are no `Problems` emitted), but the test case is named `Invalid`. Should these have invalid test data that trigger a `Problem` or should they be green path test cases with a `Valid` name?

klmilam: Whoops. These are valid cases. Fixed the names and added an invalid test case.

rules/aip0142/time_field_names.go
klmilam: Had to make a slight change.

The suggestion would return an error in the case where we have a repeated field with a suffix of "_times", which is not intended behavior.

klmilam:
I haven't proposed a change to the AIP. Should I do so?

I originally proposed this change in cl/318410399 and Luke requested that I make the change here.

rules/aip0142/time_field_names_test.go
noahdietz: Thanks looks great.

rules/aip0142/time_field_names.go
noahdietz: Oh oops, yeah, you're right, that falls through on valid plural timestamps 🤦  Apologies! In that case I think what you had before was better, and I apologize for the churn.

noahdietz:
> I haven't proposed a change to the AIP. Should I do so?

Yeah, I think it's the right thing to do, especially for API reviewers that might not be aware that this is now allowed, they will see it when reviewing the AIP. 

https://github.com/googleapis/aip

rules/aip0142/time_field_names.go
noahdietz: I realized my previous comment might not have been clear, but please revert back to the original logic, it was better :)

klmilam: Done!

klmilam:
Opened PR https://github.com/googleapis/aip/pull/539

PR #581, opened 2020-07-15
fix: Exclude "main" suffices from version checks.

lukesneeringer:


lukesneeringer:
This exception is around internal tooling; what is publicly released still corresponds to AIP-215.

Ultimately the right way to do this sort of thing is with internal extensions but we do not have a good internal way to add a granular exception like this so I am accepting the wart for the moment.

PR #585, opened 2020-07-29
fix: BatchGet rule now singularizes correctly.

### LINKED ISSUE #575: AIP 231's response-resource-field does not recognize matches as the plural of match.
message BatchGetMatchesResponse {
  repeated Match matches = 1;
}

That gets flagged with this error:
Message "BatchGetMatchesResponse" has no "Matche" type field

lukesneeringer:
Fixes #575.

### DIFF
--- a/rules/aip0231/response_resource_field.go
+++ b/rules/aip0231/response_resource_field.go
@@ -31,7 +31,8 @@ var resourceField = &lint.MessageRule{
 		// Note: Use m.GetName()[8 : len(m.GetName())-9] to retrieve the resource
 		// name from the the batch get response, for example:
 		// "BatchGetBooksResponse" -> "Books"
-		resourceMsgName := pluralize.NewClient().Singular(m.GetName()[8 : len(m.GetName())-9])
+		plural := m.GetName()[8 : len(m.GetName())-8]
+		resourceMsgName := pluralize.NewClient().Singular(plural)
 
 		for _, fieldDesc := range m.GetFields() {
 			if msgDesc := fieldDesc.GetMessageType(); msgDesc != nil && msgDesc.GetName() == resourceMsgName {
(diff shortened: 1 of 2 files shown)


lukesneeringer:
@noahdietz: I am out today, so please just merge this on my behalf (assuming you approve/agree).

rules/aip0231/response_resource_field.go
```diff
@@ -31,7 +31,8 @@ var resourceField = &lint.MessageRule{
 		// Note: Use m.GetName()[8 : len(m.GetName())-9] to retrieve the resource
 		// name from the the batch get response, for example:
 		// "BatchGetBooksResponse" -> "Books"
-		resourceMsgName := pluralize.NewClient().Singular(m.GetName()[8 : len(m.GetName())-9])
+		plural := m.GetName()[8 : len(m.GetName())-8]
```
noahdietz, line 34: The comment is helpful, but why not use `strings.TrimPrefix(n, "BatchGet")` and `strings.TrimSuffix(n, "Response")`? It would end up being more than one line, but it would be more readable and we wouldn't need magic numbers or such a detailed comment. I think the magic number was part of the cause of this bug too anyways.

rules/aip0231/response_resource_field.go
```diff
@@ -31,7 +31,8 @@ var resourceField = &lint.MessageRule{
 		// Note: Use m.GetName()[8 : len(m.GetName())-9] to retrieve the resource
```
noahdietz, line 31:
```suggestion
		// Note: Use m.GetName()[8 : len(m.GetName())-8] to retrieve the resource
```

Comment doesn't match the implementation.

rules/aip0231/response_resource_field.go
lukesneeringer: Fair point. Done.

rules/aip0231/response_resource_field.go
lukesneeringer: No longer relevant.

Issue #612, opened 2020-08-27
Missing lints for AIP-235


apasel422:
http://google.aip.dev/235

- [x] HTTP body
- [x] HTTP method
- [x] HTTP URI suffix
- [x] plural method name
- [x] request message name
- [x] request `parent` field
- [x] request `parent` field resource reference
- [x] request `names` field
- [x] request `names` field behavior
- [x] request `names` field resource reference
- [x] request `requests` field
- [x] request `requests` field behavior
- [x] request message unknown fields
- [x] response message name
- [x] soft-delete response resource field

PR #625, opened 2020-09-16
feat: Add initial rules for AIP-235.

apasel422:
These are adapted from AIP-135 and AIP-231.

We still need lints for the following, and possibly others, for #612:
- request URI
- request message resource reference
- request message unknown fields
- response message


lukesneeringer (approved):
Thanks for doing this.

Note that the tests might be easier for you to write with `testutils.ParseProto3Tmpl`, but not going to block the PR on that.

PR #642, opened 2020-10-04
fix: Permit two-word top-level collections in the URI suffix.

### LINKED ISSUE #622: AIP-136. Custom methods: URI suffix. False Detection
I have method like this one. I have a *collection* of users that should be populated from the external source "google users":
```
  rpc SyncGoogleUsers(google.protobuf.Empty) returns (google.protobuf.Empty) {
    option (google.api.http) = {
      post: "/v1/users:syncGoogleUsers"
      body: "*"
    };
  }
```

It leads to false detection when a custom method for the whole collection instead of an item from this collection and this item has similar name `User` vs `GoogleUser`.

```
problems:
  - message: Custom method should have a URI suffix matching the method name, such as "users:sync".
    location:
      start_position:
        line_number: 34
        column_number: 5
      end_position:
        line_number: 37
        column_number: 6
    rule_id: core::0136::http-uri-suffix
    rule_doc_uri: https://linter.aip.dev/136/http-uri-suffix
```

lukesneeringer:
Another tweak to the perpetually-difficult `http-uri-suffix` rule from AIP-136.

Fixes #622.

### DIFF
--- a/rules/aip0136/http_uri_suffix_test.go
+++ b/rules/aip0136/http_uri_suffix_test.go
@@ -27,6 +27,7 @@ func TestURISuffix(t *testing.T) {
 		URI        string
 		problems   testutils.Problems
 	}{
+		{"ValidTwoWordNoParent", "SearchAudioBooks", "/v1/audioBooks:search", nil},
 		{"ValidVerb", "ArchiveBook", "/v1/{name=publishers/*/books}:archive", testutils.Problems{}},
 		{"ValidVerbNested", "ArchiveBook", "/v1/{book.name=publishers/*/books}:archive", testutils.Problems{}},
 		{"ValidVerbParent", "ImportBooks", "/v1/{parent=publishers/*}/books:import", testutils.Problems{}},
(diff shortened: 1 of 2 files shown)


rules/aip0136/http_uri_suffix_test.go
```diff
@@ -27,6 +27,7 @@ func TestURISuffix(t *testing.T) {
 		URI        string
 		problems   testutils.Problems
 	}{
+		{"ValidTwoWordNoParent", "SearchAudioBooks", "/v1/audioBooks:search", nil},
```
noahdietz, line 30: Nit: Why `nil` instead of `testutils.Problems{}` like all of the other `Valid` tests?

noahdietz (approved):
One nit but LGTM

rules/aip0136/http_uri_suffix_test.go
lukesneeringer: I have been shifting toward `nil` of late. At some point I will change them all.

PR #647, opened 2020-10-07
feat: Add IsDeclarativeFriendly utility method.

lukesneeringer:
This will be needed for upcoming declarative-friendly lint checks.

### DIFF
--- a/rules/internal/utils/declarative_friendly.go
+++ b/rules/internal/utils/declarative_friendly.go
@@ -79,7 +79,7 @@ func IsDeclarativeFriendly(d desc.Descriptor) bool {
 		// If this is a Delete method (AIP-135) with a return value of Empty,
 		// try to find the resource.
 		if strings.HasPrefix(m.GetName(), "Delete") && m.GetOutputType().GetName() == "Empty" {
-			if resource := findMessage(m.GetFile(), m.GetName()[6:]); resource != nil {
+			if resource := findMessage(m.GetFile(), strings.TrimPrefix(m.GetName(), "Delete")); resource != nil {
 				return IsDeclarativeFriendly(resource)
 			}
 		}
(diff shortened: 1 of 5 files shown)


rules/internal/utils/declarative_friendly.go
```diff
@@ -0,0 +1,130 @@
...
+				if field.IsRepeated() && field.GetMessageType() != nil {
+					return IsDeclarativeFriendly(field.GetMessageType())
+				}
+			}
+		}
+
+		// If this is a Delete method (AIP-135) with a return value of Empty,
+		// try to find the resource.
+		if strings.HasPrefix(m.GetName(), "Delete") && m.GetOutputType().GetName() == "Empty" {
+			if resource := findMessage(m.GetFile(), m.GetName()[6:]); resource != nil {
```
noahdietz, line 82: Nit: Can you comment what this does or convert the magic number into a readable constant?

I think this is attempting to strip the `Delete` prefix to get the resource `message` name.

rules/internal/utils/declarative_friendly.go
```diff
@@ -0,0 +1,130 @@
...
+				return IsDeclarativeFriendly(resource)
+			}
+		}
+
+		// At this point, we probably have a custom method.
+		// Try to identify a resource by whittling away at the method name and
```
noahdietz, line 88: Additionally/alternatively, could we also look at the request message for a field annotated with `resource_reference` and attempt to resolve the resource that way?

noahdietz (approved):

rules/internal/utils/declarative_friendly.go
noahdietz: Better yet: use `strings.TrimPrefix(m.GetName(), "Delete")`

lukesneeringer: Done.

rules/internal/utils/declarative_friendly.go
lukesneeringer: Agree that it is another potential venue, although custom methods are gnarly because there is no guarantee they will have that or that it will refer to the correct resource.

PR #651, opened 2020-10-08
feat: Add `allow_missing` rule to AIP-134.

lukesneeringer:
Note: This rule applies to declarative-friendly resources only.

### DIFF
--- a/rules/aip0134/request_allow_missing_field_test.go
+++ b/rules/aip0134/request_allow_missing_field_test.go
@@ -21,6 +21,7 @@ import (
 )
 
 func TestAllowMissing(t *testing.T) {
+	problems := testutils.Problems{{Message: "include `bool allow_missing`"}}
 	for _, test := range []struct {
 		name         string
 		Style        string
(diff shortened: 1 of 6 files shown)


rules/aip0134/request_allow_missing_field_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestAllowMissing(t *testing.T) {
+	for _, test := range []struct {
+		name         string
+		Style        string
+		AllowMissing string
+		problems     testutils.Problems
+	}{
+		{"IgnoredNotDF", "", "", nil},
+		{"ValidIncluded", "style: DECLARATIVE_FRIENDLY", "bool allow_missing = 2;", nil},
+		{"Invalid", "style: DECLARATIVE_FRIENDLY", "", testutils.Problems{{Message: "include `bool allow_missing`"}}},
```
noahdietz, line 32: Nit: maybe include an invalid type for `allow_missing` test? technically that condition is not covered by these tests.

```suggestion
		{"Invalid", "style: DECLARATIVE_FRIENDLY", "", testutils.Problems{{Message: "include `bool allow_missing`"}}},
		{"InvalidType", "style: DECLARATIVE_FRIENDLY", "string allow_missing = 2;", testutils.Problems{{Message: "include `bool allow_missing`"}}},
```

noahdietz (approved):

rules/aip0134/request_allow_missing_field_test.go
lukesneeringer: Good point. Done.

PR #675, opened 2020-10-13
Add example for a violation using additional_bindings

cledoux:
Why:
- I was violating the linter in this way and this documentation did not help me learn why.

Changed:
- Added an incorrect/correct pair that uses additional_bindings


lukesneeringer (approved):
Thanks for this!

PR #676, opened 2020-10-13
feat: Add rules for AIP-164 Undelete

### LINKED ISSUE #666: AIP-164: Undelete methods must contain name.

apasel422:
Fixes #666, #667, #668.

### DIFF
--- a/docs/rules/0164/http-body.md
+++ b/docs/rules/0164/http-body.md
@@ -1,0 +1,67 @@
+---
+rule:
+  aip: 164
+  name: [core, '0164', http-body]
+  summary: Undelete methods should use `*` as the HTTP body.
+permalink: /164/http-body
+redirect_from:
+  - /0164/http-body
+---
+
+# Undelete methods: HTTP body
+
+This rule enforces that all `Undelete` RPCs use `*` as the HTTP `body`, as mandated in
+[AIP-164][].
+
+## Details
+
+This rule looks at any message beginning with `Undelete`, and complains
+if the HTTP `body` field is anything other than `*`.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+rpc UndeleteBook(UndeleteBookRequest) returns (Book) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:undelete"
+    body: ""  // This should be "*".
+  };
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+rpc UndeleteBook(UndeleteBookRequest) returns (Book) {
+  option (google.api.http) = {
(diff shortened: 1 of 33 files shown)


docs/rules/0164/http-body.md
```diff
@@ -0,0 +1,67 @@
...
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+rpc UndeleteBook(UndeleteBookRequest) returns (Book) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:undelete"
+    body: ""  // This should be "*".
```
lukesneeringer, line 30:
```suggestion
    // body: "*" should be set.
```

docs/rules/0164/http-body.md
```diff
@@ -0,0 +1,67 @@
...
+If you need to violate this rule, use a leading comment above the method.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+// (-- api-linter: core::0164::http-body=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+rpc UndeleteBook(UndeleteBookRequest) returns (Book) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:undelete"
+    body: ""
```
lukesneeringer, line 58:
```suggestion
```

rules/aip0164/http_body_test.go
```diff
@@ -0,0 +1,70 @@
...
+	}{
+		{"Valid", "*", "UndeleteBook", ""},
+		{"Invalid", "", "UndeleteBook", "HTTP body"},
+		{"Irrelevant", "", "AcquireBook", ""},
+	}
+
+	for _, test := range tests {
+		t.Run(test.testName, func(t *testing.T) {
+			// Create a MethodOptions with the annotation set.
+			opts := &dpb.MethodOptions{}
```
lukesneeringer, line 42: I am surprised you went this route and not `testutils.ParseProto3Tmpl`, but it is fine as it is.

lukesneeringer (approved):

rules/aip0164/http_body_test.go
apasel422: I just copied it from one of the other tests that was similar. It'd be good at some point to upgrade all the tests to use `ParseProto3Tmpl`.

lukesneeringer: Ah, that makes sense. Yeah, I wrote about 20 or so rules before writing `ParseProto3Tmpl`, which required me to get a change in to the underlying protoreflect library.

apasel422: I really miss `g4 cp` in GitHub's PR UI, to make these changes more obvious.

PR #677, opened 2020-10-13
Add validate_only field to AIP-163 request-unknown-fields doc

apasel422:
It's permitted at https://github.com/googleapis/api-linter/blob/005f6feace6f363c94bf559fe9845bbba964688e/rules/aip0135/request_unknown_fields.go#L35.


lukesneeringer (approved):
Good catch, thanks.

PR #687, opened 2020-10-14
refactor: Extract common lints into the util package to eliminate redundancy

apasel422:


rules/internal/utils/common_lints.go
```diff
@@ -0,0 +1,96 @@
...
+	if ref := GetResourceReference(f); ref == nil {
+		return []lint.Problem{{
+			Message:    fmt.Sprintf("The `%s` field should include a `google.api.resource_reference` annotation.", f.GetName()),
+			Descriptor: f,
+		}}
+	}
+	return nil
+}
+
+func lintHTTPBody(m *desc.MethodDescriptor, want, msg string) []lint.Problem {
```
lukesneeringer, line 61: Perhaps just expose this and use it within the individual rules?

rules/internal/utils/common_lints.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/builder"
+)
+
+// LintStringField returns a problem if the field is not a string.
+func LintStringField(f *desc.FieldDescriptor) []lint.Problem {
```
lukesneeringer, line 27: Perhaps `LintFieldType` that takes the expected type?

lukesneeringer (commented):
This is a good idea that I had mused about doing. In some cases, we might be able to stop writing long-winded tests each time, too.

I do think we need `common_lints_test.go`, regardless of whether we eliminate individual tests.

rules/internal/utils/common_lints.go
apasel422: I'd rather wait until it's needed, as it will make the message-generation more complicated.

rules/internal/utils/common_lints.go
apasel422: I like the explicitness of the separate functions, which also helps with message-generation simplicity, but I defer to you.

lukesneeringer (approved):

Issue #344, opened 2019-11-14
[Feature Request] Add (optional) fixes to the returned problems


mingzhi:
Currently, we are returning an (optional) `Suggestion` in a problem. But a suggestion is not a fix. For example, in the rule `core::0131::request-message-name`:
```protobuf
rpc GetBook(GetBookReq) returns (Book) {  // Should be `GetBookRequest`.
  option (google.api.http) = {
    get: "/v1/{name=publishers/*/books/*}"
  };
}

message GetBookReq {}
```
we will suggest changing `GetBookReq` to `GetBookRequest`. However, the fix should not only apply to the input type, but also to the message name. Therefore, there are **two** replacements in the same file or two separate files.

We need to add a slice of fixes into the returned problem:
```golang
type Fix struct {
  Description string // a human readable description that may be shown in a UI.
  Replacements []Replacement
}

type Replacement struct {
  Location Location
  NewContent string
}

type Location struct {
  FilePath string
  Range TextRange
}

type TextRange struct {
  Start Position
  End Position
}

type Position struct {
  Line, Column int
}
```

lukesneeringer:
Is this even a thing that Tricorder supports?

apasel422:
Yes, Tricorder does support fixes.

PR #694, opened 2020-10-19
Add optional fixes and demonstrate them in AIP-128 resource-annotations-field rule

apasel422:
Part of #344.


apasel422:
@mingzhi I'm not sure if this is what you had in mind, so please provide suggestions.

We will likely want some helper functions to produce these suggestions, similar to how there are location helpers, particularly for things like adding a new field to a message or removing an existing one. I'm not sure what the best approach for that is.

Issue #698, opened 2020-10-21
Missing rules for AIP-152


apasel422:
- `Run` method
  - [x] HTTP body
  - [x] HTTP method
  - [x] HTTP URI
  - [x] Request message name
  - [x] Response message name
- `Run` request
  - [x] `name` field
  - [x] `name` field `field_behavior` annotation
  - [x] `name` field `resource_reference` annotation
- [x] Job resource message name ends with `Job`

PR #707, opened 2020-11-02
feat: Add rules for AIP-152

apasel422:
Part of #698.


lukesneeringer:
Just got back from vacation. Will read this soon.

lukesneeringer (approved):

PR #738, opened 2021-01-29
feat: Require URIs to begin with a forward slash.

lukesneeringer:


noahdietz (approved):

noahdietz:
Per quality_checker, need to register the rule.

lukesneeringer:
And that is why we have him. :-)

PR #754, opened 2021-02-10
feat: validate list parent resource_reference

noahdietz:
This lints List*Request parent resource_reference annotations to check it does not reference the resource being listed.

### DIFF
--- a/docs/rules/0132/request-parent-valid-reference.md
+++ b/docs/rules/0132/request-parent-valid-reference.md
@@ -1,0 +1,77 @@
+---
+rule:
+  aip: 132
+  name: [core, '0132', request-parent-valid-reference]
+  summary: |
+    List RPCs should reference the parent resource, not the
+    listed resource.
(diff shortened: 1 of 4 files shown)


lukesneeringer (changes requested):
I wonder if we should go stronger and say to always use `child_type` with the resource.

noahdietz:
> I wonder if we should go stronger and say to always use `child_type` with the resource.

I think that would be reasonable, but previous advice has been that we can do either. There would be many API's in "violation" of this rule today if we enforced only `child_type`. Not a reason we shouldn't do it if we think it's right, but there was some merit to our existing practice that we will have to go back on.

I'm happy to change it if you think that is better though. LMK one way or the other.

lukesneeringer:
That is a fair point. Okay.

lukesneeringer (approved):

lukesneeringer (approved):

PR #794, opened 2021-03-13
Fixes 0203/unordered-list-repeated links

PeytonT:
Currently, 0203/unordered-list-repeated.md has the same permalink as 0203/optional.md

As a result, https://linter.aip.dev/203/optional displays information about UNORDERED_LIST instead of about OPTIONAL, and https://linter.aip.dev/203/unordered-list-repeated is a 404.


lukesneeringer (approved):
Thanks for fixing that!

PR #797, opened 2021-03-16
feat: AIP-123 – Complain if resource references are not strings.

lukesneeringer:
/cc @jskeet

### DIFF
--- a/docs/rules/0123/resource-reference-type.md
+++ b/docs/rules/0123/resource-reference-type.md
@@ -1,0 +1,68 @@
+---
+rule:
+  aip: 123
+  name: [core, '0123', resource-reference-type]
+  summary: Resource reference annotations should only apply to strings.
+permalink: /123/resource-reference-type
+redirect_from:
+  - /0123/resource-reference-type
+---
+
+# Resource annotation presence
+
+This rule enforces that any field with a `google.api.resource_reference`
+annotation has a `string` type, as described in [AIP-123][].
+
+## Details
+
+This rule scans all fields with a `google.api.resource_reference` annotation.
+If one is found, the type is checked, and the rule complains if the type is
+anything other than `string`.
+
+It suggests the removal of the annotation rather than fixing the type, because
+what we have observed in real life is that the annotation is usually what is
+in error rather than the selected type.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message Book {
+  string name = 1;
+
+  // This is not a resource reference; the annotation does not belong.
+  Author author = 2 [(google.api.resource_reference) = {
+    type: "library.googleapis.com/Author"
+  }];
+}
+```
+
+**Correct** code for this rule:
+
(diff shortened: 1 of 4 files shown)


docs/rules/0123/resource-reference-type.md
```diff
@@ -0,0 +1,68 @@
...
+  Author author = 2 [(google.api.resource_reference) = {
+    type: "library.googleapis.com/Author"
+  }];
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Incorrect.
```
apasel422, line 45:
```suggestion
// Correct.
```

docs/rules/0123/resource-reference-type.md
```diff
@@ -0,0 +1,68 @@
...
+message Book {
+  string name = 1;
+
+  string author = 2 [(google.api.resource_reference) = {
+    type: "library.googleapis.com/Author"
+  }];
+}
```
apasel422, line 61: Remove this second `message Book`.

rules/aip0123/resource_reference_type_test.go
```diff
@@ -0,0 +1,52 @@
...
+	annotation := ` [(google.api.resource_reference) = {
+		type: "library.googleapis.com/Author"
+	}]`
+	for _, test := range []struct {
+		name       string
+		Type       string
+		Annotation string
+		problems   testutils.Problems
+	}{
+		{"Valid", "string", annotation, nil},
```
apasel422, line 33: Add a valid case for `repeated string` as well.

rules/aip0123/resource_reference_type_test.go
```diff
@@ -0,0 +1,52 @@
...
+			  import "google/api/resource.proto";
+				message Book {
+					{{.Type}} author = 1{{.Annotation}};
+				}
+				message Author {}  // For the failure case, to make compilation work.
+			`, test)
+			field := file.GetMessageTypes()[0].GetFields()[0]
+			got := resourceReferenceType.Lint(file)
+			if diff := test.problems.SetDescriptor(field).Diff(got); diff != "" {
+				t.Errorf(diff)
```
apasel422, line 48:
```suggestion
				t.Error(diff)
```

apasel422 (approved):

docs/rules/0123/resource-reference-type.md
lukesneeringer: No, this one is indeed correct, and a valid thing that someone might want to do (even if it is not the most likely problem someone is hitting if they break this rule).

rules/aip0123/resource_reference_type_test.go
lukesneeringer: Done.

docs/rules/0123/resource-reference-type.md
apasel422: In that case, shouldn't the comment above the message be `// Correct`?

PR #798, opened 2021-03-17
refactor: Replace individual GET/DELETE HTTP body rules with one for AIP-127

apasel422:


lukesneeringer (changes requested):
I think I would prefer not do this, because it would break rule disabling (and make it less precise).

apasel422:
> I think I would prefer not do this, because it would break rule disabling (and make it less precise).

Good point. I guess it would be overkill to add support for rule-disabling synonyms, though they could also be useful for similar refactorings in the future?

lukesneeringer:
> Good point. I guess it would be overkill to add support for rule-disabling synonyms, though they could also be useful for similar refactorings in the future?

Actually, [we have this already](https://github.com/googleapis/api-linter/blob/main/lint/rule_aliases.go). So we could get around that.

The question is whether we want to replace the several rules with a single rule, which would prevent a way to disable just one of them at the service or file level.

Granted, this is an example of a rule where that is probably not very common...so maybe we can do this after all?

PR #808, opened 2021-03-23
chore: Run `go mod tidy` so build succeeds at HEAD

apasel422:


lukesneeringer:
Going to merge this as `chore:` instead of `fix:` so it stays out of the release notes.

Issue #702, opened 2020-10-29
Missing rules for AIP-132


apasel422:
- HTTP method
- HTTP body
- HTTP URI
- `parent` HTTP variable

PR #809, opened 2021-03-25
feat: Add rule for AIP-132 HTTP URI parent variable

apasel422:
Part of #702.


noahdietz (commented):
FWIW I know this will not work for one specific API - Cloud Resource Manager. This API owns Folders, Projects, and Organizations which are all top-level resources (i.e. `folders/{folder}`), but that can have a `parent` e.g. `organizations/{organization}/folders/{folder}`. Their List RPCs have a `parent`(which is annotated as an arbitrary resource `"*"`) but their HTTP bindings don't necessarily have a `parent` in them ([example](https://github.com/googleapis/googleapis/blob/15c5e21948ff6fbe41f91bdf04f6252f91a12d59/google/cloud/resourcemanager/v2/folders.proto#L55)). I think there might be one other API that defines a top-level resource, but it's escaping me.

I would suggest checking that the URI actually even has a spot for a resource name/path param - `/v2/folders` doesn't and therefore should be exempt.

rules/aip0132/http_uri_parent.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2019 Google LLC
```
noahdietz, line 1:
```suggestion
// Copyright 2021 Google LLC
```

rules/aip0132/http_uri_parent_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2019 Google LLC
```
noahdietz, line 1:
```suggestion
// Copyright 2021 Google LLC
```

apasel422:
> FWIW I know this will not work for one specific API - Cloud Resource Manager. This API owns Folders, Projects, and Organizations which are all top-level resources (i.e. `folders/{folder}`), but can have a `parent` e.g. `organizations/{organization}/folders/{folder}`. Their List RPCs have a `parent`(which is annotated as an arbitrary resource `"*"`) but their HTTP bindings don't necessarily have a `parent` in them ([example](https://github.com/googleapis/googleapis/blob/15c5e21948ff6fbe41f91bdf04f6252f91a12d59/google/cloud/resourcemanager/v2/folders.proto#L55)). I think there might be one other API that defines a top-level resource, but it's escaping me.
> 
> I would suggest checking that the URI actually even has a spot for a resource name/path param - `/v2/folders` doesn't and therefore should be exempt.

How can we distinguish between not having a spot for the resource name/path param and just forgetting to include the `parent` variable?

lukesneeringer:
> How can we distinguish between not having a spot for the resource name/path param and just forgetting to include the `parent` variable?

In other rules, I did it by checking for the presence of a `parent` field on the request message.

apasel422:
> > How can we distinguish between not having a spot for the resource name/path param and just forgetting to include the `parent` variable?
> 
> In other rules, I did it by checking for the presence of a `parent` field on the request message.

That's what I'm doing here.

lukesneeringer:
Cool. Noah, are you good with this PR?

noahdietz:
> Cool. Noah, are you good with this PR?

@lukesneeringer the edge-case I raised ([RPC](https://github.com/googleapis/googleapis/blob/15c5e21948ff6fbe41f91bdf04f6252f91a12d59/google/cloud/resourcemanager/v2/folders.proto#L55), [request message](https://github.com/googleapis/googleapis/blob/15c5e21948ff6fbe41f91bdf04f6252f91a12d59/google/cloud/resourcemanager/v2/folders.proto#L305-L315)) is still not addressed by the heuristic you suggest. Resources that _can be_ top-level, like Folder, would not pass this test.

> How can we distinguish between not having a spot for the resource name/path param and just forgetting to include the `parent` variable?

@apasel422 check the `google.api.resource` definition on the resource message in the response ([example](https://github.com/googleapis/googleapis/blob/15c5e21948ff6fbe41f91bdf04f6252f91a12d59/google/cloud/resourcemanager/v2/folders.proto#L257)). If it is a top-level resource, then _a binding_ should be allowed to not have a `parent` path param. A naive test of "top-level" is how many segments the pattern has - `folders/{folder}` `organizations/{organization}` `projects/{project}` are all top-level and only have two segments in them. 

Arguably, the example(s) I shared could have multiple patterns - one top-level resource, one parented by `organizations/{organization}`, but they do not document it as such nor do they have the HTTP Bindings for it.

This is not such a big deal, except I know there will be at least one team (the one I'm using as an example) that has to disable this rule for more than one version of their API and on multiple RPCs. Feel free to merge this as is if that is reasonable to you.

PR #810, opened 2021-03-26
feat: add client library rules

noahdietz:
feat(linter): add AIP-4232 rule with check for repeated fields

feat(docs): add client library section to linter docs

fix(docs): respect start/end in linter-group-listing

fix(docs): remove dupe breadcrumb


docs/_includes/linter-breadcrumb.html
```diff
@@ -11,18 +11,6 @@
...
-    <li class="glue-breadcrumbs__item" aria-level="2">
-      <a class="glue-breadcrumbs__link" href="{{ site.url }}/rules/">
-        Rule Documentation
-      </a>
-    </li>
-    <li class="glue-breadcrumbs__item" aria-level="2">
-      <a class="glue-breadcrumbs__link" href="{{ site.url }}/rules/core/">
-        Core rules
-      </a>
-    </li>
```
noahdietz, line 24: This was unnecessary because the following `{%- for segment in url_segments` logic does this in a more robust way.

docs/_includes/linter-group-listing.html
```diff
@@ -20,8 +20,7 @@
       <use xlink:href="#mi-arrow-circle"></use>
     </svg>
   </div>
-
-  {% assign rule_pages = site.pages | where_exp: "p", "p.rule != nil" | sort: "rule.name" | group_by: "rule.aip" -%}
+  {% assign rule_pages = site.pages | where_exp: "p", "p.rule != nil" | where_exp: "p", "p.rule.aip >= include.start" | where_exp: "p", "p.rule.aip <= include.end" | sort: "rule.name" | group_by: "rule.aip" -%}
```
noahdietz, line 23: This creates a subset of the pages that have AIP numbers within the start & end values given to this page.

noahdietz: Previously, the `start` and `end` params to this page were not being respected.

lukesneeringer (approved):
I would prefer these be in core AIPs, but I can not think of an AIP where they would be appropriate, so this makes sense. Thanks!

PR #822, opened 2021-04-01
fix: Suppress core::aip0158::request-page-token-field for non-request messages

apasel422:
Based on an internal bug report: If a message isn't actually a request but happens to have a `page_size` field, it shouldn't be required to have a `page_token` field. An example of this is a message that itself will be serialized as a page token, e.g.

```proto
message FooPageToken {
  int32 page_size = 1;
  int32 page_num = 2;
}
```

This seems to have been an oversight in #496.

It is debatable whether we should apply this to the corresponding `isPaginatedResponseMessage` function, though it seems less likely to come up.

### DIFF
--- a/rules/aip0158/aip0158.go
+++ b/rules/aip0158/aip0158.go
@@ -17,6 +17,7 @@ package aip0158
 
 import (
 	"regexp"
+	"strings"
 
 	"github.com/googleapis/api-linter/lint"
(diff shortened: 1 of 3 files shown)


lukesneeringer:
Oh, huh. Yeah, that is an interesting bug. Shocked it took two years for anyone to run into that.

lukesneeringer (approved):

PR #823, opened 2021-04-01
docs: Fix typo in core::0141::forbidden-types

apasel422:
Though it's not clear to me whether the suppression comment should be added over the field or message.

### DIFF
--- a/docs/rules/0141/forbidden-types.md
+++ b/docs/rules/0141/forbidden-types.md
@@ -53,10 +53,10 @@ If you need to violate this rule, use a leading comment above the field.
 Remember to also include an [aip.dev/not-precedent][] comment explaining why.
 
 ```proto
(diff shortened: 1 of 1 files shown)


docs/rules/0141/forbidden-types.md
```diff
@@ -49,7 +49,7 @@ message Book {
 
 ## Disabling
 
-If you need to violate this rule, use a leading comment above the method.
+If you need to violate this rule, use a leading comment above the message.
```
lukesneeringer, line 52:
```suggestion
If you need to violate this rule, use a leading comment above the field.
```

lukesneeringer (approved):
It should be field. Fixed it.

apasel422:
The example currently puts the comment above the message, though.

lukesneeringer:
Oh, thanks.

PR #838, opened 2021-04-09
docs: Replace message with method in AIP-152

ericwenn:
The rule looks at the method, not the message.


ericwenn:
Should the rule be called `response-message-lro` instead of `response-message-name`?

apasel422:
The rule actually does look at Run *methods* in order to check that the response matches the method name:

https://github.com/googleapis/api-linter/blob/eb7c4535a6a1c8b8b5a17a9565eca97c182b36b6/rules/aip0152/response_message_name.go#L26-L28

apasel422 (approved):

PR #846, opened 2021-05-15
feat: Support 'conditionally' required fields.

RocketDonkey:
google.aip.dev issue: https://github.com/aip-dev/google.aip.dev/issues/411

Given a field:

```proto
message ExportBookRequest {
  bool enable_encryption = 1;
  // Required if enable_encryption is set to true.
  string encryption_key = 2;
}
```

`encryption_key` should not be annotated with `REQUIRED` since it is
only necessary if `enable_encryption` is set. This change ensures that
if a field comment contains 'required if', it will not trigger a linter
error.

Note there are other phrases that could logically apply (e.g. 'Only
required when'), but the resulting regex would be difficult to maintain.

### DIFF
--- a/docs/rules/0203/required.md
+++ b/docs/rules/0203/required.md
@@ -18,6 +18,10 @@ machine-readable annotation, as mandated by [AIP-203][].
 This rule looks at any field with "required" (or similar forms) in the comment,
 and complains if it does not have a `google.api.field_behavior` annotation.
 
+**Note:** if a field is conditionally required (i.e. it is only required if
+another field is set), the phrases 'Required if' or 'Required when' in the
+comment will disable this rule.
+
 ## Examples
 
 **Incorrect** code for this rule:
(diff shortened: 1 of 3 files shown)


RocketDonkey:
Happy to adjust if this approach isn't correct (or drop it if you'd rather hold off on supporting this).

noahdietz:
@RocketDonkey would it be sufficient to just disable the api-linter rule for the fields in your API that are conditionally required? https://linter.aip.dev/203/required#disabling. 

IMO, your suggestion is an OK one and I've seen this case before in API reviews, but we'd have to codify this and start over at https://github.com/aip-dev/google.aip.dev suggesting an amendment to the AIP before adding it to the linter. I'd say though that simply disabling the linter rule in these cases should be sufficient, WDYT?

lukesneeringer (approved):

RocketDonkey:
@noahdietz Yeah, I am totally fine with following whatever process makes sense. I didn't initially propose a change to the [AIP](https://github.com/aip-dev/google.aip.dev/blob/master/aip/general/0203.md) because I interpreted the AIP as guidance on when to apply `field_behavior`s, while this change relaxes a linter condition when checking how those behavior were/weren't applied.

However more than happy to propose an amendment, and thanks for the quick feedback!

PR #847, opened 2021-05-17
fix: Remove the valid resource reference rule.

lukesneeringer:
It is becoming increasingly clear that this rule causes more
harm than good due to many false positives.


noahdietz (approved):

noahdietz:
Bummer :(

PR #878, opened 2021-09-06
build: upgrade to Go 1.17

Juneezee:
1. Update the go directive in `go.mod` to 1.17
2. Update GitHub workflows to use Go 1.17
3. Replace the existing `io/ioutil` functions with their new definitions in `io` and `os` packages. The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil).

No breaking changes. Everything should still work as expected.

### DIFF
--- a/.github/quality-checker/rule_documented.go
+++ b/.github/quality-checker/rule_documented.go
@@ -16,13 +16,13 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
+	"os"
 )
 
 func checkRuleDocumented(aip int, name string) []error {
 	// Ensure that the expected documentation file exists.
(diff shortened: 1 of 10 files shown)


noahdietz (commented):
Looks great, happy to merge this. Thanks for the contribution. Give me some time to verify our internal infrastructure that runs this will run with the appropriate Go version :)

noahdietz:
After considering this some more, I'm not sure we can accept this PR at the moment. I would like to retain users of previous Go versions, especially because versions >= 1.13 are still very popular. Thank you so much for taking the time though.

shwoodard:
We will reopen when we're ready to upgrade.

PR #880, opened 2021-10-04
fix(AIP-126): allow first enum value UNKNOWN

### LINKED ISSUE #871: AIP-126: Linter complains about having UNKNOWN as the first value
Linter complains about not having [UNSPECIFIED](https://linter.aip.dev/126/unspecified) as the zero value even if `UNKNOWN` is specified as the first enum value. 
[AIP-126](https://google.aip.dev/126#guidance) allows having `UNKNOWN` as the zero value and it is a meaningful default for our use case. Can the linter be updated to allow this value?

Thanks.

noahdietz:
Fixes #871 by allowing `UNKNOWN` as the first value in an enum.

Note: If this rule emits a problem, it will suggest using the `_UNSPECIFIED` format for the first value and not `UNKNOWN`. Using `UNKNOWN` is an exception for a specific case and that should be discussed with reviewers/referenced from the AIP.

### DIFF
--- a/rules/aip0126/unspecified_test.go
+++ b/rules/aip0126/unspecified_test.go
@@ -29,9 +29,9 @@ func TestUnspecified(t *testing.T) {
 	}{
 		{"Valid", "BookFormat", "BOOK_FORMAT_UNSPECIFIED", testutils.Problems{}},
 		{"ValidWithNum", "Ipv6Format", "IPV6_FORMAT_UNSPECIFIED", nil},
+		{"ValidUnknown", "BookFormat", "UNKNOWN", nil},
 		{"InvalidNoPrefix", "BookFormat", "UNSPECIFIED", testutils.Problems{{Suggestion: "BOOK_FORMAT_UNSPECIFIED"}}},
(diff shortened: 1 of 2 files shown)


rules/aip0126/unspecified_test.go
```diff
@@ -29,9 +29,9 @@ func TestUnspecified(t *testing.T) {
 	}{
 		{"Valid", "BookFormat", "BOOK_FORMAT_UNSPECIFIED", testutils.Problems{}},
 		{"ValidWithNum", "Ipv6Format", "IPV6_FORMAT_UNSPECIFIED", nil},
+		{"ValidUnknown", "BookFormat", "UNKNOWN", nil},
 		{"InvalidNoPrefix", "BookFormat", "UNSPECIFIED", testutils.Problems{{Suggestion: "BOOK_FORMAT_UNSPECIFIED"}}},
 		{"InvalidWrongSuffix", "BookFormat", "BOOK_FORMAT_UNKNOWN", testutils.Problems{{Suggestion: "BOOK_FORMAT_UNSPECIFIED"}}},
-		{"InvalidJustWrong", "BookFormat", "UNKNOWN", testutils.Problems{{Suggestion: "BOOK_FORMAT_UNSPECIFIED"}}},
```
noahdietz, line 34: Funny that this was the "JustWrong" test case, but is actually an exception in the AIP. Must've fallen out of sync.

apasel422 (approved):

PR #899, opened 2021-11-09
fix: disable all rules instead of core rules only

maxim-tschumak:
This makes the integration test independent of particular rule sets.

For example, it allows creation of rules in other spaces that may generate
violations for the sample API definition.


noahdietz (approved):

noahdietz:
Please update your branch.

PR #900, opened 2021-11-11
refactor: add MethodOption location function

maxim-tschumak:


locations/method_locations.go
```diff
@@ -48,3 +49,9 @@ func MethodOperationInfo(m *desc.MethodDescriptor) *dpb.SourceCodeInfo_Location
 func MethodSignature(m *desc.MethodDescriptor, index int) *dpb.SourceCodeInfo_Location {
 	return pathLocation(m, 4, int(apb.E_MethodSignature.TypeDescriptor().Number()), index) // MethodDescriptor.options == 4
 }
+
+// MethodOption returns the precise location of the method's option with the given name, if any.
+func MethodOption(m *desc.MethodDescriptor, option string) *dpb.SourceCodeInfo_Location {
+	fd := m.GetMethodOptions().ProtoReflect().Descriptor().Fields().ByName(protoreflect.Name(option))
+	return pathLocation(m, 4, int(fd.Number())) // MethodDescriptor.options == 4
+}
```
apasel422, line 57: Instead of taking the option as a string, what do you think of taking just the option integer and having the caller figure out the mapping between name and number?

apasel422: That way, the caller can use the descriptor itself to determine the value, like we do above, e.g.

```go
loc := MethodOption(m, int(apb.E_Http.TypeDescriptor().Number()))
```

locations/method_locations_test.go
```diff
@@ -117,3 +117,33 @@ func TestMethodSignature(t *testing.T) {
 		}
 	}
 }
+
+func TestMethodOption(t *testing.T) {
```
apasel422, line 121: This test is failing.

apasel422 (changes requested):

locations/method_locations_test.go
```diff
@@ -117,3 +117,33 @@ func TestMethodSignature(t *testing.T) {
...
+	for _, test := range []struct {
+		name      string
+		methodIdx int
+		option    string
+		want      []int32
+	}{
+		{"First", 0, "deadline", []int32{5, 20, 43}},
+		{"Second", 1, "deadline", nil},
+	} {
+		loc := MethodOption(f.GetServices()[0].GetMethods()[test.methodIdx], test.option)
```
shwoodard, line 144: imo, we should be using, https://pkg.go.dev/testing#T.Run.

e.g.,

```
func TestFoo(t *testing.T) {
    for _, test  := range []struct {
        name string
        ...
    }{
        {...},
        {...},
    } {
        t.Run(test.name, func (t *testing.T) {
            ...
            if actual != expected {
                t.Errorf("Expected, %s, was %s", expected, actual)
            }
        }
    }
}
```

shwoodard: In turn, `name` will appear in output to stdout when running tests and could optionally be omitted from the `Fatalf` or `Errorf` invocation. Finally, this is the recommended table test method from the golang blog, "[Go 1.7 also introduces a Run method for creating subtests](https://go.dev/blog/subtests)".

locations/method_locations_test.go
```diff
@@ -117,3 +117,33 @@ func TestMethodSignature(t *testing.T) {
...
+	`)
+
+	for _, test := range []struct {
+		name      string
+		methodIdx int
+		option    string
+		want      []int32
+	}{
+		{"First", 0, "deadline", []int32{5, 20, 43}},
+		{"Second", 1, "deadline", nil},
```
shwoodard, line 142: Please make test name more descriptive so that it describes the behavior that you are testing, per sub-test.

shwoodard (changes requested):

locations/method_locations.go
maxim-tschumak: Done

locations/method_locations.go
```diff
@@ -48,3 +48,8 @@ func MethodOperationInfo(m *desc.MethodDescriptor) *dpb.SourceCodeInfo_Location
 func MethodSignature(m *desc.MethodDescriptor, index int) *dpb.SourceCodeInfo_Location {
 	return pathLocation(m, 4, int(apb.E_MethodSignature.TypeDescriptor().Number()), index) // MethodDescriptor.options == 4
```
apasel422, line 49: Could you also change the functions above to use `MethodOptions`? For example:

```suggestion
	return MethodOption(int(apb.E_MethodSignature.TypeDescriptor().Number()), index)
```

locations/method_locations_test.go
```diff
@@ -117,3 +117,36 @@ func TestMethodSignature(t *testing.T) {
 		}
 	}
 }
+
+func TestMethodOption(t *testing.T) {
+	f := parse(t, `
+		import "google/api/client.proto";
```
apasel422, line 123:
```suggestion
```

locations/method_locations_test.go
```diff
@@ -117,3 +117,36 @@ func TestMethodSignature(t *testing.T) {
...
+		service Library {
+		  rpc GetBook(GetBookRequest) returns (Book) {
+		    option deprecated = true;
+		  }
+		  rpc WriteBook(NewBook) returns (Book) {}
+		}
+		message GetBookRequest{}
+		message Book {}
+		message NewBook {}
```
apasel422, line 132:
```suggestion
		  rpc UpdateBook(UpdateBookRequest) returns (Book) {}
		}
		message GetBookRequest{}
		message Book {}
		message UpdateBookRequest {}
```

locations/method_locations_test.go
```diff
@@ -117,3 +117,36 @@ func TestMethodSignature(t *testing.T) {
...
+	`)
+
+	for _, test := range []struct {
+		name        string
+		methodIdx   int
+		fieldNumber int
```
apasel422, line 138: Since the `fieldNumber` is the same for both test cases, no need for it to be a field in this struct.

apasel422 (changes requested):

locations/method_locations.go
```diff
@@ -48,3 +48,8 @@ func MethodOperationInfo(m *desc.MethodDescriptor) *dpb.SourceCodeInfo_Location
...
 }
+
+// MethodOption returns the precise location of the method's option with the given field number, if any.
+func MethodOption(m *desc.MethodDescriptor, fieldNumber int) *dpb.SourceCodeInfo_Location {
```
apasel422, line 53: Could you also change `MethodHTTPRule` and `MethodOperationInfo` above to use `MethodOption`?

apasel422 (approved):

shwoodard (approved):

PR #902, opened 2021-11-12
Allow releases to be triggered manually

apasel422:
It is sometimes useful to be able to create a release without having to wait for the scheduled run of this action.


noahdietz:
Awesome idea. Is there some documentation we should update to say "to manually trigger a release, press this button"? 

Personally, idk how to manually trigger workflows.

noahdietz (approved):
Approving, documentation update is a nit, can be done separately as well.

Issue #911, opened 2021-12-27
Implement check for AIP-193


maxim-tschumak:
google.aip.dev/193

noahdietz:
_Going through old issues._

In it's current form, there is nothing to link in the API proto re:`google.rpc.Status`. Closing for now.

PR #913, opened 2021-12-27
feat: implement a check for AIP-193 (#911)

maxim-tschumak:
Signed-off-by: Maxim Tschumak <maxim.tschumak@gmail.com>

### DIFF
--- a/docs/rules/0193/index.md
+++ b/docs/rules/0193/index.md
@@ -1,0 +1,10 @@
+---
+aip_listing: 193
+permalink: /193/
+redirect_from:
+  - /0193/
+---
+
+# Documentation
+
+{% include linter-aip-listing.md aip=193 %}
--- a/docs/rules/0193/response-status-type.md
+++ b/docs/rules/0193/response-status-type.md
@@ -1,0 +1,59 @@
+---
+rule:
+  aip: 193
+  name: [core, '0193', response-status-type]
+  summary: Services must return a google.rpc.Status message when an API error occurs
+permalink: /193/response-status-type
+redirect_from:
+  - /0193/response-status-type
+---
+
+# Response status type
+
+This rule attempts to enforce that every response message uses the same
+standard type for status objects, as mandated in [AIP-193][].
+
+## Details
+
+This rule looks at each response message type of every service in proto file and
+verifies whether messages with the names like `status`, `error`, or `warning`
+have the correct type (`google.rpc.Status`). Only fields in the top-level
+message are checked.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message GetBookResponse {
+  int32 status = 1;
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+import "google/rpc/status.proto";
+
+message GetBookResponse {
+  google.rpc.Status status = 1;
+}
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the descriptor.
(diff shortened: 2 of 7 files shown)


noahdietz:
Thanks for this but I'm not sure which guidance this rule actually applies to. There is no guidance in AIP-193 that states fields named `status`, `error`, `warning` must be of type `google.rpc.Status` or that an API must just this in their response _message_. This AIP is directed at using `google.rpc.Status` as the error response format

maxim-tschumak:
Hi Noah, this PR is more of a question whether the check can be useful here.

The AIP states that we need to use standardized types for error/warning fields in the response message (`Services must return a google.rpc.Status message when an API error occurs`). The idea of this check is to have a list of top-level field names for output type that suppose to be typed with the standardized message type. While the list is not explicitly specified in the AIP, I think this implementation is likely to catch such violations without producing false-positives.

If you think it's too much interpretation of the AIP, then let's drop this PR. We plan to use the check in our sub-org and I can simply move the implementation to our custom package. I've just thought it may be useful for a broader audience.

noahdietz:
>  The idea of this check is to have a list of top-level field names for output type that suppose to be typed with the standardized message type.

To make sure I understand (and please tell me if I don't): This lints response messages and ensures that "if a field exists with one of the targeted names (e.g. `status`), it must be type `google.rpc.Status`"?

This ^ is an over-interpretation of the AIP. The AIP is not concerned with a field in the response message, but rather the proto/schema used to construct the error returned by server, which must be `google.rpc.Status`. For gRPC, the gRPC Status is based on this proto already. For HTTP 1.1./JSON APIs they must use this proto's JSON schema to construct the response payload (see https://cloud.google.com/apis/design/errors also).

There are many cases where a field of type `google.rpc.Status` is recommended (https://google.aip.dev/search?q=status), but those are generally for async jobs that wrap some other response (e.g. google.longrunning.Operation).

If I'm not tracking, lmk and maybe add an example :)

maxim-tschumak:
> To make sure I understand (and please tell me if I don't): This lints response messages and ensures that "if a field exists with one of the targeted names (e.g. status), it must be type google.rpc.Status"?

Yes, this is what the intent of the check. "Use standard message type for errors". For example, this would generate a violation if API author specifies a top-level `error` field in the response message of type `string` (and not the standardized `google.rpc.Status`). I've seen a couple of such use-cases.

I still think it's a valid info-level check, but yes, it is maybe a over-interpretation. Feel free to drop this PR. I can move the implementation to our custom rule package.

noahdietz:
@shwoodard WDYT about the above?

noahdietz:
> For example, this would generate a violation if API author specifies a top-level error field in the response message of type string (and not the standardized google.rpc.Status).

Yeah, I think I agree with this. There are a number of places in the AIPs that recommend having a `google.rpc.Status` field and refer to AIP-193, but there is no mention of what the field _name_ should be. I think we should:
1. Update AIP-193 to put such guidance into words
2. Update the linter to enforce that a field with one of the allowed names must be of type `google.rpc.Status` (what this PR has)
3. Update the linter to enforce that a field of type `google.rpc.Status` is named with an allowed name (inverse check of this PR)
4. Update any AIP recommending use of `google.rpc.Status` as a field (151, 152, 153 afaik) to explicitly refer to the 193 guidance

maxim-tschumak:
Please let me know once the decision has been made and the guideline is updated. I'm happy to implement 3 too.

noahdietz:
I don't think this is currently lintable. Closing for the time being.

PR #937, opened 2022-03-04
feat: lint parent field resource_reference types

### LINKED ISSUE #928: validate type/child_type use on parent fields
Often, API authors will annotate a `parent` field with a `type` reference to the resource that is actually the `child_type`. Using either `type` or `child_type` references is allowed for a `parent` field, but they would have different values. If possible, such references on `parent` fields should be linted to verify that either the `type` is *not* the child resource OR `child_type` is the resource that it should be `i.e. in ListBooksRequest, the child resource is Books as indicated by the message name and as the only repeated message (that is annotated with a google.api.resource) in the response).

noahdietz:
Fixes #928


rules/aip0132/resource_reference_type_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/googleapis/api-linter/rules/internal/testutils"
+)
+
+func TestResourceReferenceType(t *testing.T) {
+	// Set up testing permutations.
+	tests := []struct {
+		testName string
+		TypeName string
+		RefType  string
+		problems testutils.Problems
```
shwoodard, line 29: With the `Message` fields on the `testutils.Problem`, the value used to compare with `Diff(...)` is, `"child_type"` which appears in both `lint.Problem` `Message`s. So the test, can have false results, I think. Please fix if you agree.

rules/aip0132/resource_reference_type.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// List methods should reference the target resource via `child_type` or the
+// parent directly via `type`.
+var resourceReferenceType = &lint.MethodRule{
+	Name: lint.NewRuleName(132, "resource-reference-type"),
+	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		repeated := utils.GetRepeatedMessageFields(m.GetOutputType())
```
shwoodard, line 29: afaict, you are kinda bundling in a check that the OutputType, i.e. the Response Message, has a repeated field that, purportedly, is the repeated resources. However, this is not the end goal of this `MethodRule` afaict, but rather the goal is to check for the correct value of `type` or `child_type` on `resource_reference`.

In turn, we are not really doing a good job of the "bundled in" check, because we don't validate that the repeated field name is the plural of the resource type. In turn, I would remove this line and the check for `len(repeated) > 0`, and rather create a different `rules/aip0132/<name>.go` file with a `*lint.MethodRule` of its own that checks not only that there is *a* repeated field, but also that, that repeated field **is correct**, in terms of name and type, and all the things.

Make sense?

rules/aip0132/resource_reference_type_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}{
+		{"ValidChildType", "library.googleapis.com/Book", "child_type", nil},
+		{"ValidType", "library.googleapis.com/Shelf", "type", nil},
```
shwoodard, line 32: How does the lint rule know that the parent type is Shelf? That string does not appear in the example Message. It makes me wonder if it is actually being validated--the the `type` (wrt `resource_reference`) is, in fact correct, `library.googleapis.com/Shelf`. I *think* that we are just checking that it is not the the `child_type`. In turn, it is not really valid, but it is not, not valid, iiuc.

rules/aip0133/resource_reference_type.go
```diff
@@ -0,0 +1,59 @@
...
+// parent directly via `type`.
+var resourceReferenceType = &lint.MethodRule{
+	Name: lint.NewRuleName(133, "resource-reference-type"),
+	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		p := m.GetInputType().FindFieldByName("parent")
+		return isCreateMethod(m) && p != nil && utils.GetResourceReference(p) != nil
+	},
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		out := m.GetOutputType()
+		if out.GetName() == "Operation" {
```
shwoodard, line 34: Afaict, the `md` docs above for this lint rule do not explain the this also does checking for LROs and also provide example for these cases.

Also, why do we only check LRO response Message semantics for Create methods, e.g. and not List methods or other types, Get, etc?

rules/aip0133/resource_reference_type.go
```diff
@@ -0,0 +1,59 @@
...
+				Location:   locations.FieldResourceReference(parent),
+			}}
+		}
+		if ref.GetChildType() != "" && resource.GetType() != ref.GetChildType() {
+			return []lint.Problem{{
+				Message:    "Create should use a child_type reference to the created resource.",
+				Descriptor: parent,
+				Location:   locations.FieldResourceReference(parent),
+			}}
+		}
```
shwoodard, line 55: Optional: While the `Message`s are different, these conditionals are the same: these could be candidates to be "dryed off".

rules/aip0133/resource_reference_type_test.go
```diff
@@ -0,0 +1,71 @@
...
+		t.Run(test.testName, func(t *testing.T) {
+			file := testutils.ParseProto3Tmpl(t, `
+				import "google/api/resource.proto";
+				import "google/longrunning/operations.proto";
+				service Library {
+					rpc CreateBook(CreateBookRequest) returns ({{ .Response }}) {
+						option (google.longrunning.operation_info) = {
+							response_type: "Book"
+							metadata_type: "Book"
+						};
```
shwoodard, line 50: It seems odd to include `option`, `google.longrunning.operation_info`, for the cases that are not LROs. I would suggest doing one of the following

1. Have two different top level tests with different test case arrays, one per top level test, and different top level test method bodies for Operations and without Operations.
2. Have one top level test case with two different test case arrays, in the same top level test, with different loops and test bodies, within each loop.
3. In the template, use a conditional that omits the `operation_info` iff the `.Response` does not `Equals` `google.longrunning.Operations`

If you opt for 1 or 2, above, you can dry off with a private method, optionally.

rules/aip0233/resource_reference_type.go
```diff
@@ -0,0 +1,63 @@
...
+package aip0233
+
+import (
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// CatchCreate methods should reference the target resource via `child_type` or
```
shwoodard, line 24: Type, "BatchCreate"

rules/aip0233/resource_reference_type.go
```diff
@@ -0,0 +1,63 @@
...
+// the parent directly via `type`.
+var resourceReferenceType = &lint.MethodRule{
+	Name: lint.NewRuleName(233, "resource-reference-type"),
+	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		repeated := utils.GetRepeatedMessageFields(m.GetOutputType())
```
shwoodard, line 29: Same comment about conflating method rules. Create a separate method rule for repeated resource type in the response?

rules/aip0233/resource_reference_type.go
```diff
@@ -0,0 +1,63 @@
...
+		p := m.GetInputType().FindFieldByName("parent")
+		return isBatchCreateMethod(m) && p != nil && utils.GetResourceReference(p) != nil && len(repeated) > 0
+	},
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		msg := m.GetOutputType()
+		if msg.GetName() == "Operation" {
```
shwoodard, line 35: Same comment, not documented.

rules/aip0233/resource_reference_type.go
```diff
@@ -0,0 +1,63 @@
...
+			info := utils.GetOperationInfo(m)
+			msg = utils.FindMessage(m.GetFile(), info.GetResponseType())
+		}
+		repeated := utils.GetRepeatedMessageFields(msg)
+		resMsg := repeated[0].GetMessageType()
```
shwoodard, line 40: Assumes that the first repeated field in the response is the resource, which I think is more reason to lint for this in another rule that is of higher precedence. See other, related, comment(s).

rules/aip0132/resource_reference_type.go
```diff
@@ -0,0 +1,58 @@
...
+		p := m.GetInputType().FindFieldByName("parent")
+		return isListMethod(m) && p != nil && utils.GetResourceReference(p) != nil && len(repeated) > 0
+	},
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		// The first repeated message field must be the paginated resource.
+		repeated := utils.GetRepeatedMessageFields(m.GetOutputType())
+		resource := utils.GetResource(repeated[0].GetMessageType())
```
shwoodard, line 36: Same as above and below: `repeated[0]` relies on the implication that the first repeated field is the resource. Do we lint for this, explicitly?

rules/aip0233/resource_reference_type_test.go
```diff
@@ -0,0 +1,74 @@
...
+			file := testutils.ParseProto3Tmpl(t, `
+				import "google/api/resource.proto";
+				import "google/longrunning/operations.proto";
+				service Library {
+					rpc BatchCreateBooks(BatchCreateBooksRequest) returns ({{ .Response }}) {
+						option (google.longrunning.operation_info) = {
+							response_type: "BatchCreateBooksResponse"
+							metadata_type: "BatchCreateBooksResponse"
+						};
+					}
```
shwoodard, line 51: Same comment regarding `operation_info` in the body of tests that are not LROs as above.

shwoodard (changes requested):
@noahdietz I think there're a few things here to consider and possibly address. Cheers!

docs/rules/0233/resource-reference-type.md
```diff
@@ -0,0 +1,66 @@
...
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message BatchCreateBooksRequest {
+  // `child_type` should be used instead of `type` when referring to the
+  // created resource on a parent field.
+  string parent = 1 [(google.api.resource_reference).type = "library.googleapis.com/Book"];
+  Book book = 2;
```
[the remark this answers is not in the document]
noahdietz: Yep! Thanks, copy-paste gone wrong.

docs/rules/0233/resource-reference-type.md
```diff
@@ -0,0 +1,66 @@
...
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+message BatchCreateBooksRequest {
+  string parent = 1 [(google.api.resource_reference).child_type = "library.googleapis.com/Book"];
+  Book book = 2;
```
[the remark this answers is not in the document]
noahdietz: Done.

docs/rules/0233/resource-reference-type.md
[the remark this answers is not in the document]
noahdietz: Good call. Done.

docs/rules/0233/resource-reference-type.md
```diff
@@ -0,0 +1,66 @@
...
+
+If you need to violate this rule, use a leading comment above the field.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+message BatchCreateBooksRequest {
+  // (-- api-linter: core::0233::resource-reference-type=disabled
+  //     aip.dev/not-precedent: We need to do this because reasons. --)
+  string parent = 1 [(google.api.resource_reference).type = "library.googleapis.com/Book"];
+  Book book = 2;
```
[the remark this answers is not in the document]
noahdietz: Done.

rules/aip0132/resource_reference_type.go
```diff
@@ -0,0 +1,58 @@
...
+
+		parent := m.GetInputType().FindFieldByName("parent")
+		ref := utils.GetResourceReference(parent)
+
+		if resource.GetType() == ref.GetType() {
+			return []lint.Problem{{
+				Message:    "List should use a child_type reference to the paginated resource, not a type reference.",
```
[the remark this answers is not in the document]
noahdietz: Good call. I'm going to use backticks instead of double quotes as that seems to be the pattern in other Problem.Message strings

rules/aip0132/resource_reference_type.go
```diff
@@ -0,0 +1,58 @@
...
+				Descriptor: parent,
+				Location:   locations.FieldResourceReference(parent),
+			}}
+		}
+		if ref.GetChildType() != "" && resource.GetType() != ref.GetChildType() {
+			return []lint.Problem{{
+				Message:    "List should use a child_type reference to the paginated resource.",
```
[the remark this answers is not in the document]
noahdietz: Done.

rules/aip0132/resource_reference_type.go
noahdietz: > kinda bundling in a check

Yes and no. My intention here is to not _lint_ the response message for the necessary fields, but just ignore those that aren't in scope of this rule. The `resource-reference-type` rule shouldn't emit a Problem when the `repeated Resource` in the response doesn't actually have a `google.api.resource` annotation for some reason or if it is a primitive (which isn't technically allowed, but I've seen it). I'd rather keep this check for that reason - so that this rule doesn't warn on things that are out of its scope.

> we are not really doing a good job of the "bundled in" check...rather create a different rules/aip0132/<name>.go file with a *lint.MethodRule of its own that checks not only that there is a repeated field, but also that, that repeated field is correct, in terms of name and type, and all the things.

Yeah, totally, we should add these rules.

rules/aip0132/resource_reference_type_test.go
noahdietz: You're right, this is a naive check, it doesn't go as far as comparing each `pattern` in the child type to those in the parent type and ensuring things line up. Do you think that belongs in this check or in a different one? It only applies to the direct `type` reference.

rules/aip0233/resource_reference_type.go
noahdietz: See previous comment :)

rules/aip0233/resource_reference_type.go
noahdietz: For AIP-233 it is already done it seems: https://linter.aip.dev/233/response-resource-field

If anything the `OnlyIf` should be expanded to only lint if the first repeated response field is annotated with a resource. We only want rules to warn on protos that fulfill all preconditions.

rules/aip0233/resource_reference_type.go
noahdietz: Good catch, thanks. Done.

rules/aip0233/resource_reference_type_test.go
noahdietz: Done!

rules/aip0133/resource_reference_type_test.go
noahdietz: Went with option 1 :)

rules/aip0132/resource_reference_type_test.go
noahdietz: Yep you are right!

rules/aip0132/resource_reference_type_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestResourceReferenceType(t *testing.T) {
+	// Set up testing permutations.
+	tests := []struct {
+		testName string
+		TypeName string
+		RefType  string
+		problems testutils.Problems
+	}{
+		{"ValidChildType", "library.googleapis.com/Book", "child_type", nil},
```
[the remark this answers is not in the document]
noahdietz: Ack!

rules/aip0132/resource_reference_type_test.go
noahdietz: I wonder if we should even allow use of `type` anymore...`child_type` is the strictly better reference type for these RPCs as it is "forward compatible" without needing to change anything. If a new parent pattern is added to a resource, and the parent fields are using a direct `type` reference, they would need to be updated to child_type in order to enable use of the new parent pattern for that resource e.g. having a resource parented by just project, and then adding organization as a parent pattern.

WDYT?

rules/aip0233/resource_reference_type.go
shwoodard: SGTM, yes, let's change the `OnlyIf` as you suggested.

rules/aip0132/resource_reference_type_test.go
shwoodard: SGTM. I wonder how many references there are that use `type. ?

rules/aip0233/resource_reference_type.go
noahdietz: I expanded the `OnlyIf` to ensure that there is a resource defined in the first repeated message in the response.

rules/aip0233/resource_reference_type.go
```diff
@@ -0,0 +1,77 @@
...
+		out := m.GetOutputType()
+		if out.GetName() == "Operation" {
+			info := utils.GetOperationInfo(m)
+			out = utils.FindMessage(m.GetFile(), info.GetResponseType())
+		}
+
+		// First repeated message field must be annotated with google.api.resource.
+		repeated := utils.GetRepeatedMessageFields(out)
+		var resource *annotations.ResourceDescriptor
+		if len(repeated) > 0 {
```
shwoodard, line 39: IIUC, if `len(repeated)` is zero then resource will alway be `nil` and the condition at the end of this statement will always be false. Hence, let's change this to,

```
if len(repeated) == 0 {
  return false
}
```

then, whole block can be outdented and `resource` can be declared with `:=` rather than `var`, etc.

rules/aip0233/resource_reference_type_test.go
```diff
@@ -23,15 +23,17 @@ import (
...
 	// Set up testing permutations.
 	tests := []struct {
-		testName string
-		TypeName string
-		RefType  string
-		problems testutils.Problems
+		testName        string
+		TypeName        string
+		RefType         string
+		RepeatedMessage string
```
shwoodard, line 29: Let's rename this, "RepeatedMessageType", please.

rules/aip0233/resource_reference_type.go
noahdietz: Done.

rules/aip0233/resource_reference_type_test.go
noahdietz: Done.

shwoodard (approved):

PR #938, opened 2022-03-04
fix(AIP-123): support name_field

### LINKED ISSUE #927: core::0123::resource-name-field should accept a resource declaring a name_field
If a message that defines a `google.api.resource` does not have a `string name` field, but declares a `name_field` in the `google.api.resource` definition that points to a `string` field in that message, this rule should not complain. It is not uncommon to annotate resources on an older API that preceded the resource annotation OR to have different API style that prefers a different name for the resource name field. 

Proto for `name_field`: https://github.com/googleapis/googleapis/blob/ada32e9805a83fdf0ec269c23ff7f5d568c34938/google/api/resource.proto#L159-L161

noahdietz:
Fixes #927


shwoodard:
I'm a little unsure about this one: reading [AIP#123](https://google.aip.dev/123), there is no mention of the `name_field` option for resources. Should we add this documentation to the AIP before linting for it? If there is precedent for adding undocumented options or if I am missing where the documentation resides, then we can disregard this concern.

shwoodard:
> I'm a little unsure about this one: reading [AIP#123](https://google.aip.dev/123), there is no mention of the `name_field` option for resources. Should we add this documentation to the AIP before linting for it? If there is precedent for adding undocumented options or if I am missing where the documentation resides, then we can disregard this concern.

Ok, I think I may get the disconnect and how the docs are intertwined: the documentation for this options should be [here](https://github.com/googleapis/api-linter/blob/main/docs/rules/0123/resource-name-field.md), iiuc.

shwoodard (changes requested):
I think we should document the linting for this `option` field, `name_field`, [here](https://github.com/googleapis/api-linter/blob/main/docs/rules/0123/resource-name-field.md), as part of this PR, based on looking at similar ones. Is that correct?

noahdietz:
I think this should be added as a "break glass" option for "legacy" APIs (i.e. adding google.api.annotation to an existing API) or those that have a PA-specific pattern (maybe corp likes their resource name fields to be "resource") and that we shouldn't encourage it in the linter docs, but rather as a `Note`/"may use name_field if it absolutely has to be done" in AIP-123 . WDYT?

shwoodard:
> I think this should be added as a "break glass" option for "legacy" APIs (i.e. adding google.api.annotation to an existing API) or those that have a PA-specific pattern (maybe corp likes their resource name fields to be "resource") and that we shouldn't encourage it in the linter docs, but rather as a `Note`/"may use name_field if it absolutely has to be done" in [AIP-123](https://google.aip.dev/123) . WDYT?

iiuc, you don't want it in the docs, because you don't want to encourage it. +1

shwoodard (approved):

noahdietz:
> > I think this should be added as a "break glass" option for "legacy" APIs (i.e. adding google.api.annotation to an existing API) or those that have a PA-specific pattern (maybe corp likes their resource name fields to be "resource") and that we shouldn't encourage it in the linter docs, but rather as a `Note`/"may use name_field if it absolutely has to be done" in [AIP-123](https://google.aip.dev/123) . WDYT?
> 
> iiuc, you don't want it in the docs, because you don't want to encourage it. +1

Yep, exactly. Thanks :)

PR #948, opened 2022-03-16
fix(AIP-191): valid characters in filename

### LINKED ISSUE #947: AIP-191 File names must have valid characters
AIP-191 states [File names must use snake_case](https://google.aip.dev/191) but `api-linter` does not check for invalid characters. 

Valid Character Ranges include:
`a-z`
`A-Z`
`0-9`
`_`

Anything else could be considered invalid. 

As an example, a proto filename should be `test_service.proto` not `test.service.proto`

Question: Should `A-Z` be considered valid?

parthea:
Fixes #947

### DIFF
--- a/rules/aip0191/aip0191.go
+++ b/rules/aip0191/aip0191.go
@@ -45,4 +45,4 @@ func hasPackage(f *desc.FileDescriptor) bool {
 }
 
 var versionRegexp = regexp.MustCompile(`^v[0-9]+(p[0-9]+)?((alpha|beta)[0-9]*)?$`)
-var validCharacterRegexp = regexp.MustCompile(`^[a-zA-Z0-9\\_\\/]*$`)
+var validCharacterRegexp = regexp.MustCompile(`^[a-z0-9\\_\\/]*$`)
--- a/rules/aip0191/aip0191.go
+++ b/rules/aip0191/aip0191.go
@@ -44,5 +44,5 @@ func hasPackage(f *desc.FileDescriptor) bool {
 	return f.GetPackage() != ""
 }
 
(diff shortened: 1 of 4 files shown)


rules/aip0191/aip0191.go
```diff
@@ -45,3 +45,4 @@ func hasPackage(f *desc.FileDescriptor) bool {
 }
 
 var versionRegexp = regexp.MustCompile("^v[0-9]+(p[0-9]+)?((alpha|beta)[0-9]*)?$")
+var validCharacterRegexp = regexp.MustCompile("^[a-zA-Z 0-9\\_\\/]*$")
```
noahdietz, line 48:
```suggestion
var validCharacterRegexp = regexp.MustCompile(`^[a-zA-Z 0-9\\_\\/]*$`)
```

I realize that the expression directly above uses `"` but I guess the golang linter wasn't run on that haha

noahdietz, line 48: There is a whitespace in between `A-Z` and `0-9`, is that necessary?

noahdietz (commented):
Looks great

rules/aip0191/aip0191.go
parthea: Great catch! I also had a doubt about `A-Z`. Do we allow uppercase?

parthea:
The status checks are stuck. I'm going to try closing and re-opening the PR.

rules/aip0191/aip0191.go
noahdietz: Probably not, so lets remove it for now, and if it ends up being "OK" we can add it back (always easier to make it less restrictive later).

noahdietz (approved):
LGTM pending removal of `A-Z` from the "valid regex"

noahdietz:
closing/reopening to trigger Actions

PR #955, opened 2022-04-15
chore: remove renovate grouping

noahdietz:
Changes from RenovateBot that update this repos dependencies will no longer be grouped in a single Pull Request (like it is [today](https://github.com/googleapis/api-linter/pull/926)). Instead, they will come in as individual Pull Requests.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,7 +1,6 @@
 {
   extends: ["config:base"],
   commitMessagePrefix: "chore: ",
-  groupName: "all",
   packageRules: [
     {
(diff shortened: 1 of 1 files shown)


GregFurmanek:
Out of curiosity, what are the implications of this change?

noahdietz:
Changes from RenovateBot that update this repos dependencies will no longer be grouped in a single Pull Request (like it is [today](https://github.com/googleapis/api-linter/pull/926)). Instead, they will come in as individual Pull Requests.

Adding this to the description.

GregFurmanek (approved):

Issue #82, opened 2019-06-15
Error levels


lukesneeringer:
As best as I can tell, `lint.Problem` lacks any concept of an error level.
Most linters support both warnings and errors, and we presumably should also (e.g. warn for "shoulds", error for "musts").

Most linters also have an "error on warnings" setting.

mingzhi:
We are using `Category` in `Problem` as a proxy for error levels, e.g., a category can be `error`, `warning`, or `commit_blocking`. We let users to define the category for each rule in their rule configuration. But we can provide a default for each rule.

lukesneeringer:
Got it, that makes sense. I just did not realize what `Category` meant.
Are the values strings?

mingzhi:
Yes, they are strings -- we don't want to pre-define categories -- it is up to users to define them in their configuration, so they can develop/use downstream tools to target their categories. For example, some teams may want to use presubmit check to target the violation of some rules of their defined category (e.g., category: "my-team-want-to-enforce-these-rules").

lukesneeringer:
Okay, I am happy with that.

PR #966, opened 2022-04-29
feat: introduce problem severity levels

maxim-tschumak:


noahdietz:
Ok so I noticed that we have a `Category string` field and it was added for a similar reason (see https://github.com/googleapis/api-linter/issues/82). What I don't understand is why there aren't any pre-defined categories. There is a reasoning stated in that issue is as follows:
> it is up to users to define them in their configuration, so they can develop/use downstream tools to target their categories

So I am wondering if we should do one of the following:
1. Go with this PR to add `Severity` as a purpose-built field
2. Introduce some pre-defined `Category` values that emulate `Severity` _publicly_
3. Introduce some pre-defined `Category` values that emulate `Severity` _internally_ (which our analyzers would key off of when reporting to critique)

WDYT? I can see the benefit of adding a `Severity` "enum", but I'm also worried about bloat considering it is slightly related to `Category`.

@GregFurmanek WDYT?

noahdietz:
> we would enable teams to decide how they want API owners to react to the problems in their (e.g. project) scope.

I don't think we want to give individual API teams (not PAs, but individual APIs within a PA) the ability to downgrade a linter warning. If they are exempt they can disable the rule in the proto with a comment or in their config. But the rule author should decide how serious a linter warning is IMO. Also, I'd be worried about config bloat/complexity. These configs should be dead simple to look at and interpret. 

We should keep a severity style attribute on the lint.Problem.

noahdietz:
> I agree with Noah. Linter is an automated test tool of sorts. Tests are pretty dimple in their function and answer one question: does the code pass specific scenario. The concept of severity is more useful in error reporting. Not so much in testing.

I might not have been clear above, but, I don't disagree with adding a severity-style field to lint.Problem that we can use in Tricorder to indicate if the finding is Blocking or not. I just believe it should be a property of lint.Problem, and not in the lint.Config, because the author of the rule should decide if the finding is blocking or not.

vinniespina (approved):

vinniespina (approved):

vinniespina (commented):
security a must

quirogas:
I'm closing this PR since it hasn't been addressed recently and we are preparing to release api-linter v2 (#1529). If you still wish to introduce this feature, please open a new issue. Thank you!

PR #969, opened 2022-05-04
(feat aip-203): add linter for field_behavior presence

ScottSuarez:
WIP blocked by https://github.com/googleapis/api-linter/pull/966/files for warning level additions
AIP 203 uses /should/ quantifier so in alignment with strong recommendation we desire to add a warning linter.

We were originally looking to add this to ONLY gcp however the original document for AIP-203 already defines this requirement strongly. We are open to making this gcp only if desired.

### DIFF
--- a/rules/aip0203/aip0203.go
+++ b/rules/aip0203/aip0203.go
@@ -28,6 +28,7 @@ import (
 func AddRules(r lint.RuleRegistry) error {
 	return r.Register(
 		203,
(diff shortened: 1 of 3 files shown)


toumorokoshi:
thanks for the change! We'll be closing this in favor of #1149 which has a little bit more of the recent guidance.

ScottSuarez:
cool  ! :)

PR #973, opened 2022-05-17
fix(AIP-132): lint OnlyIf the response message is a resource

noahdietz:
[yaqs/6147033380028416](http://yaqs/6147033380028416) is an example of an API that attempted to annotate a List `parent` field with a resource_reference to a message that isn't a resource. 

The resource_reference type should only be linted if the response message is a resource.


GregFurmanek:
Should this code include test scenario that would originally fail and this change would make it pass? #TDD

noahdietz:
> Should this code include test scenario that would originally fail and this change would make it pass? #TDD

Done! PTAL

GregFurmanek (approved):

PR #980, opened 2022-06-08
fix: handle unresolvable lro response types AIP-133 and AIP-233

### LINKED ISSUE #978: AIP-233: resource-reference-type OnlyIf hits nil pointer if response_type is unresolved
There is a chance that the variable `out` [here](https://github.com/googleapis/api-linter/blob/8873ed7a4ca67f09dc1edac7e0907d2f6f9b15a6/rules/aip0233/resource_reference_type.go#L32) is `nil`, when the type that was searched for is unresolved. Following use of the `out` variable to look up the repeated field(s) via [GetRepeatedMessageFields](https://github.com/googleapis/api-linter/blob/8873ed7a4ca67f09dc1edac7e0907d2f6f9b15a6/rules/internal/utils/find.go#L111) panics [here](https://github.com/googleapis/api-linter/blob/8873ed7a4ca67f09dc1edac7e0907d2f6f9b15a6/rules/internal/utils/find.go#L114).

Both spots should handle `nil` values better.

noahdietz:
Fixes #978 for both AIP-233 and AIP-133.

Also adds some helpers for getting the ResponseType and MetadataType from the OperationInfo.

Unfortunately, the `protoreflect` library used in this project does not handle the `nil` receiver case on proto descriptors like protobuf-go does (i.e. if a descriptor being used as a receiver is `nil`, the method will return the default unset value for the field being accessed).

### DIFF
--- a/rules/aip0133/resource_reference_type.go
+++ b/rules/aip0133/resource_reference_type.go
@@ -26,14 +26,20 @@ import (
 var resourceReferenceType = &lint.MethodRule{
 	Name: lint.NewRuleName(133, "resource-reference-type"),
 	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		out := m.GetOutputType()
+		if out.GetName() == "Operation" {
+			out = utils.GetResponseType(m)
+		}
+
+		// Unresolvable response_type for an Operation results in nil here.
+		resource := utils.GetResource(out)
 		p := m.GetInputType().FindFieldByName("parent")
-		return isCreateMethod(m) && p != nil && utils.GetResourceReference(p) != nil
+		return isCreateMethod(m) && p != nil && utils.GetResourceReference(p) != nil && resource != nil
 	},
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
 		out := m.GetOutputType()
 		if out.GetName() == "Operation" {
-			info := utils.GetOperationInfo(m)
-			out = utils.FindMessage(m.GetFile(), info.GetResponseType())
+			out = utils.GetResponseType(m)
 		}
 		resource := utils.GetResource(out)
 		parent := m.GetInputType().FindFieldByName("parent")
--- a/rules/aip0133/resource_reference_type.go
+++ b/rules/aip0133/resource_reference_type.go
@@ -26,22 +26,24 @@ import (
 var resourceReferenceType = &lint.MethodRule{
 	Name: lint.NewRuleName(133, "resource-reference-type"),
 	OnlyIf: func(m *desc.MethodDescriptor) bool {
-		out := m.GetOutputType()
-		if out.GetName() == "Operation" {
-			out = utils.GetResponseType(m)
+		// Return type of the RPC.
+		ot := m.GetOutputType()
+		if ot.GetName() == "Operation" {
+			ot = utils.GetResponseType(m)
 		}
 
 		// Unresolvable response_type for an Operation results in nil here.
-		resource := utils.GetResource(out)
+		resource := utils.GetResource(ot)
 		p := m.GetInputType().FindFieldByName("parent")
 		return isCreateMethod(m) && p != nil && utils.GetResourceReference(p) != nil && resource != nil
 	},
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
-		out := m.GetOutputType()
-		if out.GetName() == "Operation" {
-			out = utils.GetResponseType(m)
+		// Return type of the RPC.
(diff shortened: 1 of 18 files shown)


rules/aip0133/resource_reference_type.go
```diff
@@ -26,14 +26,20 @@ import (
 var resourceReferenceType = &lint.MethodRule{
 	Name: lint.NewRuleName(133, "resource-reference-type"),
 	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		out := m.GetOutputType()
```
shwoodard, line 29: nit: maybe a different variable name here? "out" makes me thing of i/o. This is the return type of an rpc? Ordinarily, I would say "outputType" or "returnMessageType", but Go likes short variables, so I'm tempted to just suggest, "ot". :/

noahdietz: Sure, done!

rules/internal/utils/extension_test.go
```diff
@@ -124,6 +124,84 @@ func TestGetOperationInfoNone(t *testing.T) {
 	}
 }
 
+func TestGetOperationInfoResponseType(t *testing.T) {
+	// Set up testing permutations.
+	tests := []struct {
+		testName     string
+		ResponseType string
+		invalid      bool
```
shwoodard, line 132: super-nit: Commonly preferred to name the variable after the positive rather than the negation, i.e. valid?

rules/internal/utils/extension_test.go
```diff
@@ -124,6 +124,84 @@ func TestGetOperationInfoNone(t *testing.T) {
...
+			if (typ == nil) != test.invalid {
+				t.Fatalf("Expected invalid(%v) response_type message", test.invalid)
+			}
+			if test.invalid {
+				return
+			}
+			if got, want := typ.GetName(), test.ResponseType; got != want {
+				t.Errorf("Response type - got %q, want %q.", got, want)
+			}
+		})
```
shwoodard, line 162: nit: A little vertical whitespace, between blocks, might be nice here.

rules/internal/utils/extension_test.go
```diff
@@ -124,6 +124,84 @@ func TestGetOperationInfoNone(t *testing.T) {
...
+							response_type: "WriteBookResponse"
+							metadata_type: "{{ .MetadataType }}"
+						};
+					}
+				}
+				message WriteBookRequest {}
+				message WriteBookMetadata {}
+			`, test)
+			typ := GetMetadataType(fd.GetServices()[0].GetMethods()[0])
+			if (typ == nil) != test.invalid {
```
shwoodard, line 192: Can you assign `typ == nil` to its own (descriptive) variable name above? I find this (bool expression) != bool hard to read.

rules/internal/utils/extension_test.go
```diff
@@ -124,6 +124,84 @@ func TestGetOperationInfoNone(t *testing.T) {
...
+							response_type: "{{ .ResponseType }}"
+							metadata_type: "WriteBookMetadata"
+						};
+					}
+				}
+				message WriteBookRequest {}
+				message WriteBookResponse {}
+			`, test)
+			typ := GetResponseType(fd.GetServices()[0].GetMethods()[0])
+			if (typ == nil) != test.invalid {
```
shwoodard, line 153: Same as below, imho, `typ == nil` should be given a var name, like,

```
if typeMissing := typ == nii; typeMissing == test.valid {
...
}

```

rules/internal/utils/extension_test.go
noahdietz: Done.

rules/internal/utils/extension_test.go
noahdietz: Done

rules/internal/utils/extension_test.go
noahdietz: Done

rules/internal/utils/extension_test.go
```diff
@@ -124,6 +124,84 @@ func TestGetOperationInfoNone(t *testing.T) {
...
+		})
+	}
+}
+
+func TestGetOperationInfoMetadataType(t *testing.T) {
+	// Set up testing permutations.
+	tests := []struct {
+		testName     string
+		MetadataType string
+		invalid      bool
```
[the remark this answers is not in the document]
noahdietz: Done.

rules/internal/utils/extension_test.go
noahdietz: Done.

shwoodard (approved):

PR #985, opened 2022-06-22
fix(AIP-215): allow versioned subpackages

### LINKED ISSUE #982: AIP-215: Relax package version component suffix if there is a version in the package
Often APIs will create subdirectories for types, like log messages, within a versioned proto package e.g. `google.cloud.foo.v1.logging`. This should be sufficient for versioning requirements and the linter should not complain if there is a version component preceding the suffix, at least.

noahdietz:
Relaxes proto package version component check, allowing the version component to exist above the leaf segment.

Fixes #982.


rules/aip0215/versioned_packages.go
```diff
@@ -33,14 +33,16 @@ var versionedPackages = &lint.FileRule{
...
+		if len(segments) == 1 && segments[0] == "" {
 			return false
 		}
 
-		// Exempt anything ending in .type, or .v1master, .v2master, .master, etc.
-		if last := p[len(p)-1]; last == "type" || strings.HasSuffix(last, "master") || strings.HasSuffix(last, "main") {
-			return false
+		// Exempt anything containing .type, or .v1master, .v2master, .master, etc.
+		for _, segment := range segments {
+			if segment == "type" || strings.HasSuffix(segment, "master") || strings.HasSuffix(segment, "main") {
```
GregFurmanek, line 43: This seems to significantly relax the rule. 
The original rule excluded checking when the last element was main or master. 
In your new schema any segment with suffix master will cause for the rule to be ignored.

shwoodard: Based on the PR description, I believe this was the intent. Am I incorrect?

noahdietz: This is indeed the intent. `master` and `main` are the "template" words used by Google's proto versioning tools. They could technically appear at any point in the proto package, just like a version could e.g. `google.foo.bar.baz.v1main.biz.buzz` or `google.foo.master.bar.biz.buz.baz`.

PR #989, opened 2022-07-07
fix: correct instances of camel-case-uri to camel-case-uris

nikklassen:


GregFurmanek (commented):
There appear to be a number of instances of this mismatch, including in tests. Perhaps we should correct and test all of them? tagging: @noahdietz

noahdietz:
@nikklassen would you mind fixing the other broken documentation examples in this same PR? You can say "no" :)

lint/rule_urls.go
```diff
@@ -8,13 +8,14 @@ import "strings"
 // At Google, we inject additional rule URL mappings into this list.
 // Example: google_rule_url_mappings.go
 // package lint
-// func init() {
-//   ruleURLMappings = append(ruleURLMappings, internalRuleURLMapping)
-// }
 //
```
GregFurmanek, line 11: Is this change intentional?

nikklassen: I reverted the change, it's due to go 1.19's `gofmt`

GregFurmanek (approved):

PR #995, opened 2022-08-05
feat(AIP-122): resource patterns must start with lowercase letter

### LINKED ISSUE #994: Linter crashes with index out of bounds when there is a slash at the front of a resource pattern
Linter crashes with index out of bounds when there is a slash at the front of a resource pattern. The following proto will crash the linter because the pattern starts with `/` and the linter splits on `/` and then looks at the string at index 0.

```proto
message Book {
    type: "library.googleapis.com/Book"
    // Collection identifiers must begin with a lower-cased letter.
    pattern: "/publishers/{publisher}/Books/{book}"
  };
  string name = 1;
}
```

is bad because [AIP-122](https://google.aip.dev/122) says that collection identifiers must start with a lower-cased letter, even though it should say, "lowercase". :P

shwoodard:
Resource patterns must start with lowercase letters per [AIP-122](https://google.aip.dev/122).

Fixes #994

### DIFF
--- a/docs/rules/0122/resource-collection-identifiers.md
+++ b/docs/rules/0122/resource-collection-identifiers.md
@@ -18,7 +18,7 @@ in [AIP-122][].
 
 This rule scans messages with a `google.api.resource` annotation, and validates
 the format of `pattern` collection identifiers, specifically that they are in
-lowerCamelCase form and that it does not start with a slash.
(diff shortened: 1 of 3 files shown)


noahdietz:
I guess the Go linting wants you to run goimports on the changed directory?

noahdietz (approved):

PR #999, opened 2022-08-18
feat: add GitHub error output format option

achew22:
In this mode, the linter runs and prints out to stdout following the
GitHub "::error" annotation to attach an error to a particular file on a
particular line.

GitHub's documentation on the matter.
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

### DIFF
--- a/cmd/api-linter/github_actions.go
+++ b/cmd/api-linter/github_actions.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,8 +22,8 @@ import (
 	"github.com/googleapis/api-linter/lint"
 )
 
-// printGithubActions returns lint errors in GitHub actions format.
-func printGithubActions(responses []lint.Response) ([]byte, error) {
+// formatGitHubActionOutput returns lint errors in GitHub actions format.
+func formatGitHubActionOutput(responses []lint.Response) []byte {
 	var buf bytes.Buffer
 	for _, response := range responses {
 		for _, problem := range response.Problems {
@@ -39,6 +39,9 @@ func printGithubActions(responses []lint.Response) ([]byte, error) {
 				fmt.Fprintf(&buf, " endColumn=%d", problem.Location.Span[3])
 			}
 
+			// GitHub uses :: as control characters (which are also used to delimit
+			// linter rules. In order to prevent confusion, replace the double colon
+			// with two Armenian full stops which are indistinguishable to my eye.
 			runeThatLooksLikeTwoColonsButIsActuallyTwoArmenianFullStops := "։։"
 			title := strings.ReplaceAll(string(problem.RuleID), "::", runeThatLooksLikeTwoColonsButIsActuallyTwoArmenianFullStops)
 			message := strings.ReplaceAll(problem.Message, "\n", "\\n")
@@ -46,5 +49,5 @@ func printGithubActions(responses []lint.Response) ([]byte, error) {
 		}
 	}
 
-	return buf.Bytes(), nil
+	return buf.Bytes()
 }
--- a/cmd/api-linter/github_actions.go
+++ b/cmd/api-linter/github_actions.go
@@ -1,0 +1,50 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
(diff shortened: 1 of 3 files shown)


cmd/api-linter/github_actions.go
```diff
@@ -0,0 +1,50 @@
...
+
+			fmt.Fprintf(&buf, "::error file=%s", response.FilePath)
+			if problem.Location != nil {
+				fmt.Fprintf(&buf, " line=%d", problem.Location.Span[0])
+				fmt.Fprintf(&buf, " col=%d", problem.Location.Span[1])
+				fmt.Fprintf(&buf, " endLine=%d", problem.Location.Span[2])
+				fmt.Fprintf(&buf, " endColumn=%d", problem.Location.Span[3])
+			}
+
+			runeThatLooksLikeTwoColonsButIsActuallyTwoArmenianFullStops := "։։"
```
noahdietz, line 42: haha I got a kick out of this one. Can you add a comment to this describing why this is necessary (b.c GH uses the `::` as a delimiter already I'm guessing) and perhaps rename this variable to "ruleDelimiter" instead?

cmd/api-linter/github_actions.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/googleapis/api-linter/lint"
+)
+
+// printGithubActions returns lint errors in GitHub actions format.
+func printGithubActions(responses []lint.Response) ([]byte, error) {
```
noahdietz, line 26: I don't see any error modes - do we need to return an error?

noahdietz, line 26: Nit: s/printGithubActions/formatGitHubActionOutput/

noahdietz: If that name isn't accepted, please at least capitalize "GitHub" accurately.

cmd/api-linter/github_actions_test.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2019 Google LLC
```
noahdietz, line 1: 2022

cmd/api-linter/github_actions.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Google LLC
```
noahdietz, line 1: 2022

noahdietz (commented):
This is really cool, thank you @achew22 ! Also, hello in another repo 🙃

noahdietz:
Ignore the conventional commit check, I fixed your PR title already and subsequent commits means the title will be the squashed message.

However looks like the `lint` job wants your to run `goimports` on the added code.

cmd/api-linter/github_actions.go
achew22: Done

achew22: I was trying to match the interface of the calling function. You're right, that was silly 😁

achew22:
Fixed the lint error with https://github.com/googleapis/api-linter/pull/1000. Sorry to grab such a prominent PR number, but congratulations on hitting a milestone!

noahdietz (approved):

noahdietz:
@achew22 do you have an example of how you are running this in GH Actions? Might be nice to have in our docs, short of developing an actual Action for it (but who has time for fun things like that?)

achew22:
@noahdietz sorry to be slow. I intended to do this weeks ago, here is what we have for our GHA config


```yaml
name: proto-lint
on:
  push:
    branches:
      - main
  pull_request:
jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: bufbuild/buf-setup-action@v1
      - run: make aip-lint-action
      - uses: bufbuild/buf-breaking-action@v1
        with:
          # The 'main' branch of the GitHub repository that defines the module.
          against: 'https://github.com/our/repo.git#branch=main'
```

Then we have a `Makefile`

```Makefile
current_dir = $(shell pwd)

.PHONY aip-lint-action:
aip-lint-action:
	buf export . --output=/tmp/buf_export && \
	cd /tmp/buf_export && \
	go run -mod=mod github.com/googleapis/api-linter/cmd/api-linter@main \
		--config=$(current_dir)/aip.lint.yaml \
		--output-format=github \
		$$(find . | grep \.proto$$)
```

We use `buf export` just to get a copy of the proto "stdlib" like

```proto
import "google/api/annotations.proto";
import "google/protobuf/timestamp.proto";
```

loaded into the appropriate directory structure. One of the "how I wish aip-linter did this" for me things would be to ship with a copy of those ~30 protos pre-compiled into aip-linter I could delete all the `buf` stuff and it would make me ever so happy.

```
google/api/
  annotations.proto
  client.proto
  field_behavior.proto
  http.proto
  httpbody.proto
  resource.proto
  visibility.proto

google/longrunning/
  operations.proto

google/rpc/
  code.proto
  error_details.proto
  status.proto

google/type/
  calendar_period.proto
  color.proto
  date.proto
  datetime.proto
  dayofweek.proto
  decimal.proto
  expr.proto
  fraction.proto
  interval.proto
  latlng.proto
  localized_text.proto
  money.proto
  month.proto
  phone_number.proto
  postal_address.proto
  quaternion.proto
  timeofday.proto

noahdietz:
@achew22 could you open an issue with this request in it? Probably a better place to discuss it :)

To be clear, it is about making common protos available for import resolution, correct?

PR #1001, opened 2022-08-19
chore: fix all lint errors

achew22:


achew22:
Sorry I couldn't figure out a less invasive fix than to just DO ALL THE THINGS, but at least the linter passes now. That's good

noahdietz:
> Sorry I couldn't figure out a less invasive fix than to just DO ALL THE THINGS, but at least the linter passes now. That's good

Oh no need to apologize @achew22 this is very helpful.

Can I ask, were you running tools via Go 1.19 or an earlier version? I ask because gofmt from Go 1.19 has some small differences and I want to know what to use in this repo.

achew22:
Correct, this is using the 1.19 toolchain. That said, my understanding of the 1.19 gofmt changes is that they are backward compatible. That is to say, if 1.19 runs on it, 1.18 will not *undo* the changes so I think you're safe to continue on your tooling journey.

noahdietz (approved):

PR #1013, opened 2022-09-08
fix(ci): use go 1.18 in release job

noahdietz:
The `release` job for `darwin/arm64` failed because the `gcc` toolchain for `darwin/arm64` is not available on ubuntu by default. Apparently Go 1.16 added support for this compilation target, so I'm updating to Go 1.18 so as to 1. capture this change and 2. get the release job a little more up to date with modern Go version. I built the X-compiled binary locally with no issue.

We should look at upping the version used across CI and in our `go.mod`.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   inspect:
     runs-on: ubuntu-latest
-    container: golang:1.14
+    container: golang:1.18
     steps:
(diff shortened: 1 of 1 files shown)


shwoodard (approved):

noahdietz:
cc @loeffel-io this is why the release job didn't work I think

loeffel-io:
@noahdietz amazing

Issue #972, opened 2022-05-13
Verify that resource referenced in request field has pattern(s) present in google.api.http annotations


noahdietz:
When an RPC has a resource name in its request, the name is often used as part of the HTTP binding. For example:

```proto
service FooService {
  rpc GetFoo(GetFooRequest) returns (Foo) {
    option (google.api.http) = {
      get: "v1/{name=projects/*/foos/*}"
    };
  };
}

message GetFooRequest {
  // Format: projects/{project}/foos/{foo}
  string name = 1 [(google.api.resource_reference).type = "foo.googleapis.com/Foo"];
}

message Foo {
  option (google.api.resource) = {
    type: "foo.googleapis.com/Foo"
    pattern: "projects/{project}/foos/{foo}"
  };

  string name = 1;
}
```

Here, `GetFooRequest.name` field is used to populate a section of the HTTP Request path. That field references a resource definition that defines a resource name pattern. The linter should enforce that the expected pattern(s) present in the `google.api.http` path can be found in the `google.api.resource` annotation for the referenced resource.

noahdietz:
acamadeo@ all yours!

noahdietz:
Fixed by #1045. Will be in the next release. Closing this as the _other_ case of child_type references in the `google.api.http` annotation is covered by #1047. Thank you @acamadeo

PR #1017, opened 2022-09-14
chore: add some helpers

noahdietz:
This adds a few helpers that will be useful in implementing #972. I started on that issue, and while I didn't get far, I figured these would be useful. @acamadeo will be taking on that issue now, so I want them to have the helpers.


rules/internal/utils/http.go
```diff
@@ -22,6 +22,13 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// HasHTTPRules returns whether the given method descriptor is annotated with
```
GregFurmanek, line 25: How about?

HasHTTPRules returns true when ...

rules/internal/utils/find_test.go
```diff
@@ -43,3 +44,46 @@ func TestFindMessage(t *testing.T) {
...
+
+			PublishingInfo publishing_info = 2;
+		}
+	`)
+	msg := file.GetMessageTypes()[0]
+
+	for _, tst := range []struct {
+		name, path string
+	}{
+		{"simple", "parent"},
```
GregFurmanek, line 74: simple -> top_level ?

rules/internal/utils/find.go
```diff
@@ -126,3 +126,26 @@ func GetRepeatedMessageFields(m *desc.MessageDescriptor) []*desc.FieldDescriptor
 
 	return fields
 }
+
+// FindFieldDotNotation attempts to find the field within the given message
+// identified by the dot-notation path e.g. book.name. This is especially useful
+// for resolving path variables in google.api.http and nested
+// fields in google.api.method_signature annotations.
+func FindFieldDotNotation(msg *desc.MessageDescriptor, f string) *desc.FieldDescriptor {
```
GregFurmanek, line 134: In essence this works like jquery, right?

So the method description might be:

Returns a field descriptor from a passed in message that corresponds to dot separated path. Example: ...

If path not found the method returns nil.

GregFurmanek, line 134: The f string parameter might be named "path" or something that would help reader understand the meaning.

rules/internal/utils/extension.go
```diff
@@ -147,3 +147,22 @@ func GetResourceReference(f *desc.FieldDescriptor) *apb.ResourceReference {
 	}
 	return nil
 }
+
+// FindResource attempts to find the google.api.resource(_definition) given the
```
GregFurmanek, line 151: Returns first resource of type matching the reference param.

rules/internal/utils/extension_test.go
```diff
@@ -312,3 +312,56 @@ func TestGetResourceReference(t *testing.T) {
...
+	})
+
+	for _, tst := range []struct {
+		name, reference string
+	}{
+		{"local_reference", "library.googleapis.com/Shelf"},
+		{"imported_reference", "library.googleapis.com/Book"},
+	} {
+		t.Run(tst.name, func(t *testing.T) {
+			if got := FindResource(tst.reference, files["shelf.proto"]); got == nil {
```
GregFurmanek, line 360: Can this be on a separate line?

In a way this single line notation hides the proper usage of the method with method's result verification.

let result = FindResource(...)

if result == nil {
 // error here
 return
}

if resutl.GetType() != tst.reference {
 // A different type of error
}

rules/internal/utils/extension_test.go
```diff
@@ -312,3 +312,56 @@ func TestGetResourceReference(t *testing.T) {
 		}
 	})
 }
+
+func TestFindResource(t *testing.T) {
```
GregFurmanek, line 316: Should there be a test that doesn't find a resource with requested type?

rules/internal/utils/http_test.go
```diff
@@ -148,3 +148,33 @@ func TestGetVariables(t *testing.T) {
 		})
 	}
 }
+
+func TestHasHTTPRules(t *testing.T) {
+	for _, tst := range []struct {
+		name       string
+		Annotation string
+	}{
+		{"has_rule", `option (google.api.http) = {get: "/v1/foos"};`},
```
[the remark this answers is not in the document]
noahdietz: The properties of the `google.api.http` method do not affect the ability of the library to extract it from the MethodOptions descriptor, so testing different values there will not have any effect on it.

rules/internal/utils/find_test.go
noahdietz: Done.

rules/internal/utils/find.go
noahdietz: I've already got a variable named `path` in the method body, so I've renamed `f` to `ref`.

rules/internal/utils/extension_test.go
noahdietz: Done.

rules/internal/utils/extension.go
noahdietz: Done.

rules/internal/utils/http_test.go
```diff
@@ -148,3 +148,33 @@ func TestGetVariables(t *testing.T) {
...
+						{{ .Annotation }}
+					}
+				}
+
+				message ListFoosRequest {}
+				message ListFoosResponse {}
+			`, tst)
+
+			want := tst.Annotation != ""
+			if got := HasHTTPRules(file.GetServices()[0].GetMethods()[0]); got != want {
```
GregFurmanek, line 175: This could also be split between the execution and check.

noahdietz: Done.

rules/internal/utils/find_test.go
```diff
@@ -43,3 +44,46 @@ func TestFindMessage(t *testing.T) {
...
+		name, path string
+	}{
+		{"top_level", "parent"},
+		{"nested", "book.name"},
+		{"double_nested", "book.publishing_info.publisher"},
+	} {
+		t.Run(tst.name, func(t *testing.T) {
+			split := strings.Split(tst.path, ".")
+			want := split[len(split)-1]
+			if f := FindFieldDotNotation(msg, tst.path); f == nil {
```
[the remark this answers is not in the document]
noahdietz: Done.

GregFurmanek (approved):
LGTM

PR #1035, opened 2022-10-25
feat: Add a new rule range for GURP AIPs

shwoodard:


lint/rule_groups.go
```diff
@@ -47,6 +47,13 @@ func aipCloudGroup(aip int) string {
 	return ""
 }
 
+func aipGURPGroup(aip int) string {
```
jskeet, line 50: I'd normally expect this to be aipGurpGroup, but I don't know what Go rules are normally like for acronyms.

shwoodard: Yes, golint even complains. :-)

Issue #1049, opened 2022-12-01
Validate that HTTPRules follow the path template syntax


acamadeo:
There's already an [existing rule](https://github.com/googleapis/api-linter/blob/c18f3154970793d9b984a376344d44e73ee4eb39/rules/aip0122/camel_case_uris.go#L27) that enforces the use of `camelCase` inside [HTTPRules](https://github.com/googleapis/googleapis/blob/16db2fb7fab4668bdfa09966513e03581d8f5e35/google/api/http.proto#L313). We could extend this by checking that the`HTTPRule` follows the path template syntax laid out [here](https://github.com/googleapis/googleapis/blob/16db2fb7fab4668bdfa09966513e03581d8f5e35/google/api/http.proto#L224).

noahdietz:
Fixed by #1051. If I'm mistaken, please ask for it to be reopened.

PR #1051, opened 2022-12-05
feat(AIP-127): Validate HTTP pattern syntax

acamadeo:
Part of #1049 

This constructs a regex containing the syntax rules defined [here](https://github.com/googleapis/googleapis/blob/83c3605afb5a39952bf0a0809875d41cf2a558ca/google/api/http.proto#L224). This approach is trickier to debug than writing a parser. However, it seems likely that with sufficient testing this rule will not require much debugging, especially since it does not seem the syntax rules will change anytime soon.

### DIFF
--- a/docs/rules/0127/http-template-syntax.md
+++ b/docs/rules/0127/http-template-syntax.md
@@ -1,0 +1,93 @@
+---
+rule:
+  aip: 127
+  name: [core, '0127', http-template-syntax]
+  summary: |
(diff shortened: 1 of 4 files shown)


noahdietz:
Try a PR title like `feat(AIP-127): enforce HTTP pattern formatting`. That will make the conventional commit bot happy :)

noahdietz (approved):
This is really good thanks @acamadeo.

PR #1074, opened 2023-01-28
feat(AIP-123): add validation of uppercase resource type

### LINKED ISSUE #1066: AIP-123: enforce UpperCamelCase of Resource Type names
http://aip.dev/123 states that the Type name in a resource definition must be UpperCamelCase e.g. `library.googleapis.com/BookLoan`. It is very easy to confuse `bookLoan` from the collection ID as the Type name and adding `library.googleapis.com/bookLoan`.

toumorokoshi:
fixes #1066. See AIP-0123 for details.

### DIFF
--- a/rules/aip0123/resource_type_name.go
+++ b/rules/aip0123/resource_type_name.go
@@ -15,12 +15,19 @@
 package aip0123
 
 import (
+	"fmt"
+	"regexp"
 	"strings"
+	"unicode"
 
 	"github.com/googleapis/api-linter/lint"
 	"github.com/googleapis/api-linter/locations"
 	"github.com/googleapis/api-linter/rules/internal/utils"
 	"github.com/jhump/protoreflect/desc"
+)
+
+var (
+	resourceNameRegex = regexp.MustCompile(`^[\p{L}A-Za-z0-9]+$`)
 )
 
 var resourceTypeName = &lint.MessageRule{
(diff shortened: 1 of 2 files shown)


rules/aip0123/resource_type_name.go
```diff
@@ -30,14 +33,39 @@ var resourceTypeName = &lint.MessageRule{
...
-
+		typeName := strings.Split(resourceType, "/")[1]
+		if !unicode.IsUpper(rune(typeName[0])) {
+			return []lint.Problem{{
+				Message:    fmt.Sprintf("Type %q must be UpperCamelCase", typeName),
+				Descriptor: m,
+				Location:   locations.MessageResource(m),
+			}}
+		}
+		matched, err := regexp.MatchString("^[\\p{L}A-Za-z0-9]+$", typeName)
```
noahdietz, line 52: Can you move this regexp into a package-global, unexported `var alphaNumericTypeName` and use [`regexp.MustCompile`](https://pkg.go.dev/regexp#MustCompile) instead? The `err` here is for compiling the regexp and I think things should fail much earlier if this regexp is wrong - most other regexps in this project are compiled via `MustCompile`. This way, we don't need to handle the `err`/report it to users on their proto.

noahdietz (commented):
One comment, but LGTM.

rules/aip0123/resource_type_name.go
toumorokoshi: great! that's much better. Will do.

noahdietz (approved):

rules/aip0123/resource_type_name.go
```diff
@@ -15,29 +15,53 @@
...
+	"unicode"
 
 	"github.com/googleapis/api-linter/lint"
 	"github.com/googleapis/api-linter/locations"
 	"github.com/googleapis/api-linter/rules/internal/utils"
 	"github.com/jhump/protoreflect/desc"
 )
 
+var (
+	resourceNameRegex = regexp.MustCompile("^[\\p{L}A-Za-z0-9]+$")
```
noahdietz, line 30:
```suggestion
	resourceNameRegex = regexp.MustCompile(`^[\\p{L}A-Za-z0-9]+$`)
```

This makes the go linter happy

noahdietz: I see several other instances of using `"` in stead of backtick...likely before the golang lint was added 🤷🏻

toumorokoshi: thank you! I just saw that as well. fixed.

noahdietz (approved):

PR #1078, opened 2023-02-04
fix(AIP-214): scope resource-expiry to resources

### LINKED ISSUE #1073: AIP-214: resource-expiry check should only apply to resource messages
The current check for the `resource-expiry` rule will run on any message with a field named `expire_time`. This isn't as targeted as the AIP is, which applies to resource messages. Using `expire_time` in a response message, for example, should be fine without being linted. The `OnlyIf` should likely check if the parent message is a Resource as well if the field is named `expire_time`.

toumorokoshi:
Only resources need the API-214 resource-expiry check.

fixes #1073

### DIFF
--- a/rules/aip0214/resource_expiry.go
+++ b/rules/aip0214/resource_expiry.go
@@ -23,7 +23,9 @@ import (
 var resourceExpiry = &lint.FieldRule{
 	Name: lint.NewRuleName(214, "resource-expiry"),
 	OnlyIf: func(f *desc.FieldDescriptor) bool {
-		return f.GetName() == "expire_time"
+		isResource := utils.IsResource(f.GetParent().(*desc.MessageDescriptor))
+		isExpireTime := f.GetName() == "expire_time"
+		return isResource && isExpireTime
 	},
 	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
 		// If this field is output only, then there is no user input permitted
--- a/rules/aip0214/resource_expiry_test.go
+++ b/rules/aip0214/resource_expiry_test.go
@@ -22,22 +22,28 @@ import (
 
 func TestResourceExpiry(t *testing.T) {
 	for _, test := range []struct {
-		name          string
-		FieldBehavior string
-		AddtlField    string
-		problems      testutils.Problems
+		name                  string
+		FieldBehavior         string
+		AddtlField            string
+		problems              testutils.Problems
+		AddResourceDefinition bool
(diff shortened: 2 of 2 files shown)


rules/aip0214/resource_expiry.go
```diff
@@ -23,7 +23,14 @@ import (
...
 	OnlyIf: func(f *desc.FieldDescriptor) bool {
-		return f.GetName() == "expire_time"
+		// only return true if the parent is a resource
+		if utils.GetResourceFromDescriptor(f.GetParent()) == nil {
+			return false
+		}
+		if f.GetName() != "expire_time" {
+			return false
+		}
+		return true
```
noahdietz, line 33:
```suggestion
		isResource := utils.GetResourceFromDescriptor(f.GetParent()) != nil
		isExpireTime := f.GetName() == "expire_time"
		return isResource && isExpireTime
```

Kind of a nit, more my preference: It is easier to understand what the conditions are when testing for the positive case i.e. "Lint this only if parent is a resource and the field name is expire_time"

rules/internal/utils/extension.go
```diff
@@ -104,6 +104,20 @@ func GetResource(m *desc.MessageDescriptor) *apb.ResourceDescriptor {
 	return nil
 }
 
+// GetResourceFromDescriptor returns the google.api.resource from
+// an arbitrary descriptor. If the descriptor is not a message descriptor,
+// nil is returned.
+func GetResourceFromDescriptor(d desc.Descriptor) *apb.ResourceDescriptor {
```
noahdietz, line 110: We probably don't need this actually, at least not for this use case.

The parent of a Field will always be a Message, so it is safe (and reasonable to `panic`) if the `desc.Descriptor` from `f.GetParent()` is not a `*desc.MessageDescriptor`.

So I would actually just use  something like `utils.IsResource((f.GetParent()).(*desc.MessageDescriptor))` instead of a `nil` comparison check with this.

rules/aip0214/resource_expiry.go
toumorokoshi: I prefer the other case slightly (guard conditions are a little easier to follow for me), but not strong enough that I won't take your suggestion. Thanks!

rules/internal/utils/extension.go
toumorokoshi: sure, sounds good.

noahdietz (approved):

PR #1080, opened 2023-02-06
chore: fix Constants and variables

aria1991:
I added some small improvements in terms of Constants and variables.

 — Updated `globalConfigs` to a constant with an empty string value
 — Simplified the `init` function to only contain the call to `rules.Add`, and remove the `if `statement as it is not needed since `rules.Add` will log the error on its own!
 — Replaced the inline comment in `defaultConfigs` with a clear and concise function comment


shwoodard:
This appears to change the type of `globalConfigs` from `lint.Configs` to `string`. Does this compile?

noahdietz:
Closing as inactive. Please open an issue first if there is something to fix.

Issue #1071, opened 2023-01-24
GitHub Action output format generates incorrect line numbers


jsocol:
When using the `--output-format github` switch, the resulting can place the `Span` values into the wrong places, e.g.

```json
[710, 32, 40]
```

produces

```
::error file=users/users.proto,endLine=40,col=32,line=710,title=core...
```

I believe the issue comes from [this switch in the formatter code](https://github.com/googleapis/api-linter/blob/main/cmd/api-linter/github_actions.go#L40-L50):

```go
				// Some findings are *line level* and only have start positions but no
				// starting column. Construct a switch fallthrough to emit as many of
				// the location indicators are included.
				switch len(problem.Location.Span) {
				case 4:
					fmt.Fprintf(&buf, ",endColumn=%d", problem.Location.Span[3])
					fallthrough
				case 3:
					fmt.Fprintf(&buf, ",endLine=%d", problem.Location.Span[2])
					fallthrough
				case 2:
					fmt.Fprintf(&buf, ",col=%d", problem.Location.Span[1])
					fallthrough
				case 1:
					fmt.Fprintf(&buf, ",line=%d", problem.Location.Span[0])
				}
```

Which does not treat the two different span lengths the same way as the JSON/YAML [marshaling code](https://github.com/googleapis/api-linter/blob/main/lint/problem.go#L134-L166)

```go
	// If `span` has four ints; they correspond to
	// [start line, start column, end line, end column].
	//
	// We add one because spans are zero-indexed, but not to the end column
	// because we want the ending position to be inclusive and not exclusive.
	if len(span) == 4 {
		return fileLocation{
			Start: position{
				Line:   int(span[0]) + 1,
				Column: int(span[1]) + 1,
			},
			End: position{
				Line:   int(span[2]) + 1,
				Column: int(span[3]),
			},
		}
	}

	// Okay, `span` has three ints; they correspond to
	// [start line, start column, end column].
	//
	// We add one because spans are zero-indexed, but not to the end column
	// because we want the ending position to be inclusive and not exclusive.
	return fileLocation{
		Start: position{
			Line:   int(span[0]) + 1,
			Column: int(span[1]) + 1,
		},
		End: position{
			Line:   int(span[0]) + 1,
			Column: int(span[2]),
		},
	}
```

I think a focused fix for the formatter might look like

```go
				startLine := problem.Location.Span[0] + 1
				startColumn := problem.Location.Span[1] + 1
				if len(problem.Location.Span) == 4 {
[...]

PR #1081, opened 2023-02-06
Fix/GitHub actions formatter

jsocol:
This fixes the issue in #1071 with the GitHub Action output using the wrong location values. It's in two commits. The first exposes `lint.FileLocationFromPBLocation` as a public method, and adds tests for it. The second uses that method to remove the conversion logic from the GitHub Action formatter and fixes the tests that had the file locations specified.


noahdietz:
cc @achew22 author of the original change for `github_actions` format

jsocol:
CLA signed and check re-run.

Definitely open to whatever feedback here. Like I mentioned on #1071, there's a more-focused path that only fixes the `Span`-reading logic in the `github` formatter, though it seems like centralizing that logic is a more future-proof solution. There are certainly other ways to do that.

noahdietz (commented):
I'm generally happy to take this change. I don't personally use this format though, which is why I'd like to see if @achew22 has any thoughts.

Thanks for opening this PR (and the issue that I never responded too 😅)

jsocol:
> Thanks for opening this PR (and the issue that I never responded too 😅)

As a fellow OSS maintainer, I share your 😅 . Thanks for this tool, it's hugely helpful!

cmd/api-linter/github_actions.go
```diff
@@ -33,22 +33,8 @@ func formatGitHubActionOutput(responses []lint.Response) []byte {
...
-					fmt.Fprintf(&buf, ",endLine=%d", problem.Location.Span[2])
-					fallthrough
-				case 2:
-					fmt.Fprintf(&buf, ",col=%d", problem.Location.Span[1])
-					fallthrough
-				case 1:
-					fmt.Fprintf(&buf, ",line=%d", problem.Location.Span[0])
-				}
+				location := lint.FileLocationFromPBLocation(problem.Location)
+				fmt.Fprintf(&buf, ",line=%d,col=%d,endLine=%d,endColumn=%d", location.Start.Line, location.Start.Column, location.End.Line, location.End.Column)
```
achew22, line 37: Why do we believe that all findings contain line and column numbers? Sorry if that's obvious from other changes in the project but this caused it to nil pointer exception for me before I made the change you're removing here

jsocol: This definitely depends on the logic in `lint/problem.go` being correct, but according to the [code comment on line 128](https://github.com/googleapis/api-linter/blob/b018fd917ad6a0fe8cae6735e7fccc2237a8dc79/lint/problem.go#L128) the underlying `Span` should be guaranteed to have either 3 (line,startCol,endCol) or 4 (startLine,startCol,endLine,endCol) values.

I'd be curious if you have a case to reproduce that, and if it also causes problems for the YAML and JSON output? I think those should panic on main if a span is ever too short

lint/problem.go
```diff
@@ -106,25 +106,25 @@ func (p Problem) GetRuleURI() string {
...
+type FileLocation struct {
+	Start Position `json:"start_position" yaml:"start_position"`
+	End   Position `json:"end_position" yaml:"end_position"`
 }
 
-// fileLocationFromPBLocation returns a new fileLocation object based on a
+// FileLocationFromPBLocation returns a new fileLocation object based on a
 // protocol buffer SourceCodeInfo_Location
-func fileLocationFromPBLocation(l *dpb.SourceCodeInfo_Location) fileLocation {
+func FileLocationFromPBLocation(l *dpb.SourceCodeInfo_Location) FileLocation {
```
jsocol, line 127: I realize now that this is more or less directly reversing a deliberate decision in #195 to not have two types of "location." If that's still desirable then maybe the other path of fixing the math directly in the github formatter, or possibly a third option of trying to centralize the "formatter" concept somewhere they could share utilities like this

quirogas:
Hi @jsocol, I realize it's been a while since we last addressed this pull request. Since we are currently releasing v2, and I believe this issue remains relevant, I think it's worth keeping #1071 open, but closing this pull request. Thank you so much for your contribution!

PR #1084, opened 2023-02-08
chore: switch to statticcheck, fix linting

toumorokoshi:
using staticcheck as golint has been deprecated.
    
golint errors ignore comments were also not being
honored, so switching to staticcheck enables the whole
repository pass some form of linter.

 Go lint had multiple findings when run on the full codebase.
    
 Fixing those to help ensure future contributors have less friction
 and confusion wondering if the errors were caused by them.


noahdietz (approved):
Awesome!

PR #1087, opened 2023-02-09
feat: enable better debugability for linting

### LINKED ISSUE #1086: feat: make api-linter more debuggable
```
$ api-linter service.proto
2023/02/08 19:10:09 runtime error: index out of range [1] with length 1
```

Very hard to debug!

Make it easier.

shwoodard:
Adds a debug flag to the cli and passes it to lint.Linter. If enabled, when a panic is recoverd, the stack will be printed by the runtime/debug package.

Closes #1086

### DIFF
--- a/cmd/api-linter/cli.go
+++ b/cmd/api-linter/cli.go
@@ -44,6 +44,7 @@ type cli struct {
 	EnabledRules            []string
 	DisabledRules           []string
 	ListRulesFlag           bool
+	DebugFlag               bool
 }
 
 // ExitForLintFailure indicates that a problem was found during linting.
@@ -62,6 +63,7 @@ func newCli(args []string) *cli {
 	var ruleEnableFlag []string
 	var ruleDisableFlag []string
 	var listRulesFlag bool
+	var debugFlag bool
 
 	// Register flag variables.
 	fs := pflag.NewFlagSet("api-linter", pflag.ExitOnError)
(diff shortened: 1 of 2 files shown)


noahdietz (approved):
tight!

cmd/api-linter/cli.go
```diff
@@ -177,7 +181,7 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
 	}
 
 	// Create a linter to lint the file descriptors.
-	l := lint.New(rules, configs)
+	l := lint.New(rules, configs, lint.Debug(c.DebugFlag))
```
toumorokoshi, line 184: nit: would a new parameter to New() be easier? In my general experience constructors just pass in the parameter that need to be configured, rather than builder options.

That said I guess this allows for one or more optional parameters, which I suppose Go doesn't allow you to do.

toumorokoshi (approved):
Nice! thanks.

cmd/api-linter/cli.go
toumorokoshi: oops.. I guess I merged the PR without giving you time to react. 

I'll resolve this and if you feel the refactor is beneficial, please add it!

noahdietz: Any change to the function signature is technically a breaking change.

Appending a variadic argument is, but a "lesser" one in that existing usages don't break in place unless some "advanced" methods are being used (type aliasing the signature, passing it as a "function pointer", using reflect (idk why anyone would here)).

And as you said, this allows us to add stuff later that would otherwise constitute another "hard" breaking change.

noahdietz: That said, if we want to avoid the break altogether, I'd recommend using a function to toggle it on like

```go
l := lint.New(rules, configs)
l.Debug(true)
```

PR #1091, opened 2023-02-10
feat: add the ability to ignore comment disables

toumorokoshi:
For strict enforcement of AIPs across an organization, it's valuable to disable the ability to have text in the proto disable linter rules.

### DIFF
--- a/cmd/api-linter/cli.go
+++ b/cmd/api-linter/cli.go
@@ -32,21 +32,23 @@ import (
 )
 
 type cli struct {
-	ConfigPath              string
-	FormatType              string
-	OutputPath              string
-	ExitStatusOnLintFailure bool
-	VersionFlag             bool
-	ProtoImportPaths        []string
-	ProtoFiles              []string
-	ProtoDescPath           []string
-	EnabledRules            []string
-	DisabledRules           []string
-	ListRulesFlag           bool
-	DebugFlag               bool
+	ConfigPath                string
+	FormatType                string
+	OutputPath                string
+	ExitStatusOnLintFailure   bool
+	VersionFlag               bool
+	ProtoImportPaths          []string
+	ProtoFiles                []string
+	ProtoDescPath             []string
+	EnabledRules              []string
+	DisabledRules             []string
+	ListRulesFlag             bool
+	DebugFlag                 bool
+	IgnoreCommentDisablesFlag bool
 }
 
 // ExitForLintFailure indicates that a problem was found during linting.
+//
 //lint:ignore ST1012 modifying this variable name is a breaking change.
 var ExitForLintFailure = errors.New("found problems during linting")
 
@@ -63,6 +65,7 @@ func newCli(args []string) *cli {
 	var ruleDisableFlag []string
 	var listRulesFlag bool
 	var debugFlag bool
+	var ignoreCommentDisablesFlag bool
 
 	// Register flag variables.
 	fs := pflag.NewFlagSet("api-linter", pflag.ExitOnError)
@@ -76,7 +79,8 @@ func newCli(args []string) *cli {
 	fs.StringArrayVar(&ruleEnableFlag, "enable-rule", nil, "Enable a rule with the given name.\nMay be specified multiple times.")
 	fs.StringArrayVar(&ruleDisableFlag, "disable-rule", nil, "Disable a rule with the given name.\nMay be specified multiple times.")
 	fs.BoolVar(&listRulesFlag, "list-rules", false, "Print the rules and exit.  Honors the output-format flag.")
-	fs.BoolVar(&debugFlag, "debug", false, "Run in deubug mode. Panics will print stack.")
+	fs.BoolVar(&debugFlag, "debug", false, "Run in debug mode. Panics will print stack.")
+	fs.BoolVar(&ignoreCommentDisablesFlag, "ignore-comment-disables", false, "If set to true, disable comments will be ignored.\nThis is helpful when strict enforcement of AIPs are necessary and\nproto definitions should not be able to disable checks.")
 
 	// Parse flags.
 	err := fs.Parse(args)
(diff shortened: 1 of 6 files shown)


cmd/api-linter/cli.go
```diff
@@ -180,7 +185,8 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
 	}
 
 	// Create a linter to lint the file descriptors.
-	l := lint.New(rules, configs, lint.Debug(c.DebugFlag))
+	l := lint.New(rules, configs, lint.Debug(c.DebugFlag),
+		lint.IgnoreCommentDisables(c.IgnoreCommentDisablesFlag))
```
shwoodard, line 189: Go does not have a line length restriction. I would suggest putting this on one line. It would be more standard.

lint/lint.go
```diff
@@ -42,6 +43,14 @@ func Debug(debug bool) LinterOption {
 	}
 }
 
+// IgnoreCommentDisables sets the flag for ignoring
+// comments which disable rules.
```
shwoodard, line 47: I think you can fit more words on the first line. I use `gq` in vim. Not sure what editor you use though.

cmd/api-linter/integration_test.go
```diff
@@ -96,6 +96,19 @@ func TestRules_DisabledByFileComments(t *testing.T) {
...
+			disableInFile := fmt.Sprintf("// (-- api-linter: %s=disabled --)", test.rule)
+			proto := disableInFile + "\n" + test.proto
+			_, result := runLinterWithFailureStatus(t, proto, "", []string{"--ignore-comment-disables"})
+			if !strings.Contains(result, test.rule) {
+				t.Errorf("Rule %q should be enabled despite a disable by file comments", test.rule)
+			}
+		})
+	}
+}
+
```
shwoodard, line 111: Please include a negative test case where you don't pass the flag and expect the rule to be disabled.

cmd/api-linter/integration_test.go
```diff
@@ -96,6 +96,19 @@ func TestRules_DisabledByFileComments(t *testing.T) {
...
 }
 
+func TestRules_DisabledByFileCommentsAndIgnoreCommentDisablesTrue(t *testing.T) {
+	for _, test := range testCases {
+		t.Run(test.testName, func(t *testing.T) {
+			disableInFile := fmt.Sprintf("// (-- api-linter: %s=disabled --)", test.rule)
+			proto := disableInFile + "\n" + test.proto
+			_, result := runLinterWithFailureStatus(t, proto, "", []string{"--ignore-comment-disables"})
+			if !strings.Contains(result, test.rule) {
+				t.Errorf("Rule %q should be enabled despite a disable by file comments", test.rule)
```
shwoodard, line 106: Errors should neither start with a capital letter nor end in a period. IOW, downcase, "Rule".

lint/rule_enabled.go
```diff
@@ -77,29 +78,32 @@ func ruleIsEnabled(rule ProtoRule, d desc.Descriptor, l *dpb.SourceCodeInfo_Loca
...
-		commentLines = append(commentLines, strings.Split(fileHeader(f), "\n")...)
-	} else {
-		commentLines = append(commentLines, strings.Split(getLeadingComments(d), "\n")...)
-	}
-	disabledRules := []string{}
-	for _, commentLine := range commentLines {
-		r := extractDisabledRuleName(commentLine)
-		if r != "" {
-			disabledRules = append(disabledRules, r)
+	if !ignoreCommentDisables {
```
shwoodard, line 82: I would move the entire contents of this block to a method like, `func ruleDisabledInline(...) bool` so that you don't have such a long block here.

shwoodard (changes requested):
Add a test in `lint_test.go` similar to [this](https://github.com/googleapis/api-linter/blob/3890e69e44084e1d2e2f83c3a3be54db9a2f4552/lint/lint_test.go#L156)?

lint/lint.go
toumorokoshi: I prefer shorter lines, but it's not a strong opinion at all. I'll change it.

cmd/api-linter/integration_test.go
toumorokoshi: there already is a negative case that doesn't pass in the flag (TestRules_DisabledByFileComment). are you suggesting I should explicitly test for `--ignore-test-disables=false`?

cmd/api-linter/integration_test.go
toumorokoshi: That's for the `err` type, correct? There are examples even in internal go styleguides that have capital letter:

```
func mustAddGameAssets(t *testing.T, dir string) {
    t.Helper()
    if err := os.WriteFile(path.Join(dir, "pak0.pak"), pak0, 0644); err != nil {
        t.Fatalf("Setup failed: could not write pak0 asset: %v", err)
    }
    if err := os.WriteFile(path.Join(dir, "pak1.pak"), pak1, 0644); err != nil {
        t.Fatalf("Setup failed: could not write pak1 asset: %v", err)
    }
}
```

toumorokoshi: I don't have a strong opinion so I'm fine with switching, but this was just the precedent of the whole file. I'll have to change them all.

lint/rule_enabled.go
toumorokoshi: ack.

toumorokoshi:
> Add a test in lint_test.go similar to [this](https://github.com/googleapis/api-linter/blob/3890e69e44084e1d2e2f83c3a3be54db9a2f4552/lint/lint_test.go#L156)?

This feels a little redundant to me to be honest (the integration test exercises this path), but I can add it.

cmd/api-linter/cli.go
toumorokoshi: sure, no opinion on line length.

cmd/api-linter/integration_test.go
noahdietz: Personally I think `t.Errorf` is not the same as constructing an `error` message returned to a user, so I use whatever casing :)

cmd/api-linter/integration_test.go
shwoodard: No that's fine.

toumorokoshi: did it anyway :)

toumorokoshi:
@shwoodard I think I addresses all of your comments. PTAL.

shwoodard (approved):

PR #1125, opened 2023-03-17
feat(AIP-122): enforce output_only resource IDs

noahdietz:
Adds rule to check that either Resource ID field allowed by AIP-122 is classified as `OUTPUT_ONLY`.

See https://github.com/aip-dev/google.aip.dev/pull/1040.

### DIFF
--- a/rules/aip0122/resource_id_output_only.go
+++ b/rules/aip0122/resource_id_output_only.go
@@ -1,0 +1,45 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aip0122
+
+import (
+	"strings"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+	"google.golang.org/genproto/googleapis/api/annotations"
+)
+
+var resourceIdOutputOnly = &lint.FieldRule{
+	Name: lint.NewRuleName(122, "resource-id-output-only"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		isRes := utils.IsResource(f.GetParent().(*desc.MessageDescriptor))
+		isId := f.GetName() == "uid" || strings.HasSuffix(f.GetName(), "_id")
+		return isRes && isId
+	},
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		behaviors := utils.GetFieldBehavior(f)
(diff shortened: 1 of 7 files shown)


rules/aip0122/resource_id_output_only_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func TestResourceIdOutputOnly(t *testing.T) {
+	for _, test := range []struct {
+		name          string
+		FieldName     string
+		FieldBehavior string
+		problems      testutils.Problems
+	}{
+		{"ValidWithSuffix", "book_id", "[(google.api.field_behavior) = OUTPUT_ONLY]", testutils.Problems{}},
```
toumorokoshi, line 30: do you think there's an issue where someone has a bespoke field with the `_id` suffix?

Or is there an AIP that prevents that?

toumorokoshi: a more robust way of doing this is to make the check look for the resource singular (e.g. extract some suffix from the message or type), then use `{resource}_id` for the check. 

But there might be enough consistency to do that today (the AIPs don't enforce message/type consistency.

toumorokoshi: Actually, I think something like this has to be done, since [aip-122 explicitly states](https://google.aip.dev/122#fields-representing-another-resource):

> If using the resource name is not possible and using the ID component alone is strictly necessary, the field should use an _id suffix (e.g. shelf_id).

which means it's very possible that users are using `{some_other_resource}_id` as a user-settable value today.

toumorokoshi (commented):
so I think there's a snag here - see my comment. I don't have a great shortcut to solve it - I think we're missing some vital information about the resource singular name to solve it. 

Anyway let's discuss in-thread.

rules/aip0122/resource_id_output_only_test.go
noahdietz: > resource singular , then use `{resource}_id` for the check.
>
> But there might be enough consistency to do that today

Yeah I think things are close enough today to do this. While we'd like the message and resource type to be the same, if the linter consistently uses the `type` name/`singular` field, I think that is always the right answer (even if the message name is different).

 I'll add this logic.

noahdietz: Done! Also did a bit of refactoring to consolidate `Resource.type` parsing logic.

rules/aip0122/resource_id_output_only.go
```diff
@@ -0,0 +1,60 @@
...
+
+var resourceIdOutputOnly = &lint.FieldRule{
+	Name: lint.NewRuleName(122, "resource-id-output-only"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		var idName string
+		p := f.GetParent().(*desc.MessageDescriptor)
+
+		// Build an expected ID field name based on the Resource `singular`
+		// field or by parsing the `type`.
+		isRes := utils.IsResource(p)
```
toumorokoshi, line 33: nit: this could be reduced to just `utils.GetResource(p)` with a nil pointer check?

toumorokoshi (approved):
awesome, thanks!

rules/aip0122/resource_id_output_only.go
noahdietz: I thought about that and then kept it this way for a few reasons:
1. because this API exists to use and consolidate that `!= nil` check in case we ever need to refactor that
2. because we use the `isRes` later on in the `return`, outside of the `if isRes`, so makes sense to keep a local variable

So I will keep it as is for now.

PR #1135, opened 2023-04-13
feat(AIP-148): use uid instead of id

### LINKED ISSUE #1068: AIP-148: Recommend uid instead of id
In http://aip.dev/148, the field `uid` is considered a standard field (used by http://aip.dev.122). When there is an `string id` field on a resource message, it should be suggested as `uid` instead.

noahdietz:
The standard field for the Resource ID segment within a resource is `uid`, not `id` per AIP-148. So let's enforce that.

Fixes #1068

### DIFF
--- a/docs/rules/0148/use-uid.md
+++ b/docs/rules/0148/use-uid.md
@@ -15,9 +15,8 @@ mandated in [AIP-148][].
 
 ## Details
 
-This rule looks for fields named `first_name` and `last_name`, and complains if
-it finds them, suggesting the use of `given_name` and `family_name`
-(respectively) instead.
+This rule looks on resource messages for a field named `id`, complains if it
+finds it, and suggests the use of `uid` instead.
 
 ## Examples
 
@@ -25,9 +24,14 @@ it finds them, suggesting the use of `given_name` and `family_name`
 
 ```proto
 // Incorrect.
-message Human {
-  string first_name = 1;  // Should be `given_name`.
-  string last_name = 2;   // Should be `family_name`
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
+  };
+
+  string name = 1;
+  string id = 2; // Should be `uid`
 }
 ```
 
@@ -35,9 +39,14 @@ message Human {
 
 ```proto
 // Correct.
-message Human {
-  string given_name = 1;
-  string family_name = 2;
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
+  };
+
+  string name = 1;
+  string uid = 2;
 }
 ```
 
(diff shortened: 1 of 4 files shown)


docs/rules/0148/use-uid.md
```diff
@@ -0,0 +1,63 @@
...
+---
+
+# Human names
+
+This rule encourages the use of `uid` instead of `id` for resource messages, as
+mandated in [AIP-148][].
+
+## Details
+
+This rule looks for fields named `first_name` and `last_name`, and complains if
```
toumorokoshi, line 18: looks like this should be changed?

docs/rules/0148/use-uid.md
```diff
@@ -0,0 +1,63 @@
...
+(respectively) instead.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message Human {
+  string first_name = 1;  // Should be `given_name`.
```
toumorokoshi, line 29: change examples too?

docs/rules/0148/use-uid.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the field or its
+enclosing message. Remember to also include an [aip.dev/not-precedent][]
+comment explaining why.
+
+```proto
+// (-- api-linter: core::0148::human-names=disabled
```
toumorokoshi, line 51: change api linter example?

side note: I think we could generate a ton of this documentation with the right metadata.

rules/aip0148/use_uid_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/googleapis/api-linter/rules/internal/testutils"
+)
+
+func TestUseUid(t *testing.T) {
+	for _, test := range []struct {
+		FieldName string
+		problems  testutils.Problems
+	}{
+		{"uid", nil},
+		{"id", testutils.Problems{{Suggestion: "uid"}}},
```
toumorokoshi, line 29: maybe a quick test case for an unrelated field (e.g. `foo_id`?)

toumorokoshi (changes requested):
Thanks! I think the example and docs need to be updated, but the code LGTM.

rules/aip0148/use_uid_test.go
noahdietz: Done.

docs/rules/0148/use-uid.md
noahdietz: Fixed

docs/rules/0148/use-uid.md
noahdietz: Done.

docs/rules/0148/use-uid.md
noahdietz: Ha yeah, forgot to update this completely. Done.

rules/aip0148/use_uid.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+var useUid = &lint.FieldRule{
+	Name: lint.NewRuleName(148, "use-uid"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		return utils.IsResource(f.GetOwner())
+	},
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		want := "uid"
+		target := "id"
```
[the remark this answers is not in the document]
noahdietz: Done! Good idea.

toumorokoshi (approved):
Awesome, thanks!

shwoodard (approved):

PR #1142, opened 2023-05-06
feat(AIP-133): enforce required request fields

toumorokoshi:
Enforces the following from AIP-133:

    The request message must not contain any other required fields ...
    except those described in this or another AIP.


rules/aip0133/request_required_fields.go
```diff
@@ -0,0 +1,64 @@
...
+	Name:   lint.NewRuleName(133, "request-required-fields"),
+	OnlyIf: isCreateRequestMessage,
+	LintMessage: func(m *desc.MessageDescriptor) (problems []lint.Problem) {
+		resourceMsgName := getResourceMsgNameFromReq(m)
+
+		// Rule check: Establish that there are no unexpected fields.
+		allowedRequiredFields := map[string]*builder.FieldType{
+			"parent": nil, // AIP-133
+			fmt.Sprintf("%s_id", strings.ToLower(strcase.SnakeCase(resourceMsgName))): nil,
+		}
```
shwoodard, line 39: This feel like it should be defined as a var at the package level rather than in a method.

rules/aip0133/request_required_fields.go
```diff
@@ -0,0 +1,64 @@
...
+
+		for _, fieldDesc := range m.GetFields() {
+			if msgDesc := fieldDesc.GetMessageType(); msgDesc != nil && msgDesc.GetName() == resourceMsgName {
+				allowedRequiredFields[fieldDesc.GetName()] = nil // AIP-133
```
shwoodard, line 43: Use `delete`?

shwoodard (changes requested):

rules/aip0133/request_required_fields.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+		}
```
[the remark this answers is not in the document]
toumorokoshi: No problem, this is using a pattern from request_unknown_fields to populate the
field that represents the resource body:
(rules/aip0133/request_unknown_fields.go L35)

There's probably a more readable way to do this, I'll take a look.

rules/aip0133/request_required_fields.go
toumorokoshi: delete wouldn't work here, since it's using a map as a set and existence in the
map (even if the value is nil) is the way to check that it exists in the set.

rules/aip0133/request_required_fields.go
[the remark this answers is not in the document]
toumorokoshi: Okay found one, PTAL. I refactored the other one as well to hopefully be more readable.

rules/aip0133/request_required_fields.go
shwoodard: `val, ok := map[key]` ok is false if key is not in the map.

rules/aip0133/request_required_fields.go
```diff
@@ -0,0 +1,64 @@
...
+				continue
+			}
+			// Iterate remaining fields. If they're not in the allowed list,
+			// add a problem.
+			if _, ok := allowedRequiredFields[string(f.GetName())]; !ok {
+				problems = append(problems, lint.Problem{
+					Message: fmt.Sprintf(
+						"Create RPCs must only require fields explicitly described in AIPs, not %q.",
+						f.GetName(),
+					),
```
shwoodard, line 56: Move fmt.Sprintf to one line given that there is no line limit in Go and this is more common.

rules/aip0133/request_required_fields.go
shwoodard: Ack

shwoodard:
Do we have a complementary Rule that checks that fields that should be required have `REQUIRED` field_behavior on them?

toumorokoshi:
> Do we have a complementary Rule that checks that fields that should be required have REQUIRED field_behavior on them?

Yep we're covered there:

https://linter.aip.dev/133/request-resource-behavior
https://linter.aip.dev/133/request-parent-behavior

shwoodard:
> > Do we have a complementary Rule that checks that fields that should be required have REQUIRED field_behavior on them?
> 
> Yep we're covered there:
> 
> https://linter.aip.dev/133/request-resource-behavior https://linter.aip.dev/133/request-parent-behavior

Great!

PR #1143, opened 2023-05-06
feat(AIP-123): add singular and plural validation

### LINKED ISSUE #722: AIP-122, AIP-123: Validate the `singular` and `plural` fields of google.api.resource annotations.
I think we're missing a linter that enforces consistency between the plural/singular fields within [google.api.ResourceDescriptor][resource-descriptor] - specifically:

- `singular` should be in sync with `type` (the resource type)
- `plural` should be in sync with the collection identifier in `pattern` (the resource name)

[resource-descriptor]: https://github.com/googleapis/googleapis/blob/master/google/api/resource.proto

From AIP-122 (Resource names):

> **The collection identifier segments in a resource name must be the plural form of the noun used for the resource.** (For example, a collection of Publisher resources is called publishers in the resource name.)

From AIP-123 (Resource types):

> APIs must define a resource type for each resource in the API, according to the following pattern: {Service Name}/{Type}. **The type name must be singular and use PascalCase** (UpperCamelCase).

From [google/api/resource.proto][resource-descriptor]:

>  // **The plural name used in the resource name and permission names**, such as
>  // 'projects' for the resource name of 'projects/{project}' and the permission
>  // name of 'cloudresourcemanager.googleapis.com/projects.get'. It is the same
>  // concept of the `plural` field in k8s CRD spec
>  // https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/
>  //
>  // Note: The plural form is required even for singleton resources. See
>  // https://aip.dev/156
> string plural = 5;
>
>  // The same concept of the `singular` field in k8s CRD spec
>  // https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/
>  // **Such as "project" for the `resourcemanager.googleapis.com/Project` type**.
>  string singular = 6;

From [the k8s CRD spec][k8s-crd]:

> // **plural name to be used in the URL**: /apis/<group>/<version>/<plural>
> plural: crontabs

[k8s-crd]: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/

As such, what are your thoughts about a linter rules that enforce:

* The plural field must be URL safe (i.e. no whitespace) and use lowerCamelCase
* The collection identifiers in the resource name `pattern` values match the `plural` field
* The `Type` component of the resource `type` field (`{Service Name}/{Type}`) matches the UpperCamelCase version of the `singular` field

toumorokoshi:
Matching AIP-123 in requiring singular and plural annotations, and for singular matching the lowerCamelCase of type.

Fixes #722, but needs guidance updated in the AIPs around the values first.


rules/aip0123/resource_singular.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/stoewer/go-strcase"
+)
+
+var resourceSingular = &lint.MessageRule{
+	Name: lint.NewRuleName(123, "resource-singular"),
+	OnlyIf: func(m *desc.MessageDescriptor) bool {
+		r := utils.GetResource(m)
+		return r != nil || r.GetType() == ""
```
noahdietz, line 31:
```suggestion
		return r != nil && r.GetType() != ""
```
Should it be "type is not empty"? Otherwise, a non-resource message will always get a Problem, wouldn't it?

rules/aip0123/resource_singular.go
```diff
@@ -0,0 +1,58 @@
...
+				Location:   l,
+			}}
+		}
+		_, typeName, ok := utils.SplitResourceTypeName(r.GetType())
+		lowerTypeName := strcase.LowerCamelCase(typeName)
+		if !ok || lowerTypeName != s {
+			return []lint.Problem{{
+				Message: fmt.Sprintf(
+					"Resource singular should be lower camel case of type: %q, not %q.",
+					lowerTypeName, s,
```
noahdietz, line 50: Nit: IDK why but I feel like including what it _shouldn't_ be (i.e. what it already is in the CL) is extraneous, so I'd just do the following:
```suggestion
					"Resource singular should be lower camel case of type: %q.",
					lowerTypeName,
```

rules/aip0123/resource_plural.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var resourcePlural = &lint.MessageRule{
+	Name:   lint.NewRuleName(123, "resource-plural"),
+	OnlyIf: hasResourceAnnotation,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
```
noahdietz, line 27: Do you want to include a format check too e.g. lowerCamelCase only?

rules/aip0123/resource_singular.go
```diff
@@ -0,0 +1,58 @@
...
+	OnlyIf: func(m *desc.MessageDescriptor) bool {
+		r := utils.GetResource(m)
+		return r != nil || r.GetType() == ""
+	},
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		r := utils.GetResource(m)
+		l := locations.MessageResource(m)
+		s := r.GetSingular()
+		if s == "" {
+			return []lint.Problem{{
```
noahdietz, line 38: Can you include a `Suggestion: fmt.Sprintf("singular: %q", strcase.LowerCamelCase(typeName)),`? Will that land in the right spot given the location?

I'd say the same for `plural` but I think it's a tad harder to guess the correct pluralization even with a helper lib :)

rules/aip0123/aip0123.go
```diff
@@ -35,7 +35,9 @@ func AddRules(r lint.RuleRegistry) error {
 		resourceAnnotation,
 		resourceNameField,
 		resourcePattern,
+		resourcePlural,
 		resourceReferenceType,
+		resourceSingular,
```
noahdietz, line 40: I think another idea for these annotations, `singular` and `plural` is to ensure that they appear in the `pattern`(s) as the resource_id segment and collectionId segment respectively.

You can save it for another PR though, feel free to file another issue or add it to the existing issue if you'd like, for future implementation by you or someone else.

rules/aip0123/resource_singular.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/stoewer/go-strcase"
+)
+
+var resourceSingular = &lint.MessageRule{
+	Name: lint.NewRuleName(123, "resource-singular"),
+	OnlyIf: func(m *desc.MessageDescriptor) bool {
```
noahdietz, line 29: Can this just use `hasResourceAnnotation` like the `plural` rule does? Would resolve my other comment on this logic too.

rules/aip0123/resource_plural.go
toumorokoshi: done!

rules/aip0123/aip0123.go
toumorokoshi: ack,tracking here: https://github.com/googleapis/api-linter/issues/1146

rules/aip0123/resource_singular.go
toumorokoshi: I'll take a stab. It's a little bit more code but not terrible.

rules/aip0123/resource_singular.go
toumorokoshi: done

toumorokoshi:
@noahdietz finally ready for a pass! Thanks.

rules/internal/utils/casing.go
```diff
@@ -0,0 +1,68 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+// ToUpperCamelCase returns the UpperCamelCase of a string, including removing
+// delimiters (_,-,., ) and using them to denote a new word.
+func ToUpperCamelCase(s string) string {
```
noahdietz, line 19: Sooooo we could totally switch to using this but....we already use `strcase` for this type of stuff: https://pkg.go.dev/github.com/stoewer/go-strcase 😬 Not sure we need this file, unless we want to drop the external/non-google dependency (which would be fine if that was a goal, I do see value in such a "clean" module approach)

toumorokoshi: strcase doesn't filter out "-" or "_" unfortunately. I did see some hacks with double-conversion to snake case, then to upperCamelCase, but I figured it might be best to roll our own that matches our own definition of "upperCamelCase".

I did find this version of strcase that matches our definition: https://github.com/iancoleman/strcase. But I decided against switching it because we need to vet that it works for the other use cases as well.

noahdietz (approved):

rules/internal/utils/casing.go
toumorokoshi: ack, here's a tracking issue: https://github.com/googleapis/api-linter/issues/1150. Agreed that mixed usage is confusing: i'll follow up to try to clean that up. 

Once nice thing about our own handrolled one is we can add options as needed.

PR #1149, opened 2023-05-11
feat(AIP-203): require field_behavior

toumorokoshi:
Adding guidance from https://github.com/aip-dev/google.aip.dev/pull/1100


rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+
+	"bitbucket.org/creachadair/stringset"
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var requiredOneOfFieldBehavior = stringset.New(
+	"OPTIONAL", "REQUIRED", "OUTPUT_ONLY",
```
shwoodard, line 26: Should this include `IMMUTABLE`?

toumorokoshi: Is immutable enough by itself? I figured that we need immutable + (optional | required) to fully define the field. 

there's also discussion about immutable + output_only being valid, with immutable just stating it doesn't change for the lifecycle of the request.

shwoodard: IMO, `IMMUTABLE` is enough. It has sufficient semantic meaning on its own but maybe check with others too?

toumorokoshi: is an immutable field required? 

for a third party, we could maybe loop in @jskeet or @noahdietz?

The result would be an AIP change most likely.

noahdietz: I wouldn't say that Immutable is sufficient to communicate if the field is required or not. An Immutable field could be optional, and the service default is used if unset, which then cannot be changed.

Immutable only conveys the mutability of a field post-resource creation. It does not convey what constitutes a valid request.

shwoodard: The md suggests that it checks if the field_behavior has a valid value not if it is sufficient to determine if it is required. IMO, `IMMUTABLE` is a valid value.

toumorokoshi: ack. would you prefer a separate rule to force adding one of REQUIRED / OPTIONAL?

This is what the corresponding AIP change is proposing (which, BTW, needs reviews as well if someone is able to spare some time: https://github.com/aip-dev/google.aip.dev/pull/1100)

docs/rules/0203/field-behavior-required.md
```diff
@@ -0,0 +1,69 @@
...
+  REQUIRED, or OPTIONAL.
+permalink: /203/field-behavior-required
+redirect_from:
+  - /0203/field-behavior-required
+---
+
+# Field Behavior Required
+
+This rule enforces that each field has a `google.api.field_behavior` annotation
+with valid values, as mandated by[AIP-203][]
```
noahdietz, line 15: Nit: missing space
```suggestion
with valid values, as mandated by [AIP-203][]
```

docs/rules/0203/field-behavior-required.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Details
+
+This rule looks at all fields and ensures they have a
+`google.api.field_behavior` annotation. It also verifies that they have at least
+one of the options `OUTPUT_ONLY`, `REQUIRED`, or `OPTIONAL`: all fields must
+fall into of these categories.
```
noahdietz, line 22: missing a word
```suggestion
fall into one of these categories.
```

rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+
+	"bitbucket.org/creachadair/stringset"
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var requiredOneOfFieldBehavior = stringset.New(
```
noahdietz, line 25: I'd recommend a different name b.c it sounds like it pertains to `oneof` fields only. Maybe `minimumFieldBehaviorValues` or `minimumRequiredFieldBehavior` or something

rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,51 @@
...
+var fieldBehaviorRequired = &lint.FieldRule{
+	Name: lint.NewRuleName(203, "field-behavior-required"),
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		fieldBehavior := utils.GetFieldBehavior(f)
+		if len(fieldBehavior) == 0 {
+			return []lint.Problem{{
+				Message:    "google.api.field_behavior annotation must be set",
+				Descriptor: f,
+			}}
+		}
```
noahdietz, line 38: What about `oneof` fields? how do you think this pertains to them? All `REQUIRED` or all `OPTIONAL` or all `OUTPUT_ONLY`? Sounds like another rule we need to enforce that they are the same.

rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var requiredOneOfFieldBehavior = stringset.New(
+	"OPTIONAL", "REQUIRED", "OUTPUT_ONLY",
+)
+
+var fieldBehaviorRequired = &lint.FieldRule{
+	Name: lint.NewRuleName(203, "field-behavior-required"),
```
noahdietz, line 30: I think it is pretty obvious that this will result in many many more findings for folks, especially with the cascading affect of 1 field being added, field_behavior is recommended, then it is added, then the following finding from the  "consistency" rule e.g. "all fields should have field_behavior or none", followed by any other ApiLinter rule that fires on the fields being changed just to add field_behavior.

I'm not going to lie, I'm nervous about the impact that will have on users and on the perception of ApiLinter.

noahdietz, line 30: I kind of liked the guidance in AIP-203 for OUTPUT_ONLY:
> Notably, fields in response messages (a message which only ever acts as a return value to an RPC, usually 
> ending in Response) **should not** be described as output only because this is already implied.

Do you think that is worth incorporating in this rule? e.g. only lint fields that are used in the context of a request. I realize that will complicate the `OnlyIf`, but I want your take.

docs/rules/0203/field-behavior-required.md
toumorokoshi: thanks!

docs/rules/0203/field-behavior-required.md
toumorokoshi: thanks!

rules/aip0203/field_behavior_required.go
toumorokoshi: when with the latter. Agree with the OneOf confusion, good call!

rules/aip0203/field_behavior_required.go
toumorokoshi: I'm definitely open to suggestions. 

I thought about the cascading effect issue, and thought about maybe adding to
the error message that one of each must be set?

I think we agree that having these fields be required for new fields is better
than not having them. And hopefully, since findings only affect modified lines,
we should see gradual introduction to this new behavior.

The only issue I can think of is that the error is annoying enough that we don't
want to introduce it altogether. Which, I think based on the impact to clients,
is a price we have to pay.

I'm open to a high-bandwidth discussion on this too. WDYT?

toumorokoshi: ac k and agreed, I'll filter out responses.

noahdietz: FWIW the scope of a "response only" message is larger than just "messages that end with Response". Not sure how deep/broad you want to go here, but any nested message in a response (or lro metadata) message would be in scope of the exception.

toumorokoshi: yeah I just realized that. 

The only way I can think of is another service message that looks at RPCs ->
requests, then iterate through messages that way. thoughts?

This rule is going to be pretty loud (not noisy, it's very clear feedback) so I agree with reducing
the cases where it runs here.

noahdietz: I think we should add this rule, and do one of the follow:
- refine the scope to only apply to Resource & Request messages (you've kind of started on this already) -> this reduces the blast radius
- remove the "optional-consistency" [rule](https://linter.aip.dev/203/optional-consistency) (I guess its just for OPTIONAL, which is better than the alternative) -> eliminates the cascading issue
- do nothing different, we really do want anything & everything annotated, see what happens

toumorokoshi: I'll likely go for:

- refine scope to resource + request messages. I *think* I can do this by making it a method Lint and grabbing the request time. If we recursively dig through fields and look at their structs it should work.
- remove optional consistency. 

If that sounds good.

rules/aip0203/field_behavior_required.go
toumorokoshi: yep makes sense. tracking issue: https://github.com/googleapis/api-linter/issues/1157

rules/aip0203/field_behavior_required.go
toumorokoshi: I don't think immutable is sufficient information for a client. 

For example, if it's `IMMUTABLE`, is the field required or optional? IMO a client can't generate itself unless it know that answer.

rules/aip0203/field_behavior_required.go
noahdietz: Yeah I think that sounds good.

noahdietz (commented):
LGTM great job. I'll let @shwoodard approve as I think he's got an open question still

rules/aip0203/field_behavior_required.go
toumorokoshi: So I stand by what I said, but the linter seem to imply that OPTIONAL is mutually exclusive.

I've written an issue but for this and I'll round back (https://github.com/aip-dev/google.aip.dev/issues/1119). For now I'll accept that IMMUTABLE by itself is a valid annotation, because somehow it implies required.

rules/aip0203/field_behavior_required.go
toumorokoshi: done!

rules/aip0203/field_behavior_required.go
noahdietz: > IMMUTABLE by itself is a valid annotation, because somehow it implies required.

That's news to me 🙃

shwoodard: > I don't think immutable is sufficient information for a client.
> 
> For example, if it's `IMMUTABLE`, is the field required or optional? IMO a client can't generate itself unless it know that answer.

Ack. Do we say anywhere in an AIP that use of `IMMUTABLE` **must** be accompanied by another FieldBehavior value?

toumorokoshi: no, although it is noted that *optional* can only be used if no other annotation is used (which, IMO, is not great behavior).

I've filed https://github.com/aip-dev/google.aip.dev/pull/1120 to fix that. I'll follow-up with another CL after this one once the referenced PR is merged.

rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		p := checkFieldBehavior(f)
+		if p != nil {
+			problems = append(problems, *p)
+		}
+	}
+	return problems
+}
+
+func checkFieldBehavior(f *desc.FieldDescriptor) *lint.Problem {
```
shwoodard, line 54: Why return a reference and then indirect in the calling function. Recommend, changing this to return `lint.Problem` and then remove the `&` below and the `*` above.

rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,75 @@
...
+	fieldBehavior := utils.GetFieldBehavior(f)
+	if len(fieldBehavior) == 0 {
+		return &lint.Problem{
+			Message: fmt.Sprintf("google.api.field_behavior annotation must be set, and have one of %v",
+				minimumRequiredFieldBehavior,
+			),
```
shwoodard, line 60: nit: atypical to put this on multiple lines.

rules/aip0203/field_behavior_required.go
```diff
@@ -0,0 +1,75 @@
...
+			Descriptor: f,
+		}
+	}
+	// check for at least one valid annotation
+	if !minimumRequiredFieldBehavior.Intersects(fieldBehavior) {
+		return &lint.Problem{
+			Message: fmt.Sprintf(
+				"google.api.field_behavior must have at least one of the following behaviors set: %v",
+				minimumRequiredFieldBehavior,
+			),
```
shwoodard, line 70: nit: atypical to put this on multiple lines.

toumorokoshi: 120+ chars on a line feels long to me, but I'm not strongly opinionated.

rules/aip0203/field_behavior_required.go
toumorokoshi: there could be no problems with the field behavior, note the nil at the end.

But if desirable I can just switch to a problem list like the others: this is a
little more versatile anyway.

toumorokoshi: done, PTAL.

toumorokoshi:
@shwoodard thanks! ready for another pass from you. Noah deferred to your approval since you had blocking comments before.

shwoodard (approved):

Issue #1144, opened 2023-05-06
feat(AIP-122): enforce resource names as alternating collection and identifiers.


toumorokoshi:
AIP-122 states:

> Resource name components should usually alternate between collection identifiers (example: publishers, books, users) and resource IDs (example: 123, les-miserables, vhugo1802).

toumorokoshi:
fixed!

PR #1151, opened 2023-05-12
feat(AIP-123): verify resource patterns alternate

toumorokoshi:
Linter to verify the folowing from AIP-122:

Resource name components should usually alternate between collection identifiers (example: publishers, books, users) and resource IDs (example: 123, les-miserables, vhugo1802).

fixes feat(AIP-122): enforce resource names as alternating collection and identifiers. #1144

### DIFF
--- a/rules/aip0122/resource_name_components_alternate.go
+++ b/rules/aip0122/resource_name_components_alternate.go
@@ -1,0 +1,57 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aip0122
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var resourceNameComponentsAlternate = &lint.MessageRule{
+	Name:   lint.NewRuleName(122, "resource-name-components-alternate"),
+	OnlyIf: utils.IsResource,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		resource := utils.GetResource(m)
+		for _, p := range resource.GetPattern() {
+			expectIdentifier := false
+			components := strings.Split(p, "/")
+			for _, c := range components {
+				if expectIdentifier != isIdentifier(c) {
+					problems = append(problems, lint.Problem{
+						Message: fmt.Sprintf(
+							"Resource pattern %q must alternate between collection and identifier.",
+							p,
+						),
+						Descriptor: m,
+						Location:   locations.MessageResource(m),
+					})
+					break
+				}
+				expectIdentifier = !expectIdentifier
+			}
+		}
+		return problems
+	},
+}
+
+func isIdentifier(s string) bool {
+	return s[0] == '{'
+}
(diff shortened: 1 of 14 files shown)


rules/aip0122/resource_name_components_alternate.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+var resourceNameComponentsAlternate = &lint.MessageRule{
+	Name:   lint.NewRuleName(122, "resource-name-components-alternate"),
+	OnlyIf: utils.IsResource,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		resource := utils.GetResource(m)
+		for _, p := range resource.GetPattern() {
+			expectIdentifier := false
```
shwoodard, line 34: `identifierExpected` sounds more like a boolean whereas with the current variable name, I would expect a non-bool value.

rules/aip0122/resource_name_components_alternate.go
```diff
@@ -0,0 +1,57 @@
...
+							"Resource pattern %q must alternate between collection and identifier.",
+							p,
+						),
+						Descriptor: m,
+						Location:   locations.MessageResource(m),
+					})
+					break
+				}
+				expectIdentifier = !expectIdentifier
+			}
```
shwoodard, line 49:
```suggestion
			parts := strings.Split(p, "/")
			for i, part := range parts {
				if i % 2 == 1 && !isIdentifier(c) {
					problems = append(problems, lint.Problem{
						Message: fmt.Sprintf("Resource pattern %q must alternate between collection and identifier. %q is not an identifier", p, part),
						Descriptor: m,
						Location:   locations.MessageResource(m),
					})

					break
				}
			}
```

I find the switching `expectIdentifier` to be confusing. Use of the mod operator seems clearer?

rules/aip0122/resource_name_components_alternate.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		return problems
+	},
+}
+
+func isIdentifier(s string) bool {
+	return s[0] == '{'
```
shwoodard, line 56: Guard against `len(s)` is zero.

shwoodard, line 56: Consider using regexp to check for trailing `}` and at least one char in between, `/\{[a-z][a-zA-Z]*\}/`?

shwoodard (changes requested):

rules/aip0122/resource_name_components_alternate.go
shwoodard: Also, go has no line length restrictions, and putting the whole of fmt.Sprintf on one line would be more common.

rules/aip0122/resource_name_components_alternate.go
toumorokoshi: ack. I think it'll disappear in the refactor you suggested but I'll carry that
point with me in future code!

rules/aip0122/resource_name_components_alternate.go
toumorokoshi: ack on both, I applied the suggestion but left "parts" as "components", since
that's the verbage used in the AIP: https://google.aip.dev/122#guidance.

rules/aip0122/resource_name_components_alternate.go
toumorokoshi: with the regexp this is now guarded!

toumorokoshi:
fyi I moved this to AIP-123 since `pattern` is dictated in the AIP-123, and it looks like other pattern verification is there.

rules/aip0123/resource_name_components_alternate.go
```diff
@@ -0,0 +1,56 @@
...
+	Name:   lint.NewRuleName(123, "resource-name-components-alternate"),
+	OnlyIf: utils.IsResource,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		resource := utils.GetResource(m)
+		for _, p := range resource.GetPattern() {
+			components := strings.Split(p, "/")
+			for i, c := range components {
+				identifierExpected := i%2 == 1
+				if identifierExpected != isIdentifier(c) {
```
shwoodard, line 40: IMO this is less readable than,

```go
if identifierExpected && !isIdentifier(c) {...}
```

shwoodard: I guess they're not equivalent. You're checking for if there was an identifier where it was not expected, and my variant is slightly different. Disregard.

shwoodard (approved):

Issue #1104, opened 2023-03-01
AIP-121: require GET and LIST


toumorokoshi:
AIP-121 states: 

> A resource must support at minimum [Get](https://google.aip.dev/131): clients must be able to validate the state of resources after performing a mutation such as [Create](https://google.aip.dev/133), [Update](https://google.aip.dev/134), or [Delete](https://google.aip.dev/135).

> A resource must also support [List](https://google.aip.dev/132), except for [singleton resources](https://google.aip.dev/156) where more than one resource is not possible.

toumorokoshi:
I took a brief look at this one, and it's a bit tricky with the  current architecture. Some notes:

- the current method of binding the method to the resource is to retrieve the output type from the RPC itself. I wanted to bind this linter check to the message, but you can't easily look up the RPCs attached to it.
- I was thinking of just traversing the file, but that's also difficult because the file doesn't necessarily contain the RPCs that depend on it (it's common to have message declarations separate from service definitions).

toumorokoshi:
some more notes:

- we could look at the service, and RPCs defined in the same package.
- there are files that contain resources that are defined in the same proto file that are used by different services. 
- Service -> defined all RPCs, can use those to figure that.
  - look for Get / List for those.
  - collect all RPCs
- look at all return types? beyond standard method.
  - what is the value of a resource with only custom methods?
  - cloud build trigger could just have `:trigger` but must be gettable.

toumorokoshi:
more feedback: start with get, add list later (prevent false positives, especially while singletons are allowed).

noahdietz:
@toumorokoshi did you want to include this in CS-2312 since we skipped it in CS-2306?

toumorokoshi:
The major issue with linting for LIST is [Singleton resources](https://google.aip.dev/156). Until we have a way to determine which are singletons and which are not, we can't enforce it.

Maybe a good use case for `google.api.resource`?

noahdietz:
A _heuristic_ for identifying a singleton is if the resource pattern(s) end in a static segment (not an ID segment) e.g. `instances/{instance}/settings` where `/settings` being a static segment indicates it is a singleton.

I think this is pretty reliable.

toumorokoshi:
you're right, that could work. I guess there's no harm in adding a linter rule that checks for `List` except for resources with that heuristic.

Also we may want to consider putting this one in Cloud: I think *all* resources benefit, but the absolute need for List seems like a cloud-specific need.

[...]

PR #1152, opened 2023-05-13
feat(AIP-121): lint resource has a Get

toumorokoshi:
from AIP-121:

  A resource must support at minimum Get:
  clients must be able to validate the state of resources
  after performing a mutation such as Create, Update, or Delete.

Refactoring a few methods into a new util method, since it's best to have the same detection logic on Create/Get/Update/Delete/List methods.

Related: #1104

### DIFF
--- a/docs/rules/0121/resource-must-support-get.md
+++ b/docs/rules/0121/resource-must-support-get.md
@@ -49,6 +49,7 @@ service Foo {
 message Book {
   option (google.api.resource) = {
     type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
   };
 }
 ```
@@ -69,6 +70,7 @@ service Foo {
 message Book {
   option (google.api.resource) = {
     type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
   };
 }
 ```
--- a/docs/rules/0121/resource-must-support-get.md
+++ b/docs/rules/0121/resource-must-support-get.md
@@ -1,0 +1,79 @@
+---
+rule:
+  aip: 121
+  name: [core, '0121', resource-must-support-get]
+  summary: All resource names must use camel case in collection identifiers.
+permalink: /121/resource-must-support-get
+redirect_from:
+  - /0121/resource-must-support-get
+---
+
+# Resource must support get
+
+This rule enforces that all resources support the Get operation as mandated in
+[AIP-121][].
+
+## Details
+
+This rule scans a service for Create, Update, and List methods for resources,
+and ensures each one has a Get method.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+service Foo {
+  // Book has a create, but no Get method.
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+}
+
+message Book {
+	option (google.api.resource) = {
+		type: "library.googleapis.com/Book"
+	};
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+service Foo {
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+	rpc GetBook(GetBookRequest) returns (Book) {};
+}
+
+message Book {
+	option (google.api.resource) = {
+		type: "library.googleapis.com/Book"
+	};
+}
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the service.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+// (-- api-linter: core::0121::resource-must-support-get=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+service Foo {
+  // Book has a create, but no Get method.
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+}
+
+message Book {
+	option (google.api.resource) = {
+		type: "library.googleapis.com/Book"
+	};
+}
+```
+
+If you need to violate this rule for an entire file, place the comment at the
+top of the file.
+
+[aip-121]: https://aip.dev/121
+[aip.dev/not-precedent]: https://aip.dev/not-precedent
--- a/docs/rules/0121/resource-must-support-get.md
+++ b/docs/rules/0121/resource-must-support-get.md
@@ -30,9 +30,10 @@ service Foo {
 }
 
 message Book {
-	option (google.api.resource) = {
-		type: "library.googleapis.com/Book"
-	};
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
+  };
 }
 ```
 
@@ -42,13 +43,13 @@ message Book {
 // Correct.
 service Foo {
   rpc CreateBook(CreateBookRequest) returns (Book) {};
-	rpc GetBook(GetBookRequest) returns (Book) {};
+  rpc GetBook(GetBookRequest) returns (Book) {};
 }
 
 message Book {
-	option (google.api.resource) = {
-		type: "library.googleapis.com/Book"
-	};
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+  };
 }
 ```
 
@@ -66,9 +67,9 @@ service Foo {
 }
 
 message Book {
-	option (google.api.resource) = {
-		type: "library.googleapis.com/Book"
-	};
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+  };
 }
 ```
 
--- a/rules/aip0121/resource_must_support_get.go
+++ b/rules/aip0121/resource_must_support_get.go
@@ -57,7 +57,6 @@ var resourceMustSupportGet = &lint.ServiceRule{
 						"Missing Standard Get method for resource %q", t,
 					),
 					Descriptor: s,
-					Location:   s.GetSourceInfo(),
 				})
 			}
 		}
--- a/rules/aip0121/resource_must_support_get.go
+++ b/rules/aip0121/resource_must_support_get.go
@@ -1,0 +1,70 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aip0121
+
+import (
+	"fmt"
+
+	"bitbucket.org/creachadair/stringset"
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+// TODO: test against LROs
+var resourceMustSupportGet = &lint.ServiceRule{
+	Name: lint.NewRuleName(121, "resource-must-support-get"),
+	LintService: func(s *desc.ServiceDescriptor) []lint.Problem {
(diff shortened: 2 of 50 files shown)


docs/rules/0121/resource-must-support-get.md
```diff
@@ -0,0 +1,79 @@
...
+```proto
+// Incorrect.
+service Foo {
+  // Book has a create, but no Get method.
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+}
+
+message Book {
+	option (google.api.resource) = {
+		type: "library.googleapis.com/Book"
```
noahdietz, line 34: Can you include a `pattern` just to be safe :)

docs/rules/0121/resource-must-support-get.md
```diff
@@ -0,0 +1,79 @@
...
+	};
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+service Foo {
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
```
noahdietz, line 44: Nit: spacing

docs/rules/0121/resource-must-support-get.md
```diff
@@ -0,0 +1,79 @@
...
+service Foo {
+  // Book has a create, but no Get method.
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+}
+
+message Book {
+	option (google.api.resource) = {
+		type: "library.googleapis.com/Book"
+	};
+}
```
noahdietz, line 36: Nit: consistent spacing/tabbing please

rules/internal/utils/extension.go
```diff
@@ -65,6 +65,26 @@ func GetResponseType(m *desc.MethodDescriptor) *desc.MessageDescriptor {
...
+
+	if m.GetOutputType().GetName() != "Operation" {
+		return m.GetOutputType()
+	}
+
+	info := GetOperationInfo(m)
+	if info == nil {
+		return nil
+	}
+	typ := FindMessage(m.GetFile(), info.GetResponseType())
```
noahdietz, line 83: I would leverage the existing `GetResponseType` to do the LRO check/`operation_info` bit instead of hand rolling it. If it returns nil, its not an LRO.

rules/internal/utils/extension.go
```diff
@@ -65,6 +65,26 @@ func GetResponseType(m *desc.MethodDescriptor) *desc.MessageDescriptor {
 	return typ
 }
 
+// GetOutputOrLROResponseMessage returns the OutputType if the response is
+// not an LRO, or the ResponseType otherwise.
+func GetOutputOrLROResponseMessage(m *desc.MethodDescriptor) *desc.MessageDescriptor {
```
noahdietz, line 70: Maybe just `GetMethodResponseType`? I kind of feel like the existing `GetResponseType` should be renamed `GetOperationResponseType` and then we can use `GetResponseType` for the OutputOrOperation helper...perhaps that's another PR though.

rules/aip0121/resource_must_support_get.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		for t, m := range resourcesWithOtherMethods {
+			if !resourcesWithGet.Contains(t) {
+				problems = append(problems, lint.Problem{
+					Message: fmt.Sprintf(
+						"resource %q must support Get method", t,
+					),
+					Descriptor: m,
+					Location:   locations.MessageResource(m),
```
noahdietz, line 64: If the resource is not in the diff, but the service is, this finding will not show up.

I think we want these findings to appear on the `service` itself since it is a `ServiceRule` and the service is where it will be fixed.

rules/aip0121/resource_must_support_get.go
```diff
@@ -0,0 +1,70 @@
...
+					resourcesWithOtherMethods[t] = message
+				}
+			}
+		}
+
+		for t, m := range resourcesWithOtherMethods {
+			if !resourcesWithGet.Contains(t) {
+				problems = append(problems, lint.Problem{
+					Message: fmt.Sprintf(
+						"resource %q must support Get method", t,
```
noahdietz, line 61: In line with where the finding will be reported in CL, perhaps this message should be the following:
```suggestion
						"Missing Standard Get method for resource %q", t,
```

rules/aip0121/resource_must_support_get.go
toumorokoshi: ack! makes sense. I was trying to point someone in the right direction, but you're right - I forgot about tricorder's behavior not matching here.

docs/rules/0121/resource-must-support-get.md
toumorokoshi: sorry :( Not sure what's happening there.. I'll look at my formatter.

rules/aip0121/resource_must_support_get.go
toumorokoshi: Actually would that work? The service will likely be unchanged as well,
so marking that line won't show up either.

Maybe on the offending method would work? presumably the offending method would
have been added to have triggered the error.

rules/internal/utils/extension.go
toumorokoshi: I made the refactor since at least for my ide, it's a one-command change. 

I'll add it in the commit message too.

rules/internal/utils/extension.go
toumorokoshi: sounds good!

rules/aip0121/resource_must_support_get.go
noahdietz: From a general use PoV (not in Critique/a CL), where would you want to see a finding that says "Missing Standard Get RPC for resource Foo"? On the `service` that implements other RPCs related to that resource? On the resource message itself?

Then, from a Critique perspective, _when_ would you want to check for this? When a proto `service` for a resource is being authored/edited? When a resource `message` is being edited?

I think in all cases, I would personally prefer to see such a finding on/relating to the proto `service` that "looks like" it manages that resource. This is the most actionable place to emit such a finding. Emitting it in relation to the resource `message`, for example, would likely be received with confusion and "i can't fix that in this CL".

toumorokoshi: yes, sounds good, I'll fix that first. I'm just worried about critique *not* showing the error because the service name wasn't among the modified lines of the proto.

rules/aip0121/resource_must_support_get.go
```diff
@@ -0,0 +1,70 @@
...
+		resourcesWithOtherMethods := map[string]*desc.MessageDescriptor{}
+		// Iterate all RPCs and try to find resources. Mark the
+		// resources which have a Get method, and which ones do not.
+		for _, m := range s.GetMethods() {
+			if utils.IsGetMethod(m) {
+				t := utils.GetResource(m.GetOutputType()).GetType()
+				resourcesWithGet.Add(t)
+			} else if utils.IsCreateMethod(m) || utils.IsUpdateMethod(m) {
+				message := utils.GetResponseType(m)
+				if message != nil {
```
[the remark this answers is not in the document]
toumorokoshi: ack.

rules/aip0121/resource_must_support_get.go
```diff
@@ -0,0 +1,70 @@
...
+					t := utils.GetResource(message).GetType()
+					resourcesWithOtherMethods[t] = message
+				}
+			} else if utils.IsListMethod(m) {
+				message := utils.GetListResourceMessage(m)
+				if message != nil {
```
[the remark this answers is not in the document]
toumorokoshi: ack.

docs/rules/0121/resource-must-support-get.md
```diff
@@ -0,0 +1,80 @@
...
+```proto
+// Correct.
+service Foo {
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+  rpc GetBook(GetBookRequest) returns (Book) {};
+}
+
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
```
noahdietz, line 51: Missing `pattern`

docs/rules/0121/resource-must-support-get.md
```diff
@@ -0,0 +1,80 @@
...
+// (-- api-linter: core::0121::resource-must-support-get=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+service Foo {
+  // Book has a create, but no Get method.
+  rpc CreateBook(CreateBookRequest) returns (Book) {};
+}
+
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
```
noahdietz, line 71: Missing `pattern`

rules/aip0121/resource_must_support_get.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		for t := range resourcesWithOtherMethods {
+			if !resourcesWithGet.Contains(t) {
+				problems = append(problems, lint.Problem{
+					Message: fmt.Sprintf(
+						"Missing Standard Get method for resource %q", t,
+					),
+					Descriptor: s,
+					Location:   s.GetSourceInfo(),
```
noahdietz, line 60: `Location` isn't necessary when it is directly on the element e.g. the `service` descriptor. The Finding produced will default to relating to the `Descriptor`'s location. So this can be removed.

rules/aip0133/resource_reference_type.go
```diff
@@ -29,19 +29,19 @@ var resourceReferenceType = &lint.MethodRule{
 		// Return type of the RPC.
 		ot := m.GetOutputType()
 		if ot.GetName() == "Operation" {
-			ot = utils.GetResponseType(m)
+			ot = utils.GetOperationResponseType(m)
 		}
```
noahdietz, line 33: I realize this isn't the diff, but we need to clean this up, because this isn't a robust enough check. Compute and Cloud SQL, while not compliant so they might not have api-linter running, define their own `Operation` type and are not using LRO. This should be consolidated into the more robust `utils.GetResponseType(m)`. There are likely other spots, I think we need a fixit :)

rules/internal/utils/extension.go
```diff
@@ -65,6 +65,20 @@ func GetResponseType(m *desc.MethodDescriptor) *desc.MessageDescriptor {
...
+// GetResponseType returns the OutputType if the response is
+// not an LRO, or the ResponseType otherwise.
+func GetResponseType(m *desc.MethodDescriptor) *desc.MessageDescriptor {
+	if m == nil {
+		return nil
+	}
+
+	if m.GetOutputType().GetName() != "Operation" {
+		return m.GetOutputType()
+	}
```
noahdietz, line 77: As mentioned in another comment, this probably isn't the best check. "Operation" doesn't necessarily correlate to LRO. In the interest of being robust, let's check if the output type is a `google.longrunning.Operation`, which obviously requires checking the type's package and the name of the message itself.

rules/aip0133/resource_reference_type.go
toumorokoshi: ack. Here's a tracking issue: https://github.com/googleapis/api-linter/issues/1154

rules/internal/utils/extension.go
toumorokoshi: ack, I added the package check.

toumorokoshi:
Thanks @noahdietz! addressed and re-requested

rules/aip0133/resource_reference_type.go
```diff
@@ -26,12 +26,7 @@ import (
 var resourceReferenceType = &lint.MethodRule{
 	Name: lint.NewRuleName(133, "resource-reference-type"),
 	OnlyIf: func(m *desc.MethodDescriptor) bool {
-		// Return type of the RPC.
-		ot := m.GetOutputType()
-		if ot.GetName() == "Operation" {
-			ot = utils.GetOperationResponseType(m)
-		}
-
+		ot := utils.GetResponseType(m)
```
noahdietz, line 29: If you are doing this fix in this PR, it is also in rules/aip0233/resource_reference_type.go :)

noahdietz (approved):

rules/aip0133/resource_reference_type.go
toumorokoshi: done! fixed in a couple places.

PR #1156, opened 2023-05-24
feat: add request-required-fields for std methods

toumorokoshi:
As a follow-up to #1142, adding check for required fields for other standard methods. Each AIP has guidance on which fields are required vs not required.


noahdietz (commented):
Just a few nits!

docs/rules/0131/request-required-fields.md
```diff
@@ -0,0 +1,86 @@
...
+  // Format: publishers/{publisher}/books/{book}
+  string name = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (google.api.resource_reference) = {
+      type: "library.googleapis.com/Book"
+  }];
+
+  // (-- api-linter: core::0131::request-required-fields=disabled
+  //     aip.dev/not-precedent: We really need this field to be required because
+  // reasons. --)
```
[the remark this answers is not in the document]
toumorokoshi: oh weird, thanks! not sure what happened there.

docs/rules/0134/request-required-fields.md
```diff
@@ -0,0 +1,65 @@
...
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message UpdateBookRequest {
+  Book book = 1 [(google.api.field_behavior) = REQUIRED];
+  // Non-standard required field.
+  google.protobuf.FieldMask update_mask = 2 [(google.api.field_behavior) = REQUIRED];
```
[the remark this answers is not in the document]
toumorokoshi: oops sorry, fixed.

rules/aip0131/request_required_fields.go
```diff
@@ -0,0 +1,51 @@
...
+
+// The get request message should not have unrecognized fields.
+var requestRequiredFields = &lint.MessageRule{
+	Name:   lint.NewRuleName(131, "request-required-fields"),
+	OnlyIf: isGetRequestMessage,
+	LintMessage: func(m *desc.MessageDescriptor) (problems []lint.Problem) {
+		// Rule check: Establish that there are no unexpected fields.
+		allowedRequiredFields := map[string]*builder.FieldType{
+			"name": nil,
+		}
```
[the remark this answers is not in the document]
toumorokoshi: ack, will fix.

toumorokoshi (commented):
Addressed! thanks, PTAL.

rules/aip0133/request_required_fields.go
```diff
@@ -33,10 +33,10 @@ var requestRequiredFields = &lint.MessageRule{
 		resourceMsgName := getResourceMsgNameFromReq(m)
 
 		// Rule check: Establish that there are no unexpected fields.
-		allowedRequiredFields := map[string]*builder.FieldType{
-			"parent": nil, // AIP-133
-			fmt.Sprintf("%s_id", strings.ToLower(strcase.SnakeCase(resourceMsgName))): nil,
-		}
+		allowedRequiredFields := stringset.New(
+			"parent",
+			fmt.Sprintf("%s_id", strings.ToLower(strcase.SnakeCase(resourceMsgName))),
```
noahdietz, line 38: I would use the split type name (e.g. the `{Type Name}` portion of `type: "{Service Name}/{Type Name}"`), not the message name. We have a few resources that do not match the message name, as odd as that is, and it is not a requirement that they match.

Maybe even future proof this by checking the `singular` field first, then falling back on name parsing?

Sorry I missed this in my first go-around 🙈

toumorokoshi: overall agreed, I wasn't super happy with how the message name was retrieved
originally, I just was opting for minimal-possible-change vs rewriting the
world.

But since you brought it up, I'll see what I can do here. agreed that we should
get the resource name from the singular, then type.

toumorokoshi: ah, so one issue with this is how AIP-134 recommends the field name being the
singular of the resource.

If we had a standard field (e.g. resource) so we *know* that the field
represents the resource, it's much easier to grab that field and grab the
singular type. 

Actually, I think I can do this by:

1. making this a methodrule, and matching on "Create".
2. grabbing the output type, which *must* be the resource.
3. using that annotation to get the intended prefix name.

Okay, I'll try that.

noahdietz: > If we had a standard field (e.g. resource) so we know that the field
represents the resource, i

Yeah I mean you can look up the resource message field in the same request message, there must only be one, so you wouldn't have to make it a `MethodRule`, but I see how going via response type would make it more fool proof.

Make sure to use the `utils.GetResponseType` that handles LROs as well :)

toumorokoshi: I might as well - I do think there's some more refactoring to be done here, but
I'll do that incrementally and after the cloud standards when I have a few more
cycles. 

Addressed! PTAL.

rules/aip0133/request_required_fields.go
```diff
@@ -26,19 +26,25 @@ import (
...
-	LintMessage: func(m *desc.MessageDescriptor) (problems []lint.Problem) {
-		resourceMsgName := getResourceMsgNameFromReq(m)
+	OnlyIf: utils.IsCreateMethod,
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		ot := utils.GetResponseType(m)
+		r := utils.GetResource(ot)
+		resourceMsgName := ""
+		if r != nil {
+			resourceMsgName = utils.GetResourceSingular(r)
+		}
```
noahdietz, line 38: I would just handle `nil` in `GetResourceSingular`, but this is fine for now.

noahdietz (approved):

rules/aip0133/request_required_fields.go
toumorokoshi: done!

PR #1158, opened 2023-05-26
fix(AIP-158): reference resource plural for field name

noahdietz:
As we start to require `plural` in `google.api.resource`, we should referencing it in rules that attempt to pluralize words for things like field names or collection IDs. This is a reasonable heuristic to use for the AIP-158 `response-plural-first-field` when that first field is a resource.

Fixes http://b/271479822

### DIFF
--- a/rules/aip0158/response_plural_first_field.go
+++ b/rules/aip0158/response_plural_first_field.go
@@ -19,6 +19,7 @@ import (
 	"github.com/googleapis/api-linter/locations"
 	"github.com/googleapis/api-linter/rules/internal/utils"
 	"github.com/jhump/protoreflect/desc"
+	"github.com/stoewer/go-strcase"
 )
 
 var responsePluralFirstField = &lint.MessageRule{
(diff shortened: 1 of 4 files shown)


rules/aip0158/response_plural_first_field.go
```diff
@@ -30,7 +31,16 @@ var responsePluralFirstField = &lint.MessageRule{
...
 		firstField := m.GetFields()[0]
-		want := utils.ToPlural(firstField.GetName())
+
+		// If the field is a resource, use the `plural` annotation to decide the
+		// appropriate plural field name.
+		var want string
+		if res := utils.GetResource(firstField.GetMessageType()); res != nil && res.GetPlural() != "" {
+			want = strcase.SnakeCase(res.GetPlural())
+		} else {
+			want = utils.ToPlural(firstField.GetName())
```
toumorokoshi, line 41: this already existed of course, but I feel like pluralizing the resource singular or the type would be more robust than the field name.

rules/aip0158/response_plural_first_field.go
```diff
@@ -30,7 +31,16 @@ var responsePluralFirstField = &lint.MessageRule{
 		// Throw a linter warning if, the first field in the message is not named
 		// according to plural(message_name.to_snake().split('_')[1:-1]).
 		firstField := m.GetFields()[0]
-		want := utils.ToPlural(firstField.GetName())
+
+		// If the field is a resource, use the `plural` annotation to decide the
+		// appropriate plural field name.
+		var want string
+		if res := utils.GetResource(firstField.GetMessageType()); res != nil && res.GetPlural() != "" {
```
toumorokoshi, line 38: nit: make a util function similar to GetResourceSingular? https://github.com/googleapis/api-linter/blob/ececf6c459201eccbaf506e57d0f735c62ba7aa0/rules/internal/utils/resource.go#L25.

toumorokoshi (approved):

rules/aip0158/response_plural_first_field.go
noahdietz: I think this is meant to support the scenario where the paginated field isn't a resource. I agree with you, but I don't know if we should change it right now. There are maybe one or two Standard List-like fields that paginate a `repeated string` field that I know of. This rule being in AIP-158, not AIP-132 is poignant.

rules/aip0158/response_plural_first_field.go
noahdietz: Good call! Done.

rules/aip0158/response_plural_first_field.go
noahdietz: Opened #1160

PR #1165, opened 2023-06-01
fix(AIP-133): compare types instead for message fields

noahdietz:
Change the filter for the resource message field to compare the field type to the response type (which should be the resource for a Standard Create), instead of a case sensitive derivation of the resource type name or singular name.

Add some tests for when `singular` is set.

Should be a fix for item 2 of http://yaqs/4607975166683643904 (internal link)

### DIFF
--- a/rules/aip0133/request_required_fields.go
+++ b/rules/aip0133/request_required_fields.go
@@ -32,9 +32,7 @@ var requestRequiredFields = &lint.MethodRule{
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
 		ot := utils.GetResponseType(m)
 		r := utils.GetResource(ot)
-		// The `singular` will be lowerCamelCase, while the Type Name will be
-		// UpperCamelCase. For simplicity, upper-case it always.
-		resourceMsgName := strcase.UpperCamelCase(utils.GetResourceSingular(r))
+		resourceMsgName := utils.GetResourceSingular(r)
 
 		// Rule check: Establish that there are no unexpected fields.
 		allowedRequiredFields := stringset.New(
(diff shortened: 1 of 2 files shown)


rules/aip0133/request_required_fields.go
```diff
@@ -45,8 +45,9 @@ var requestRequiredFields = &lint.MethodRule{
 			if !utils.GetFieldBehavior(f).Contains("REQUIRED") {
 				continue
 			}
-			// Skip the check with the field that is the body.
-			if t := f.GetMessageType(); t != nil && t.GetName() == resourceMsgName {
+			// Skip the check with the field that is the resource, which for
+			// Standard Create, is the output type.
+			if t := f.GetMessageType(); t != nil && t.GetName() == ot.GetName() {
```
toumorokoshi, line 50: FYI 134/request_required_fields has the same issue.

And I think there may be a slightly more robust way to do this -> check the message type rather than the name. This is pretty easily done by omitting an input field that has the same type as the output field.  

It gets rid of the verifying that the name field matches the resource singular, but that's done in a separate linter rule anyway.

toumorokoshi (changes requested):
Wow nice! I was in the middle of implementing this, but you beat me to it :) I owe you.

Quick suggestion on a more robust check, as well as a request to fix 134/request_required_fields if you could.

rules/aip0133/request_required_fields.go
noahdietz: I *think* that's what this check does already. It compares the request field's message type name to the name of the output type (which should be the resource for Standard Create).

Am I misunderstanding your comment?

noahdietz: I'll do 0134 in a separate PR if you don't mind

toumorokoshi: I was suggesting direct pointer checks vs name checks (to avoid the same message). But in practice yes I think it'll work fine.

toumorokoshi (approved):

rules/aip0133/request_required_fields.go
toumorokoshi: I'm fine with it in a separate PR. thanks!

noahdietz: > I was suggesting direct pointer checks vs name checks (to avoid the same message). But in practice yes I think it'll work fine.

Oh yeah I see...yeah Idk I'm scared by that haha

PR #1166, opened 2023-06-01
fix(AIP-203): ignore behavior in internal comments

noahdietz:
All of the field behavior rules that suggest the use of the annotation over documenting it in a comment should ignore internal comment contents.

Refactors `separateInternaComments` into a utility API.

Should be a fix for item 3 in http://yaqs/4607975166683643904 (internal link).

### DIFF
--- a/rules/aip0192/aip0192.go
+++ b/rules/aip0192/aip0192.go
@@ -16,8 +16,6 @@
 package aip0192
 
 import (
-	"strings"
-
 	"github.com/googleapis/api-linter/lint"
 	"github.com/jhump/protoreflect/desc"
 )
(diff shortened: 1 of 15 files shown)


rules/aip0192/aip0192.go
```diff
@@ -48,40 +46,3 @@ func isDeprecated(d desc.Descriptor) bool {
 		return false
 	}
 }
-
-func separateInternalComments(comments ...string) struct {
```
toumorokoshi, line 52: I think the file that contains this function now in utils/ wasn't added? But otherwise LGTM.

toumorokoshi (commented):
Woo, thank you! Looks good, just missing a file I think.

toumorokoshi (approved):
LGTM, thanks!

PR #1168, opened 2023-06-01
fix(134): compare message types for required fields

noahdietz:
Refactors `core::0134::request-required-fields` into a `MethodRule` so as to get access to the expected response type a.k.a the exact resource type being updated. Then fix the message type field check to compare the Resource message type to the field message type if applicable. Similar to #1165


toumorokoshi (approved):
thanks!

PR #1170, opened 2023-06-01
feat: resource-names-alternating-components check for AIP-122

alexander-fenster:
Adding a rule to check that resource names components alternate between `collection_name` and `{resource_id}` as described in [AIP-122](https://google.aip.dev/122).


alexander-fenster:
Oops, already covered by #1151.

PR #1172, opened 2023-06-01
feat: test across multiple files, ignore name on resource

shwoodard:
This commit tests that the logic works across multiple files and also lints response messages that are resources or of depth greater than 0.

### DIFF
--- a/rules/aip0203/field_behavior_required.go
+++ b/rules/aip0203/field_behavior_required.go
@@ -30,31 +30,8 @@ var fbs = stringset.New(
 var fieldBehaviorRequired = &lint.MethodRule{
 	Name: lint.NewRuleName(203, "field-behavior-required"),
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
-		var ps []lint.Problem
-
 		req := m.GetInputType()
-		resp := m.GetOutputType()
-
-		reqProblems := problems(req, 0, 0)
-
-		var respProblems map[string][]lint.Problem
-		if r := utils.GetResource(resp); r != nil {
-			respProblems = problems(resp, 0, 0)
-		} else {
-			respProblems = problems(resp, 1, 0)
-		}
-
-		for _, p := range reqProblems {
-			ps = append(ps, p...)
-		}
-
-		for msgType, p := range respProblems {
-			// Avoid dups
-			if _, ok := reqProblems[msgType]; !ok {
-				ps = append(ps, p...)
-			}
-		}
-
+		ps := problems(req)
 		if len(ps) == 0 {
 			return nil
 		}
@@ -63,24 +40,21 @@ var fieldBehaviorRequired = &lint.MethodRule{
 	},
 }
 
-func problems(m *desc.MessageDescriptor, minDepth, currDepth int) map[string][]lint.Problem {
-	ps := make(map[string][]lint.Problem)
+func problems(m *desc.MessageDescriptor) []lint.Problem {
+	var ps []lint.Problem
 
 	for _, f := range m.GetFields() {
-		mt := f.GetMessageType()
-
-		if minDepth <= currDepth {
-			p := checkFieldBehavior(f)
-			if p != nil {
-				name := m.GetFullyQualifiedName()
-				ps[name] = append(ps[name], *p)
-			}
+		if r := utils.GetResource(m); r != nil && f.GetName() == "name" {
+			continue
 		}
 
-		if mt != nil {
-			for name, p := range problems(mt, minDepth, currDepth+1) {
-				ps[name] = append(ps[name], p...)
-			}
+		p := checkFieldBehavior(f)
+		if p != nil {
+			ps = append(ps, *p)
+		}
+
+		if mt := f.GetMessageType(); mt != nil {
+			ps = append(ps, problems(mt)...)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rules/aip0203/field_behavior_required.go
```diff
@@ -23,47 +23,62 @@ import (
...
-func checkFields(m *desc.MessageDescriptor) []lint.Problem {
-	problems := []lint.Problem{}
+func problems(m *desc.MessageDescriptor) []lint.Problem {
+	var ps []lint.Problem
+
 	for _, f := range m.GetFields() {
-		asMessage := f.GetMessageType()
-		if asMessage != nil {
-			problems = append(problems, checkFields(asMessage)...)
+		if r := utils.GetResource(m); r != nil && f.GetName() == "name" {
```
noahdietz, line 47:
```suggestion
		if utils.IsResource(m) && f.GetName() == "name" {
```

rules/aip0203/field_behavior_required_test.go
```diff
@@ -131,12 +201,90 @@ func TestFieldBehaviorRequiredNested(t *testing.T) {
...
+			f := ds["service.proto"]
+			it := f.GetServices()[0].GetMethods()[0].GetInputType()
+			fd := it.GetFields()[0].GetMessageType().GetFields()[0]
+
+			if diff := tc.problems.SetDescriptor(fd).Diff(fieldBehaviorRequired.Lint(f)); diff != "" {
+				t.Errorf(diff)
+			}
+
+			want := "resource.proto"
+			if got := fd.GetFile().GetName(); got != want {
```
noahdietz, line 285: Sorry, how does this actually check that the `Descriptor` included in the `Problem` is from the file not being directly linted? I think this is just testing that the template files were parsed correctly?

rules/aip0203/field_behavior_required.go
```diff
@@ -23,47 +23,62 @@ import (
...
 var fieldBehaviorRequired = &lint.MethodRule{
 	Name: lint.NewRuleName(203, "field-behavior-required"),
 	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
-		// we only check requests, as OutputTypes are always
-		// OUTPUT_ONLY
-		it := m.GetInputType()
-		return checkFields(it)
+		req := m.GetInputType()
+		ps := problems(req)
+		if len(ps) == 0 {
```
toumorokoshi, line 35: this could be returned directly too: 0 problems wouldn't add any.

rules/aip0203/field_behavior_required.go
```diff
@@ -23,47 +23,62 @@ import (
...
+			ps = append(ps, problems(mt)...)
 		}
-		problems = append(problems, checkFieldBehavior(f)...)
 	}
-	return problems
+
+	return ps
 }
 
-func checkFieldBehavior(f *desc.FieldDescriptor) []lint.Problem {
```
toumorokoshi, line 52: nit: this was an intentional choice to ensure that one could easily attach new errors, without the refactor to a list.

But i'm not strongly opinionated if you change it.

rules/aip0203/field_behavior_required.go
```diff
@@ -23,47 +23,62 @@ import (
 	"github.com/jhump/protoreflect/desc"
 )
 
-var minimumRequiredFieldBehavior = stringset.New(
+var fbs = stringset.New(
```
toumorokoshi, line 26: fbs doesn't represent all field behaviors (e.g. unordered_list). The naming of this variable was intentional to indicate the required fbs.

If the variable indicates that it's only the required set, I'd be fine with shortening the name.

rules/aip0203/field_behavior_required.go
```diff
@@ -23,47 +23,62 @@ import (
...
+	}
+
+	if !fbs.Intersects(fb) {
 		// check for at least one valid annotation
-	} else if !minimumRequiredFieldBehavior.Intersects(fieldBehavior) {
-		problems = append(problems, lint.Problem{
+		return &lint.Problem{
 			Message: fmt.Sprintf(
-				"google.api.field_behavior must have at least one of the following behaviors set: %v", minimumRequiredFieldBehavior),
+				"google.api.field_behavior must contain at least one, \"%v\"", fbs),
```
toumorokoshi, line 78: could you ad the field name here as well? this'll help in case the linter finding gets lost again like last time.

toumorokoshi (changes requested):
Thanks! a couples changes would be great.

rules/aip0203/field_behavior_required_test.go
shwoodard: The file being linted is "service.proto", the field in the problem is in "resource.proto".

rules/aip0203/field_behavior_required.go
shwoodard: All the other rules return nil for no problems.

rules/aip0203/field_behavior_required.go
shwoodard: I think it's implied by the nature of the rule and the docs in the md, but I don't feel strongly.

rules/aip0203/field_behavior_required.go
shwoodard: Done.

rules/aip0203/field_behavior_required.go
toumorokoshi: I'm not strongly opinionated enough to block this change, but FWIW the term "fbs" requires me to go read the code to understand how the variable is actually used.

It's also kind of a misnomer: "fbs" could mean easily be interpreted as *all* field behaviors, not just required ones.

toumorokoshi (approved):
Thanks!

PR #1173, opened 2023-06-05
feat(AIP192): align trailing comment check with protobuf

noahdietz:
Changes the test for `core::0192::only-leading-comments` to check put "trailing" comments in the place that protobuf actually parses them from. This is the issue that blocked the adoption of `jhump/protoreflect` v1.10.2 because of the following parsing change:

> [jhump/protoreflect] used the comment immediately following the closing brace (}) as the element's trailing comment. However, protoc uses the comment immediately following the opening brace ({) as the trailing comment. The behavior in this package has been changed to match that of protoc.

This may end up getting changed at some point as there is an open issue for it: https://github.com/protocolbuffers/protobuf/issues/12921. But for now, in order to bring this package up to date, we need to adjust the check. FWIW, public google apis should never have trailing comments anyways, so this isn't as big of an issue.

Fixes a few other test errors created by new expectations introduced by newer versions of `jhump/protoreflect`.


toumorokoshi (approved):
Thanks for aligning our comment parsing behavior with protoc!

PR #1182, opened 2023-06-26
fix(AIP-203): remove optional conflict

toumorokoshi:
The AIPs now allow OPTIONAL to be used with other descriptive
annotations (e.g. INPUT_ONLY or IMMUTABLE). Removing the rule
accordingly to minimize confusion.


jskeet (approved):
I think it's fine to remove the rule for now - in time we may want to prohibit specific combinations, e.g. REQUIRED and OPTIONAL, but that can come later.

PR #1200, opened 2023-07-17
fix(AIP-203): handle recursive field behavior check

toumorokoshi:
Some messages can be recursive, which lead to stackoverflows.


noahdietz:
already merged in #1199

PR #1212, opened 2023-07-28
fix: refactor Standard Method message helpers into utils

noahdietz:
All Standard Method request message `Is{Method}(Request|Response)Message` helpers moved into to `utils` and exported for other rules to use. This is also necessary for some internal rules.

Some already exported helpers in `aip0162` are left behind, marked deprecated, and made to be passthroughs to the migrated methods. rules should not be importing eachother, it was probably not a good idea to export a function from a rule group package.

### DIFF
--- a/rules/aip0132/response_unknown_fields.go
+++ b/rules/aip0132/response_unknown_fields.go
@@ -15,6 +15,7 @@
 package aip0132
 
 import (
+	"regexp"
 	"strings"
 
 	"bitbucket.org/creachadair/stringset"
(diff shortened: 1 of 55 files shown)


rules/aip0132/response_unknown_fields.go
```diff
@@ -33,10 +34,15 @@ var respAllowedFields = stringset.New(
 	"unreachable_locations", // Wrong, but a separate AIP-217 rule catches it.
 )
 
+// TODO: Refactor the use of this regexp in this rule to a more robust solution.
+//
+// Deprecated: Do not use this.
+var listRespMessageRegexp = regexp.MustCompile("^List([A-Za-z0-9]*)Response$")
```
andrei-scripniciuc, line 40: since this regexp is only used to extract the resource name from an AIP 132 or AIP 162 list response message, perhaps just extract that into a method instead. you can also change the regex to trim _revisions if needed. I assume you can then leave the new regex as non-deprecated since it serves a distinct and valid use case.

noahdietz: Good call, let's do this. I do think that this is kind of a roundabout way of getting the expected field name, but it has worked thus far so let's just make this case better, and perhaps change the logic later on.

Done.

noahdietz: Note: I added a `message_test.go` file just for this one new method because it has actual logic, unlike the regex wrapper methods.

andrei-scripniciuc (approved):

PR #1238, opened 2023-08-16
feat(AIP-121): lint for mutable reference cycles

### LINKED ISSUE #1109: AIP-121: lint against cyclic references
AIP-121 states:

> The relationship between resources, such as parent-child or [resource references](https://google.aip.dev/122#fields-representing-another-resource), must be representable via a [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph).

That should be added.

noahdietz:
Introduces rules to lint for mutable resource reference cycles.

Fixes #1109


rules/aip0121/no_cycles.go
```diff
@@ -0,0 +1,76 @@
...
+			// Skip unresolvable resource references, or those that don't resolve
+			// to a message e.g. file-level google.api.resource_defintion.
+			if resMsg == nil {
+				continue
+			}
+			references[field] = resMsg
+		}
+
+		origRes := utils.GetResource(m)
+		for origRef, refMsg := range references {
```
toumorokoshi, line 52: this logic will detect single-deep references, but it won't detect any cycles deeper than that (e.g. a cycle of A -> B -> C -> A). I don't think there are practically many of these examples, but still something worth catching.

Could you rewrite the algorithm in a recursive fashion? you can keep a set of seen resources and pass that along to each call, then raise a problem when you find a resource that you've encountered before ( starting wit the original resource).

docs/rules/0121/no-cycles.md
```diff
@@ -0,0 +1,129 @@
...
+  name: [core, '0121', no-cycles]
+  summary: Resources must not form a resource reference cycle.
+permalink: /121/no-cycles
+redirect_from:
+  - /0121/no-cycles
+---
+
+# Resource must support get
+
+This rule enforces that resources do not create reference cycles as mandated in
```
toumorokoshi, line 13:
```suggestion
This rule enforces that resources do not create cycles of mutable references as mandated in
```

toumorokoshi: slightly clearer wording, wdyt?

docs/rules/0121/no-cycles.md
```diff
@@ -0,0 +1,129 @@
...
+[AIP-121][].
+
+## Details
+
+This rule scans the fields of every resource and ensures that any references to
+other resources do not create a cycle between them.
```
toumorokoshi, line 19:
```suggestion
other resources do not create a mutable cycle between them.
```

docs/rules/0121/no-cycles.md
```diff
@@ -0,0 +1,129 @@
+---
+rule:
+  aip: 121
+  name: [core, '0121', no-cycles]
+  summary: Resources must not form a resource reference cycle.
+permalink: /121/no-cycles
```
toumorokoshi, line 6: how about naming the rule `no-mutable-cycles`? it might be clearer. 

At least in conversations with folks there's always a question on whether output-only cycles are allowed.

toumorokoshi (changes requested):
Thanks! mostly minor nits. Main comment is on whether we can detect deeper cycles.

docs/rules/0121/no-cycles.md
noahdietz: Yeah for sure. Done in a separate commit.

docs/rules/0121/no-cycles.md
noahdietz: Done

docs/rules/0121/no-cycles.md
noahdietz: For sure, I can do this, just after all other comments are resolved as I have to rename everything.

rules/aip0121/no_cycles.go
noahdietz: Yep, I sure can.

noahdietz: Done! PTAL

docs/rules/0121/no-cycles.md
noahdietz: Done.

rules/aip0121/no_mutable_cycles.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+var noMutableCycles = &lint.MessageRule{
+	Name:   lint.NewRuleName(121, "no-mutable-cycles"),
+	OnlyIf: utils.IsResource,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		res := utils.GetResource(m)
+
+		for _, field := range m.GetFields() {
```
toumorokoshi, line 33: nit: you could refactor this whole loop into findCycle, I think.

It looks like this logic exists just to help get the original resource reference. But there's nothing special about the original resource. You could just return whatever field is involved in the cycle, and as long as your problem enumerates all the seen resources, you could fix one of those chains in the cycle.

so basically if you get rid of orig and just use seen, you can get similar behavior.

rules/aip0121/no_mutable_cycles.go
```diff
@@ -0,0 +1,90 @@
...
+			continue
+		}
+		ref := utils.GetResourceReference(f)
+		// Focus on direct type references.
+		if ref.GetChildType() != "" {
+			continue
+		}
+		if ref.GetType() == orig.GetType() {
+			return &lint.Problem{
+				Message: "mutable resource reference introduces a resource reference cycle",
```
toumorokoshi, line 74: since you're collecting seen nodes anyway, can you print their google.api.resource.type? that would be super helpful for debugging.

toumorokoshi (approved):
LGTM! A couple suggestions but nothing blocking.

This is really cool, thanks for adding this.

rules/aip0121/no_mutable_cycles.go
noahdietz: Great idea! done

rules/aip0121/no_mutable_cycles.go
noahdietz: Yeah, let me stew on this a bit, trying to remember why I didn't push this into `findCycle` in the first place.

noahdietz: Ok fixed it! Done!

It was just a tad tricky because we wanted to ensure that the original message's field that started the chain ultimately ended up as the descriptor/location on the finding, but we want to lint all of the fields in the original resource and only the one that creates a cycle in the recursive call. Figured it out though and added a test that involves multiple cycles on one resource across multiple fields, one deep and one shallow.

PR #1241, opened 2023-08-24
feat(AIP-203): add resource name IDENTIFIER enforcement

noahdietz:
Enforces AIP-203 guidance added in https://github.com/aip-dev/google.aip.dev/pull/1201 that the resource's name field must have `google.api.field_behavior = IDENTIFIER`.

### DIFF
--- a/docs/rules/0203/resource-name-identifier.md
+++ b/docs/rules/0203/resource-name-identifier.md
@@ -1,0 +1,71 @@
+---
+rule:
+  aip: 203
+  name: [core, '0203', resource-name-identifier]
+  summary: Resource name field must have field behavior IDENTIFIER.
+permalink: /203/resource-name-identifier
+redirect_from:
+  - /0203/resource-name-identifier
+---
+
+# Resource name: IDENTIFIER
+
+This rule enforces that the field on a resource representing the resource's name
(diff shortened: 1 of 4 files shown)


toumorokoshi (approved):

sam-utila:
This breaks the lint for me.
I have an identifier, but it still fails on linting.

noahdietz:
@sam-utila can you open an issue instead of commenting on the PR? Also please include the proto that is failing.

worth noting this hasn't been released yet so you are using `main` which is probably not advisable.

sam-utila:
No problem, considering I can always put new linting rules on `disabled`.

Just wondering that sometimes `IDENTIFIER`s can also be `OUTPUT_ONLY` and this is something that is now not encouraged by this AIP - haven't sinked in yet :-) .

flyfire2002:
I too am interested in the implication for `name` fields currently marked as `OUTPUT_ONLY`. I guess semantically `IDENTIFIER` doesn't preclude a behavior that is `OUTPUT_ONLY` ("Server ignores it if set as input"). However now we don't have a language for https://google.aip.dev/203#output-only scenario 3 where the name is the immutable field.

PR #1245, opened 2023-08-29
feat(AIP-122): disallow embedded resource

noahdietz:
Disallow embedding resources in other resources, use resource_references instead.

Enforce guidance added in https://github.com/aip-dev/google.aip.dev/pull/1196.

### DIFF
--- a/docs/rules/0122/embedded-resource.md
+++ b/docs/rules/0122/embedded-resource.md
@@ -1,0 +1,107 @@
+---
+rule:
+  aip: 122
+  name: [core, '0122', embedded-resource]
+  summary: Resource references should not be embedded resources.
+permalink: /122/embedded-resource
+redirect_from:
+  - /0122/embedded-resource
+---
+
+# Resource reference type
+
+This rule enforces that references to resource are via
+`google.api.resource_reference`, not by embedding the resource message, as
+mandated in [AIP-122][].
+
+## Details
+
+This rule complains if it sees a resource field of type `message` that is also
+annotated as a `google.api.resource`.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
+  };
+  string name = 1;
+
+  // Incorrect. Resource references should not be embedded resource messages.
+  Author author = 2;
+}
+
+message Author {
+    option (google.api.resource) = {
+        type: "library.googleapis.com/Author"
+        pattern: "authors/{author}"
+    };
+
+    string name = 1;
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
+  };
+  string name = 1;
+
+  string author = 2 [(google.api.resource_reference) = {
+    type: "library.googleapis.com/Author"
+  }];
+}
+
+message Author {
+    option (google.api.resource) = {
+        type: "library.googleapis.com/Author"
+        pattern: "authors/{author}"
+    };
+
+    string name = 1;
+}
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the method.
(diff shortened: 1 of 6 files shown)


docs/rules/0122/embedded-resource.md
```diff
@@ -0,0 +1,107 @@
...
+permalink: /122/embedded-resource
+redirect_from:
+  - /0122/embedded-resource
+---
+
+# Resource reference type
+
+This rule enforces that references to resource are via
+`google.api.resource_reference`, not by embedding the resource message, as
+mandated in [AIP-122][].
```
toumorokoshi, line 15: is it worth mentioning there are exceptions? just thinking about the corpeng folks who like this feature running into this all the time.

we could just disable the rule in the corpeng directory too.

rules/aip0122/embedded_resource_test.go
```diff
@@ -0,0 +1,106 @@
...
+				string name = 1;
+			}
+		`, test)
+			m := f.FindMessage("Book")
+
+			want := test.problems
+			if len(want) > 0 {
+				want[0].Descriptor = m.FindFieldByName("library")
+			}
+			if len(want) == 2 {
```
toumorokoshi, line 98: nit: make this > 1, to be similar to the pattern above?

this also makes this line more resilient in case there's a three-problem test case. but not something difficult to fix when it comes up, of course.

docs/rules/0122/embedded-resource.md
```diff
@@ -0,0 +1,107 @@
...
+    };
+
+    string name = 1;
+}
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the method.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
```
toumorokoshi, line 78: this might be a good place to link to notable exceptions in the AIPs.

toumorokoshi (approved):

rules/aip0122/embedded_resource.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+var embeddedResource = &lint.MessageRule{
+	Name:   lint.NewRuleName(122, "embedded-resource"),
+	OnlyIf: utils.IsResource,
+	LintMessage: func(m *desc.MessageDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		for _, f := range m.GetFields() {
+			mt := f.GetMessageType()
+			if mt == nil || !utils.IsResource(mt) {
```
toumorokoshi, line 32: before I forget: can you add the check for a revision resource? I think using the message type may be sufficient. 

This probably isn't the right place to put the guard condition (maybe updating OnlyIf?)

toumorokoshi (changes requested):
Sorry, forgot that we need to update this to exclude revisions.

docs/rules/0122/embedded-resource.md
noahdietz: I don't think we need to enumerate exceptions here since the AIP will do so, that's what the "as mandated" is for. Probably best not to duplicate every edge case to guidance in the linter documentation.

I will make sure corpeng get a heads up to disable this rule before it gets into prod.

rules/aip0122/embedded_resource.go
noahdietz: I first considered the `OnlyIf`, but thought we should be more principled and only exempt valid resource snapshots in revision resources, but not allow another random resource to be embedded in the revision resource. Working on that now.

noahdietz: Done!

toumorokoshi (approved):
Great, thanks for doing that!

I think this can be merged even without merging https://github.com/aip-dev/google.aip.dev/pull/1192: no one is using the revisions pattern yet anyway, so the existence of the condition wouldn't impact it's usage as a check now.

noahdietz:
> Great, thanks for doing that!
> 
> I think this can be merged even without merging [aip-dev/google.aip.dev#1192](https://github.com/aip-dev/google.aip.dev/pull/1192): no one is using the revisions pattern yet anyway, so the existence of the condition wouldn't impact it's usage as a check now.

I agree! Good idea. :)

PR #1250, opened 2023-09-11
fix(AIP-121): ignore standard method lookalikes

noahdietz:
Ignore methods that look like standard methods but don't actually operate on a resource as a standard method would. This prevents false positive findings related for messages that aren't actually resources but have some standard method lookalikes. Similar to #1235

### DIFF
--- a/rules/aip0121/resource_must_support_get.go
+++ b/rules/aip0121/resource_must_support_get.go
@@ -33,16 +33,16 @@ var resourceMustSupportGet = &lint.ServiceRule{
 		// Iterate all RPCs and try to find resources. Mark the
 		// resources which have a Get method, and which ones do not.
 		for _, m := range s.GetMethods() {
-			if utils.IsGetMethod(m) {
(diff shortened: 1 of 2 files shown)


rules/aip0121/resource_must_support_get.go
```diff
@@ -33,16 +33,16 @@ var resourceMustSupportGet = &lint.ServiceRule{
 		// Iterate all RPCs and try to find resources. Mark the
 		// resources which have a Get method, and which ones do not.
 		for _, m := range s.GetMethods() {
-			if utils.IsGetMethod(m) {
+			if utils.IsGetMethod(m) && utils.IsResource(utils.GetResponseType(m)) {
```
jskeet, line 36: This presumably won't don't oddities like streaming requests or responses, but it's tighter than it was.

jskeet (approved):

rules/aip0121/resource_must_support_get.go
noahdietz: That's a good point, it won't. I might come back for that though, same goes for the Must have List rule as well.

PR #1266, opened 2023-10-10
chore(AIP-121): fix title for no-mutable-cycles doc

faube:
Previous title was for resource-must-support-get
Use the rule summary as the new page title.


noahdietz (approved):
Thanks, copy paste strikes again :)

PR #1270, opened 2023-10-13
feat(AIP-148): add uid-format rule

noahdietz:
As mandated in AIP-148, the `uid` field **must** have `(google.api.field_info).format = UUID4`. This rule enforces that.

Also fix incorrect doc title for `use-uid` rule doc.

### DIFF
--- a/docs/rules/0148/uid-format.md
+++ b/docs/rules/0148/uid-format.md
@@ -58,8 +58,6 @@ enclosing message. Remember to also include an [aip.dev/not-precedent][]
 comment explaining why.
 
 ```proto
-// (-- api-linter: core::0148::uid-format=disabled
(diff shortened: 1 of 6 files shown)


docs/rules/0148/uid-format.md
```diff
@@ -0,0 +1,79 @@
...
+
+```proto
+// Correct.
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/Book"
+    pattern: "books/{book}"
+  };
+
+  string name = 1;
```
andrei-scripniciuc, line 49: let's include the IDENTIFIER annotation on name fields in these examples, so that they're complete.

noahdietz: Done!

andrei-scripniciuc (approved):

PR #1271, opened 2023-10-13
feat(AIP-148): lint IP Address field format

noahdietz:
As mandated in AIP-148, and IP Address field must have `(google.api.field_info).format` with one of the IP Address formats. This rule enforces that.

### DIFF
--- a/rules/aip0148/ip_address_format.go
+++ b/rules/aip0148/ip_address_format.go
@@ -1,0 +1,56 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


rules/aip0148/ip_address_format.go
```diff
@@ -0,0 +1,56 @@
...
+
+var ipAddressFormats = []annotations.FieldInfo_Format{
+	annotations.FieldInfo_IPV4,
+	annotations.FieldInfo_IPV6,
+	annotations.FieldInfo_IPV4_OR_IPV6,
+}
+
+var ipAddressFormat = &lint.FieldRule{
+	Name: lint.NewRuleName(148, "ip-address-format"),
+	OnlyIf: func(fd *desc.FieldDescriptor) bool {
```
andrei-scripniciuc, line 34: we should also check that the field is of type string, to comply with go/aip/148#well-known-string-fields. I'm not sure if we should throw an error for this for an already existing field, where changing the type is a breaking change. Do we have any precedents for this kind of change to the linter?

noahdietz: We already have a linter rule for that guidance in AIP-202 :) https://linter.aip.dev/202/string-only-format

noahdietz: Oh I misunderstood the comment. I think that is a good idea, thanks.

noahdietz: Done!

andrei-scripniciuc (approved):

Issue #1108, opened 2023-03-03
Can you lint to ensure Java package name to be "com.google" for new Google services?


suztomo:
Can you lint to ensure java package name to be "com.google" for new Google services?

Without java_package, there's a risk that the proto writer creates a service that translates to "google." package rather than the appropriate "com.google". It happened in b/270579866. Can you detect such case?

Reference: http://linter.aip.dev/191

suztomo:
The immediate need is implemented in google-cloud-java https://github.com/googleapis/google-cloud-java/pull/9172. It would be nice to catch the mistake at api-linter.

PR #1289, opened 2023-11-21
feat(AIP-191): enforce java package has the prefix `com.google`

JoeWang1127:
Fix: #1108


noahdietz:
@JoeWang1127 Thanks for the PR! I think this makes sense to do, but perhaps not on the _public_ linter since anyone that uses it (and many non-Googlers do) doesn't need this guidance, and it would probably be annoying to them. I will contact you offline with instructions.

JoeWang1127:
@noahdietz thanks for reply.

Issue #1208, opened 2023-07-24
Feature request(AIP-192): Add a linter check for missing backticks


parthea:
See the code snippet below which has a missing backtick. This can cause issue in client library documentation and a linter rule would help prevent issues upstream.

```
  //  `"application/json"`. The default value is set to "application/json"`.
```
should be
```
  //  `"application/json"`. The default value is set to `"application/json"`.
```

https://github.com/googleapis/googleapis/blob/674883e3c55a16ed558ea0df1125537a956ff945/google/cloud/tasks/v2beta3/target.proto#L196

PR #1292, opened 2023-11-29
feat(AIP-192): Catch unpaired backticks in comments

acamadeo:
Part of #1208


rules/aip0192/closed_backticks.go
```diff
@@ -0,0 +1,183 @@
...
+	// If the backtick is only opening, it needs a closing backtick.
+	if backtick.opening && !backtick.closing {
+		return missingClosing
+	}
+	// Otherwise, the backtick is closing. It is paired if the last backtick was
+	// opening.
+	if prevBacktickOpen {
+		return paired
+	}
+	return missingOpening
```
acamadeo, line 149: One thing I realize now is that this implementation marks backticks that are both `opening` and `closing` as `missingOpening`. Perhaps a smarter way to handle this may be to assign a priority ranking among the `separators` and use that to decide whether it's `missingOpening` or `missingClosing`.

So for example, if ` ` is a stronger separator than `:`, you could have:
```
" `:thing" --> missingClosing
"thing:` " --> missingOpening
```

quirogas:
Since this hasn't been addressed in a while, I'm closing it

Issue #1090, opened 2023-02-10
AIP-123: disallow spaces in patterns


noahdietz:
During a recent API review, I discovered that the ApiLinter either didn't run on the CL or didn't catch that a google.api.resource pattern has spaces in the value. This is not allowed per aip.dev/123, so we need to make sure we are linting for it.

PR #1293, opened 2023-12-01
fix(AIP-123): disallow spaces in pattern

Juneezee:
Closes https://github.com/googleapis/api-linter/issues/1090.


noahdietz (approved):
Thanks!

Issue #1130, opened 2023-04-05
AIP-213: Recommend use of common types


noahdietz:
I often see the following fields during API review:
* `start_time/end_time`
* `string color`

These would be better represented using the common types we support e.g. `google.type.Interval` and `google.type.Color` respectively.

Perhaps we should add some rules to recommend those types when it _looks_ like they are mimicking them.

noahdietz:
Note: Some common type usage is recommended in more pertinent guidance e.g. the above `Interval` guidance is mentioned in AIP-145 already

We should probably have some actual guidance written down somewhere, in AIP-213 or elsewhere, (perhaps starting with one or two types rather than all of them) that we can then build rules around.

PR #1297, opened 2023-12-12
feat(AIP-213): Lint message fields that should use common types

acamadeo:
Part of #1130


rules/aip0213/common_types_fields.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/jhump/protoreflect/desc"
+)
+
+var fieldNameToCommonType = map[string]string{
+	"duration": "google.protobuf.Duration",
+
+	"color":  "google.type.Color",
+	"colour": "google.type.Color",
+
+	"dollars": "google.type.Money",
```
andrei-scripniciuc, line 30: for selecting common field names, I suggest using the public google API definitions here to identify commonly used patterns, for example: https://github.com/search?q=repo%3Agoogleapis%2Fgoogleapis+google.type.Money+lang%3Aproto&type=code


To give one specific example, for google.type.Money, I suggest dropping all proposed aliases, and instead just linting fields ending in "price", "cost", "_fee". you can look up the public google API repository here to see what commonly used patterns we use today: https://github.com/search?q=repo%3Agoogleapis%2Fgoogleapis+google.type.Money+lang%3Aproto&type=code

if the currency is implied by the field name we probably don't want to enforce using the Money type, since it's duplicating information and can lead to ambiguous / confusing behavior when the currency conflicts with the field name

also, "money" is very unlikely to be used as a field name, I have not seen examples of that. Perhaps "money_value" since it's used in 3 examples in public google APIs? https://github.com/search?q=repo%3Agoogleapis%2Fgoogleapis+%22google.type.Money+money%22+lang%3Aproto&type=code

rules/aip0213/common_types_fields.go
```diff
@@ -0,0 +1,63 @@
...
+	"time_of_day": "google.type.TimeOfDay",
+
+	"timezone":  "google.type.TimeZone",
+	"time_zone": "google.type.TimeZone",
+}
+
+var commonTypesFields = &lint.FieldRule{
+	Name: lint.NewRuleName(213, "common-types-fields"),
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		// Flag this field if it has a name in `fieldNameToCommonType` but
```
andrei-scripniciuc, line 51: I suggest checking for postfixes instead of equality. It's common for fields to contain more than just the type, for example "background_color", "max_retry_duration"

rules/aip0213/common_types_fields.go
acamadeo: Done

rules/aip0213/common_types_fields.go
acamadeo: Done

quirogas:
Since this hasn't been addressed in a while, I'm closing it

PR #1304, opened 2023-12-20
feat: undelete should not be required for declarative-friendly interfaces

andrei-scripniciuc:
according to https://google.aip.dev/164#declarative-friendly-resources:

> Soft-deletable resources have a poorer experience than hard-deleted resources in declarative clients

this contradicts the requirement that declarative-friendly interfaces implement soft-delete.


noahdietz (commented):
Please remove the appropriate `docs/` page as well

bgrant0607:
There are more occurrences:
https://github.com/search?q=repo%3Agoogleapis%2Fapi-linter%20declarativeFriendlyRequired&type=code

bgrant0607:
Ok, this is just removing the check from 164, not from everywhere.

noahdietz (approved):

PR #1325, opened 2024-01-24
fix(AIP-123): allow name in nested messages

noahdietz:
It is reasonable to allow the use of a `name` field in a nested message since nested messages are most often authored as a small piece of a larger message, thus never really candidates for being a resource.

### DIFF
--- a/docs/rules/0123/resource-annotation.md
+++ b/docs/rules/0123/resource-annotation.md
@@ -10,15 +10,15 @@ redirect_from:
 
 # Resource annotation presence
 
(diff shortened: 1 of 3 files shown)


andrei-scripniciuc (changes requested):
thanks for sending this! could you update the rule docs as well with this exception?

noahdietz:
> thanks for sending this! could you update the rule docs as well with this exception?

Yep! Done.

andrei-scripniciuc (approved):

PR #1335, opened 2024-02-16
fix(AIP 133-135): fix false positive in 133-135

andrei-scripniciuc:
When a resource name `Operation`, the linter assumes that any method returning `google.longrunning.Operation` is actually returning the resource type: https://github.com/googleapis/api-linter/blob/main/rules/aip0133/aip0133.go#L52C1-L61C2

This change ignores the heuristic that assumes the response message contains the resource name, if the response is `google.longrunning.Operation`.


rules/internal/utils/type_name.go
```diff
@@ -40,3 +40,19 @@ func GetTypeName(f *desc.FieldDescriptor) string {
...
+
+	// Usually the response message will be the resource message, and its name will
+	// be part of method name (make a double check here to avoid the issue when
+	// method or output naming doesn't follow the right principles)
+	// Ignore this rule if the return type is an LRO
+	if strings.Contains(m.GetName()[len(expectedVerb):], m.GetOutputType().GetName()) && !IsOperation(m.GetOutputType()) {
+		return m.GetOutputType().GetName()
+	}
+	return m.GetName()[len(expectedVerb):]
+}
```
noahdietz, line 58:
```suggestion
}

```
Add new line at elf

noahdietz (approved):

PR #1347, opened 2024-02-29
fix(AIP-121): recognize standard list revisions

noahdietz:
Since AIP-162 recommends Revisions being a proper resource collection, it will also be checked for standard List and Get RPCs as per AIP-121. However, the rule enforcing the Standard List requirement was ignoring list revision methods explicitly. This brings `List{Resource}Revisions` into consideration for that rule.

Fixes [internal bug](http://b/316598392)

### DIFF
--- a/rules/aip0121/resource_must_support_list.go
+++ b/rules/aip0121/resource_must_support_list.go
@@ -39,7 +39,7 @@ var resourceMustSupportList = &lint.ServiceRule{
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


noahdietz:
May be overtaken by https://github.com/googleapis/api-linter/pull/1348

andrei-scripniciuc (approved):

noahdietz:
This should be necessary anymore since #1348 brings "new" list revisions methods into scope of Standard List.

PR #1366, opened 2024-04-10
feat(AIP-133): ignore create methods with invalid LRO response types

andrei-scripniciuc:
This addresses a nil dereferencing error due to `ot` being `nil` in https://github.com/andrei-scripniciuc/api-linter/blob/f6108f07532aae07c4f305c7da99092f0114cea2/rules/aip0133/request_required_fields.go#L50

### DIFF
--- a/rules/internal/utils/method.go
+++ b/rules/internal/utils/method.go
@@ -45,12 +45,7 @@ func IsCreateMethodWithResolvedReturnType(m *desc.MethodDescriptor) bool {
 		return false
 	}
 
-	res := m.GetOutputType()
-
-	if info := GetOperationInfo(m); info != nil {
-		res = FindMessage(m.GetFile(), info.GetResponseType())
-	}
-	return res != nil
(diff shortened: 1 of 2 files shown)


rules/internal/utils/method.go
```diff
@@ -36,6 +36,23 @@ func IsCreateMethod(m *desc.MethodDescriptor) bool {
...
+	if !IsCreateMethod(m) {
+		return false
+	}
+
+	res := m.GetOutputType()
+
+	if info := GetOperationInfo(m); info != nil {
+		res = FindMessage(m.GetFile(), info.GetResponseType())
+	}
+	return res != nil
```
noahdietz, line 53: Could use [`GetResponseType`](https://pkg.go.dev/github.com/googleapis/api-linter/rules/internal/utils#GetResponseType) in this package:
```suggestion
	return GetResponseType(m) != nil
```

noahdietz (approved):

PR #1387, opened 2024-05-15
feat(aip-136): response message name

liveFreeOrCode:
[AIP-136](https://google.aip.dev/136)

> - Custom methods should return a response message matching the RPC name, with a Response suffix.
>   - When operating on a specific resource, a custom method may return the resource itself

### DIFF
--- a/docs/rules/0136/response-message-name.md
+++ b/docs/rules/0136/response-message-name.md
@@ -20,10 +20,9 @@ This rule looks at any method that is not a standard method, and complains if
 the name of the corresponding output message does not match the name of the RPC
 with the suffix `Response` appended, or the resource being operated on.
 
-**Note:** To identify the resource being operated on, the rule inspects the
-name path parameter, which maps to the `name` field on the input type checking
-that the resource message derived from `(google.api.resource_reference).type`
-matches the response resource.
+**Note:** To identify the resource being operated on, the rule checks for
+the request field `name`, and comparing its `(google.api.resource_reference).type`,
+if present, to the response message's `(google.api.resource).type`, if present.
 
 ## Examples
 
--- a/docs/rules/0136/response-message-name.md
+++ b/docs/rules/0136/response-message-name.md
@@ -1,0 +1,113 @@
+---
+rule:
+  aip: 136
+  name: [core, '0136', response-message-name]
+  summary: Custom methods must have standardized response message names.
+permalink: /136/response-message-name
+redirect_from:
+  - /0136/response-message-name
+---
+
+# Custom methods: Response message
+
+This rule enforces that all custom methods should take a response message
+matching the RPC name, with a `Response` suffix, or the resource being operated
+on [AIP-136][].
+
+## Details
+
+This rule looks at any method that is not a standard method, and complains if
+the name of the corresponding input message does not match the name of the RPC
+with the suffix `Response` appended, or the resource itself. It will use the
+`(google.api.resource_reference)` on the first field of the request message to
+determine the resource name that should be used.
+
+## Examples
+
+### Response Suffix
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+// Should be `ArchiveBookResponse`.
+rpc ArchiveBook(ArchiveBookRequest) returns (ArchiveBookResp) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:archive"
+    body: "*"
+  };
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+rpc ArchiveBook(ArchiveBookRequest) returns (ArchiveBookResponse) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:archive"
+    body: "*"
+  };
+
+```
+
+### Resource
+
+**Incorrect** code for this rule:
+
+```proto
+// Incorrect.
+// Should be `Book`.
+rpc ArchiveBook(ArchiveBookRequest) returns (Author) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:archive"
+    body: "*"
+  };
+}
+
+message ArchiveBookRequest {
+    // The book to archive.
+    // Format: publishers/{publisher}/books/{book}
+    string name = 1 [(google.api.resource_reference).type = "library.googleapis.com/Book"];
+}
+```
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+rpc ArchiveBook(ArchiveBookRequest) returns (Book) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:archive"
+    body: "*"
+  };
+}
+
+message ArchiveBookRequest {
+    // The book to archive.
+    // Format: publishers/{publisher}/books/{book}
+    string name = 1 [(google.api.resource_reference).type = "library.googleapis.com/Book"];
+}
+```
+
+## Disabling
+
+If you need to violate this rule, use a leading comment above the method.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+// (-- api-linter: core::0136::response-message-name=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+rpc ArchiveBook(ArchiveBookRequest) returns (ArchiveBookResp) {
+  option (google.api.http) = {
+    post: "/v1/{name=publishers/*/books/*}:archive"
+    body: "*"
+  };
+}
+```
+
+If you need to violate this rule for an entire file, place the comment at the
+top of the file.
+
+[aip-136]: https://aip.dev/136
+[aip.dev/not-precedent]: https://aip.dev/not-precedent
--- a/docs/rules/0136/response-message-name.md
+++ b/docs/rules/0136/response-message-name.md
@@ -17,10 +17,13 @@ on [AIP-136][].
 ## Details
 
 This rule looks at any method that is not a standard method, and complains if
-the name of the corresponding input message does not match the name of the RPC
-with the suffix `Response` appended, or the resource itself. It will use the
-`(google.api.resource_reference)` on the first field of the request message to
-determine the resource name that should be used.
+the name of the corresponding output message does not match the name of the RPC
+with the suffix `Response` appended, or the resource being operated on.
+
+**Note:** To identify the resource being operated on, the rule inspects the
+name path parameter, which maps to the `name` field on the input type checking
+that the resource message derived from `(google.api.resource_reference).type`
+matches the response resource.
 
 ## Examples
 
@@ -30,10 +33,10 @@ determine the resource name that should be used.
 
 ```proto
 // Incorrect.
-// Should be `ArchiveBookResponse`.
-rpc ArchiveBook(ArchiveBookRequest) returns (ArchiveBookResp) {
+// Should be `TranslateTextResponse`.
+rpc TranslateText(TranslateTextRequest) returns (Text) {
   option (google.api.http) = {
-    post: "/v1/{name=publishers/*/books/*}:archive"
+    post: "/v1/{project=projects/*}:translateText"
     body: "*"
   };
 }
@@ -43,12 +46,12 @@ rpc ArchiveBook(ArchiveBookRequest) returns (ArchiveBookResp) {
 
 ```proto
 // Correct.
-rpc ArchiveBook(ArchiveBookRequest) returns (ArchiveBookResponse) {
+rpc TranslateText(TranslateTextRequest) returns (TranslateTextResponse) {
   option (google.api.http) = {
-    post: "/v1/{name=publishers/*/books/*}:archive"
+    post: "/v1/{project=projects/*}:translateText"
     body: "*"
   };
-
+}
 ```
 
 ### Resource
@@ -66,16 +69,15 @@ rpc ArchiveBook(ArchiveBookRequest) returns (Author) {
 }
 
 message ArchiveBookRequest {
-    // The book to archive.
-    // Format: publishers/{publisher}/books/{book}
-    string name = 1 [(google.api.resource_reference).type = "library.googleapis.com/Book"];
+  // The book to archive.
+  // Format: publishers/{publisher}/books/{book}
+  string name = 1 [(google.api.resource_reference).type = "library.googleapis.com/Book"];
 }
 ```
 
 **Correct** code for this rule:
 
 ```proto
-// Correct.
 rpc ArchiveBook(ArchiveBookRequest) returns (Book) {
   option (google.api.http) = {
     post: "/v1/{name=publishers/*/books/*}:archive"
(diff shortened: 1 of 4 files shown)


rules/aip0136/response_message_name.go
```diff
@@ -0,0 +1,64 @@
...
+	OnlyIf: isCustomMethod,
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		matchingResponseName := m.GetName() + "Response"
+		suggestion := matchingResponseName
+
+		got := m.GetOutputType().GetName()
+		want := stringset.New(matchingResponseName)
+
+		// Get the resource name from the first request message field, if it
+		// is a resource reference
```
liveFreeOrCode, line 40: Interested to hear opinions about this approach - not sure if there are better ways of doing this.  Using `name` with a resource_reference seems to be convention for custom methods acting on a resource, but it's not spelled out in the AIPs.

noahdietz: Yeah this will be a bit hand wavy.

What I think you are trying to do here (correct me if I'm wrong) is correlate a resource referenced in the request to the allowed resource message type name. There are a few issues with this (and I share them for education not trying to roast you 😅 ):
1. Exactly what you mentioned - there is no clearly defined pattern for finding *the* resource from the request that would be reasonable as a response message (could be field `name`, `parent`, `location`, `book`, etc.) - `name` is the _typical_ one for "should return the resource" but I'm not willing to bet this is always the case, ya know?
2. A resource being referenced in the request doesn't mean it is reasonable to return in the response i.e. we don't get enough signal from the field name alone to understand the semantics of the method
3. The reference may be a direct reference to a parent resource type rather than an indirect one via a child_type reference

It all kind of boils down to lately I have tried to avoid banking rules on heuristics informed by style i.e., field naming, since it's caused more friction than we'd like.

I'd separate the concerns and simply check if the response type is annotated as a resource (using `utils.IsResource`) as an adequate alternative to a `{MethodName}Response` response message name.

From there, we can add a separate rule to check "is this the right resource to return". This will give us a bit more flexibility. This is my rule of thumb no pun intended - rules should lint 1 thing and 1 thing only, anything extra is split into a separate rule.

WDYT? feel free to disagree/counter/ask questions, not trying to force this one

liveFreeOrCode: > From there, we can add a separate rule to check "is this the right resource to return". This will give us a bit more flexibility. This is my rule of thumb no pun intended - rules should lint 1 thing and 1 thing only, anything extra is split into a separate rule.

I'm not entirely sure about this statement.  When you say "rules should lint 1 thing and 1 thing only, anything extra is split into a separate rule."  Can that be accomplished currently by this framework?

To me that would look something like this:

```
rules/aip0136/response_message_name_response_suffix.go
rules/aip0136/response_message_name_resource.go
```

But these are mutually exclusive such that one will always fail even when the other succeeds.  The logic for how we _lint the response message name_ has to happen in a single lint rule, right?

> I'd separate the concerns and simply check if the response type is annotated as a resource (using utils.IsResource) as an adequate alternative to a {MethodName}Response response message name.

I think this is a perfectly acceptable first pass, as it lints to the spirit of the rule.  Perhaps a second pass could to ascertain all of the models from the given request type and see if the model being returned is any of those.  Assuming it would be safe to guide developers to not response with implicitly modified resources derived from input.

noahdietz: > But these are mutually exclusive such that one will always fail even when the other succeeds. The logic for how we lint the response message name has to happen in a single lint rule, right?

I'm kind of splitting hairs here, so 🐻 with me, but...what I was trying to suggest was:

* one rule enforcing that a custom method response message either: ends with `{Method}Response` OR is _any_ resource
* one rule enforcing that if a custom method response message is a resource, that it is *the right* resource

The second one is harder as we've discussed, and probably won't be _perfect_. WDYT?

> I think this is a perfectly acceptable first pass, as it lints to the spirit of the rule.

Agreed!

> Perhaps a second pass could to ascertain all of the models from the given request type and see if the model being returned is any of those. Assuming it would be safe to guide developers to not response with implicitly modified resources derived from input.

Yeah, exactly, and agreed, we'd run the risk of recommending the wrong resource 😅 --> developers dislike/lose trust in the linter

liveFreeOrCode: Rethinking some of this @noahdietz, let me know if this sounds right to you.

According to https://google.aip.dev/136#resource-based-custom-methods

> The parameter for the resource's name must be called name, and be the only variable in the URI path.

What this tells me is that the only time a _resource_ could be returned is when `name` is a path param, which would mean it would exist on the request message type.

I think it would be safe to say:
- If the response is a resource type:
  - And if the `name` parameter is provided (this should be linted separately)
    - Then the name.resource_reference.type == response.resouce.type

liveFreeOrCode: I'm nervous about releasing something that would allow this to pass, which could deviate from the spirit of the rules, and also as you mentioned, cause distrust in the tool

```proto
service Library {
  rpc ArchiveBook(ArchiveBookRequest) returns (Author);
}
```

noahdietz: Good point! That was a (somewhat) recent change specifically to address the ambiguity of resource-based custom methods. It gives us a leg to stand on here.

Agreed your example would be scary to allow :)

> If the response is a resource type:
And if the name parameter is provided (this should be linted separately)
Then the name.resource_reference.type == response.resouce.type

Perfect, let's do it. Love the "this should be linted separately" 🙏🏻 

Thanks for working through that with me!

rules/aip0136/response_message_name.go
```diff
@@ -0,0 +1,69 @@
...
+		// in https://google.aip.dev/136#resource-based-custom-methods
+
+		response := utils.GetResponseType(m)
+		requestType := utils.GetResourceReference(m.GetInputType().FindFieldByName("name")).GetType()
+		resourceOperatedOn := utils.FindResourceMessage(requestType, m.GetFile())
+
+		// Output type has `Response` suffix
+		if len(utils.LintMethodHasMatchingResponseName(m)) == 0 {
+			return nil
+		}
```
noahdietz, line 46: I'd move this above the resource-related stuff to short circuit the rule. Note the saving of the suffix-based findings for later...

```suggestion
    // Output type has `Response` suffix
    suffixFindings := utils.LintMethodHasMatchingResponseName(m)
		if len(suffixFindings) == 0 {
			return nil
		}

		response := utils.GetResponseType(m)
		requestType := utils.GetResourceReference(m.GetInputType().FindFieldByName("name")).GetType()
		resourceOperatedOn := utils.FindResourceMessage(requestType, m.GetFile())
```

rules/aip0136/response_message_name.go
```diff
@@ -0,0 +1,69 @@
...
+		// A response is considered valid if
+		// - The response name matches the RPC name with a `Response` suffix
+		// - The response is the resource being operated on
+		// To identify the resource being operated on, we inspect the Resource
+		// Reference of the input type's `name` field. This guidance is documented
+		// in https://google.aip.dev/136#resource-based-custom-methods
+
+		response := utils.GetResponseType(m)
+		requestType := utils.GetResourceReference(m.GetInputType().FindFieldByName("name")).GetType()
+		resourceOperatedOn := utils.FindResourceMessage(requestType, m.GetFile())
```
noahdietz, line 41: We don't technically need to do a message type look up - we can simply match the resource_reference type to the response's `google.api.resource` type. So a slightly simpler thing to do would be:

```
response := utils.GetResponseType(m)
requestResourceType := utils.GetResourceReference(m.GetInputType().FindFieldByName("name")).GetType()

if !utils.IsResource(response) || requestResourceType == "" {
  return suffixFindings // from the "lint method has matching response" call moved earlier, see other comment
}

responseResourceType := utils.GetResource(response).GetType()
if responseResourceType == requestResourceType {
  // all good!
  return nil
}

// return findings defined below for the either-or guidance violation
```

This also reads straight down a bit better, exiting early whenever possible.

Note: in testing you will need to add a resource annotation on the response message

rules/aip0136/response_message_name.go
```diff
@@ -0,0 +1,69 @@
...
+
+		// Output type has `Response` suffix
+		if len(utils.LintMethodHasMatchingResponseName(m)) == 0 {
+			return nil
+		}
+
+		// Response is the resource being operated on
+		if response == resourceOperatedOn {
+			return nil
+		}
```
noahdietz, line 51: The best way to compare protos is using [proto.Equal](https://pkg.go.dev/google.golang.org/protobuf@v1.34.1/proto#Equal), but see other comments for a simpler approach to checking this

rules/aip0136/response_message_name.go
```diff
@@ -0,0 +1,69 @@
...
+
+		msg := "Custom methods should return a message matching the RPC name, with a `Response` suffix, or the resource being operated on, not %q."
+		got := m.GetName()
+
+		suggestion := got + "Response"
+		if utils.IsResource(response) && resourceOperatedOn != nil {
+			suggestion += " or " + resourceOperatedOn.GetName()
```
noahdietz, line 58: So this needs to be a "machine readable" suggestion i.e. one that could be dropped in directly as a replacement, not just human readable: https://github.com/googleapis/api-linter/blob/a7290a746589155df64410cdc8267e1ede1a694e/lint/problem.go#L33-L40

So if the request has the annotated `name` field, the resource message name could be the suggestion (i see now why you looked up the actual message backing the Resource annotation), but it could reasonably be the `Response` suffixed message as well (I think this is probably most common tbh)

Honestly, I don't know if we should provide a `Suggestion` at all given that there are potentially two reasonable options for a resource-oriented custom method.

docs/rules/0136/response-message-name.md
```diff
@@ -0,0 +1,159 @@
...
+permalink: /136/response-message-name
+redirect_from:
+  - /0136/response-message-name
+---
+
+# Custom methods: Response message
+
+This rule enforces that all custom methods should take a response message
+matching the RPC name, with a `Response` suffix, or the resource being operated
+on [AIP-136][].
```
noahdietz, line 15:
```suggestion
This rule enforces that custom methods have a response message that is named to
match the RPC name with a `Response` suffix, or that is the resource being operated
on, as described in [AIP-136][].
```

Just wanted to tighten it up a bit, take the "should" out of it, the guidance on AIP-136 will dictate the requirement level.

docs/rules/0136/response-message-name.md
```diff
@@ -0,0 +1,159 @@
...
+## Details
+
+This rule looks at any method that is not a standard method, and complains if
+the name of the corresponding output message does not match the name of the RPC
+with the suffix `Response` appended, or the resource being operated on.
+
+**Note:** To identify the resource being operated on, the rule inspects the
+name path parameter, which maps to the `name` field on the input type checking
+that the resource message derived from `(google.api.resource_reference).type`
+matches the response resource.
```
noahdietz, line 26: Love that you included this.

Just want to tighten it up a bit.
```suggestion
**Note:** To identify the resource being operated on, the rule checks for
the request field `name`, and comparing its `(google.api.resource_reference).type`,
if present, to the response message's `(google.api.resource).type`, if present.
```

rules/aip0136/response_message_name.go
liveFreeOrCode: We need the message however for a suggestion.

According to https://google.aip.dev/client-libraries/4231#resource-messages

> The universal resource type usually (but not always) ends with a name that matches the name of the message.

So I figured, because I needed to fetch the message type from the file for the suggestion, it could serve both purposes without having to do the type match.  I can still do the type match as well for defensive purposes if you think it'll help?

noahdietz: See later comment re:the `Suggestion` :)

rules/aip0136/response_message_name.go
liveFreeOrCode: Yea, I debated on this for a while.  This is the nudge I need to remove the suggestion entirely :)

rules/aip0136/response_message_name.go
liveFreeOrCode: Yeah, replied before I got through all the comments haha :)

Addressed this here: https://github.com/googleapis/api-linter/pull/1387/commits/b007953dcd47b31c661c4f18e1694236715c3e4b

However, I didn't want to return the `suffixFindings` because I wanted to make sure the error message was the same through all of the situations so that consumers understood it could be either or, and in no situation should we suggest something... yet at least.

noahdietz: > However, I didn't want to return the suffixFindings because I wanted to make sure the error message was the same through all of the situations so that consumers understood it could be either or, and in no situation should we suggest something... yet at least.

Fair enough, I agree, that's fine by me!

liveFreeOrCode: Cleaning up one more typo in my next commit, then should be GTG for a final pass

rules/aip0136/response_message_name_test.go
```diff
@@ -30,7 +30,8 @@ func TestResponseMessageName(t *testing.T) {
 			problems        testutils.Problems
 		}{
 			{"Valid", "ArchiveBook", "ArchiveBookResponse", testutils.Problems{}},
-			{"Invalid", "ArchiveBook", "ArchiveBookResp", testutils.Problems{{Suggestion: "ArchiveBookResponse"}}},
+			{"Invalid", "ArchiveBook", "ArchiveBookResp", testutils.Problems{{
+				Message: "Custom methods should return a message matching the RPC name, with a `Response` suffix, or the resource being operated on, not \"ArchiveBookResp\"."}}},
```
noahdietz, line 34: Nit: the `testutils.Problems` `Diff` method actually does a `strings.Contains`, so we can pick a chunk of the expected problem message and just use that e.g. 
```suggestion
				Message: "matching the RPC name, with a `Response` suffix}}},
```

https://github.com/googleapis/api-linter/blob/a7290a746589155df64410cdc8267e1ede1a694e/rules/internal/testutils/problems.go#L54-L58

might seem odd but makes for easier/cleaner tests, especially considering there is only one rule being run here so no chance to confuse with another rule

noahdietz (approved):
Very well done, thank you @liveFreeOrCode

rules/aip0136/response_message_name.go
noahdietz: Feel free to hit the "re-request review" button whenever: 
image

rules/aip0136/response_message_name_test.go
liveFreeOrCode: Oh, cool.  Good to know.  I'll sneak that in really quick. Also qq about releases.  I noticed this didn't fire off a release: https://github.com/googleapis/api-linter/commit/46a6e43d1d6bb6a6e79131866f16f0b1dfd2e326

Should I rename this PR before it's merged?  Is it because of the (AIP-136) in the PR title?

noahdietz: > Also qq about releases. I noticed this didn't fire off a release: https://github.com/googleapis/api-linter/commit/46a6e43d1d6bb6a6e79131866f16f0b1dfd2e326
>
> Should I rename this PR before it's merged? Is it because of the ([AIP-136](https://google.aip.dev/136)) in the PR title?

Nope, all is WAI, it's in the open release PR: https://github.com/googleapis/api-linter/pull/1388

I just haven't pressed the buttons yet knowing we'd have this.

PR #1394, opened 2024-05-29
feat(AIP-134): update_mask must be OPTIONAL

noahdietz:
Implements enforcement for AIP-134 change https://github.com/aip-dev/google.aip.dev/pull/1327.

Previously, `update_mask` could be either `OPTIONAL` or `REQUIRED`. Now it can and must only be `OPTIONAL`.

Adds `locations.FieldBehavior` helper for annotation source code location, and refactors one other place to use it.

### DIFF
--- a/docs/rules/0134/update-mask-optional-behavior.md
+++ b/docs/rules/0134/update-mask-optional-behavior.md
@@ -1,0 +1,72 @@
+---
+rule:
+  aip: 134
+  name: [core, '0134', update-mask-optional-behavior]
+  summary: Standard Update `update_mask` field must be `OPTIONAL`.
+permalink: /134/update-mask-optional-behavior
+redirect_from:
+  - /0134/update-mask-optional-behavior
+---
+
+# Update methods: Mask field expected behavior
+
+This rule enforces that the `update_mask` field of a Standard `Update` request
+uses `google.api.field_behavior = OPTIONAL`, as mandated in [AIP-134][].
+
+## Details
+
+This rule looks at any field named `update_mask` that's in a `Update*Request`
+and complains if it the field is not annotated with
+`google.api.field_behavior = OPTIONAL`.
+
+## Examples
+
+**Incorrect** code for this rule:
+
+```proto
+message UpdateBookRequest {
+  Book book = 1;
+
+  // Incorrect. Must be `OPTIONAL`.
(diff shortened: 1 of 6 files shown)


docs/rules/0134/update-mask-optional-behavior.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+# Update methods: Mask field expected behavior
+
+This rule enforces that the `update_mask` field of a Standard `Update` request
+uses `google.api.field_behavior = OPTIONAL`, as mandated in [AIP-134][].
+
+## Details
+
+This rule looks at any field named `update_mask` that's in a `Update*Request`
```
jskeet, line 18: a => an

docs/rules/0134/update-mask-optional-behavior.md
```diff
@@ -0,0 +1,72 @@
...
+and complains if it the field is not annotated with
```
jskeet, line 19: Remove "it"

docs/rules/0134/update-mask-optional-behavior.md
```diff
@@ -0,0 +1,72 @@
...
+`google.api.field_behavior = OPTIONAL`.
+
+## Examples
+
+**Incorrect** code for this rule:
```
jskeet, line 24: This shows a field with an incorrect field_behavior - what if the behavior is missing entirely? (My guess is that this is covered by a different rule.)

jskeet: (I see there's a test for this - maybe it could be included as another incorrect example? Not sure if we ever have multiple incorrect examples.)

noahdietz: We don't typically include multiple incorrect examples, no, but yes it is handled.

docs/rules/0134/update-mask-optional-behavior.md
noahdietz: Done, thanks!

docs/rules/0134/update-mask-optional-behavior.md
noahdietz: Done, thanks!

jskeet (approved):

PR #1396, opened 2024-06-04
fix(aip-130): identify standard and custom methods

liveFreeOrCode:
There is a problem with the Response Message Name linter of AIP-136 currently.

For reference: https://github.com/googleapis/api-linter/blob/main/rules/aip0136/response_message_name.go

This, and all linter methods in aip0136 used a method `isCustomMethod`.  The logic of this method is incomplete/incorrect. It assumed anything with a `":"` in it's URI was considered a custom method.  But, according to https://google.aip.dev/130, Batch methods are also considered Standard methods.  Batch methods have the following URI segments: `:batchGet`, `:batchUpdate`, `:batchCreate` and `:batchDelete`.

What happens?

```proto
service Library {
  // This fails the aip0136 response_message_name lint rule
  rpc BatchDeleteBooks(BatchDeleteBooksRequest) returns (google.protobuf.Empty) {
    option (google.api.http) = {
      post: "/v1/{parent=publishers/*}/books:batchDelete"
      body: "*"
    };
  }
}
```

### DIFF
--- a/rules/aip0136/aip0136.go
+++ b/rules/aip0136/aip0136.go
@@ -16,11 +16,7 @@
 package aip0136
 
 import (
-	"strings"
-
 	"github.com/googleapis/api-linter/lint"
-	"github.com/googleapis/api-linter/rules/internal/utils"
-	"github.com/jhump/protoreflect/desc"
 )
 
 // AddRules accepts a register function and registers each of
@@ -41,23 +37,4 @@ func AddRules(r lint.RuleRegistry) error {
 		// httpNameVariable,
 		// httpParentVariable,
 	)
-}
-
-func isCustomMethod(m *desc.MethodDescriptor) bool {
-	// Anything with a `:` in the method URI is automatically a custom
-	// method, regardless of the RPC name.
-	for _, httpRule := range utils.GetHTTPRules(m) {
-		if strings.Contains(httpRule.GetPlainURI(), ":") {
-			return true
-		}
-	}
-
-	// Methods with no `:` in the URI are standard methods if they begin with
-	// one of the standard method names.
-	for _, prefix := range []string{"Get", "List", "Create", "Update", "Delete", "Replace"} {
-		if strings.HasPrefix(m.GetName(), prefix) {
-			return false
-		}
-	}
-	return true
 }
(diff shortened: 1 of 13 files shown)


rules/aip0136/http_uri_suffix_test.go
```diff
@@ -39,14 +39,8 @@ func TestURISuffix(t *testing.T) {
 		{"ValidOneWord", "Translate", "/v3:translate", testutils.Problems{}},
 		{"ValidStdMethod", "GetBook", "/v1/{name=publishers/*/books/*}", testutils.Problems{}},
 		{"ValidTwoWordNoun", "WriteAudioBook", "/v1/{name=publishers/*/audioBooks/*}:write", testutils.Problems{}},
-		{"ValidListRevisions", "ListBookRevisions", "/v1/{name=publishers/*/books/*}:listRevisions", testutils.Problems{}},
```
liveFreeOrCode, line 42: The [History](https://google.aip.dev/162#history) of resource revisions indicates these _used_ to be custom methods, hinting at the fact that they are no longer.

rules/aip0136/aip0136.go
```diff
@@ -42,22 +38,3 @@ func AddRules(r lint.RuleRegistry) error {
...
-	// method, regardless of the RPC name.
-	for _, httpRule := range utils.GetHTTPRules(m) {
-		if strings.Contains(httpRule.GetPlainURI(), ":") {
-			return true
-		}
-	}
-
-	// Methods with no `:` in the URI are standard methods if they begin with
-	// one of the standard method names.
-	for _, prefix := range []string{"Get", "List", "Create", "Update", "Delete", "Replace"} {
```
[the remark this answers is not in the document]
noahdietz: We can leave it out. Thanks for calling this out.

rules/aip0136/http_uri_suffix_test.go
noahdietz: Right, AIP-162 is being reworked...technically there are still instances of these methods, but new ones are no longer allowed. Let's roll with this for now, I think they will still pass most custom method rules anyways.

noahdietz (approved):
Thanks for this, good catch. Good reminder for me that double checking the heuristic(s) used to identify things should be reviewed regularly...while checking for `:` in the HTTP path, it is incomplete...

777easy (approved):

777easy:
Close

PR #1403, opened 2024-07-03
feat(AIP-123): enforce singular as ID segment

noahdietz:
_Fair warning, this is a big PR._

Enforces the following guidance from AIP-123:
> Pattern variables **must*** be the singular form of the resource type e.g. a pattern variable representing a `Topic` resource ID is named `{topic}`.

It takes special consideration for stuttering child resource names as documented in [AIP-122 Nested Collections](https://aip.dev/122#nested-collections). It does this in a somewhat awkward way in that it compares the resource `singular` form in `snake_case` to the _parent_ resource ID segment (if present). If the parent ID segment is a prefix of the snake_case `singular` for the resource, then it is considered nested e.g. `{author}` parent ID segment compared to `author_credit` would be considered nested and eligible for reduction to just `credit` in the resource pattern(s). This handles child resources that are singletons as well.

It only recommends this reduction in the following cases:
* for the first pattern if it is just incorrect
* for any incorrect pattern following the first _iff_ the first is in fact reduced (because the guidance requires consistency)

The same needs to be done for `plural` as collection ID in resource patterns and they should be merged/released together.

For internal bug b/343465929

P.S.: also updates the "bugged" `resource-singular` rule documentation - copy+paste strikes again.


docs/rules/0123/resource-pattern-singular.md
```diff
@@ -0,0 +1,85 @@
...
+  summary: Resource patterns must use the singular as the resource ID segment
+permalink: /123/resource-pattern-singular
+redirect_from:
+  - /0123/resource-pattern-singular
+---
+
+# Resource `pattern` use of `singular`
+
+This rule enforces that messages that have a `google.api.resource` annotation
+use the `singular` as the resource ID segment, as described in [AIP-123][].
```
jskeet, line 14: Add "form"? ("use the `singular` form as the resource ID segment")

rules/aip0123/aip0123.go
```diff
@@ -76,6 +78,91 @@ func getVariables(pattern string) []string {
...
 }
 
+// isTopLevelResourcePattern determines if the given pattern is that of a
+// top-level resource by checking how many segments it has - top-level resource
+// patterns have only two segments, thus one delimeter.
+func isTopLevelResourcePattern(pattern string) bool {
+	return strings.Count(pattern, "/") == 1
+}
+
+// getparentIDVariable is a helper that returns the parent resource ID segment
```
jskeet, line 88: getp => getP

rules/aip0123/aip0123.go
```diff
@@ -76,6 +78,91 @@ func getVariables(pattern string) []string {
 	return answer
 }
 
+// isTopLevelResourcePattern determines if the given pattern is that of a
```
jskeet, line 81: While this is syntactically correct, I think we often treat things like "project/{project}/locations/{location}" as *effectively* top-level. I don't know whether that's going to be relevant here, but I thought I'd mention it.

jskeet (commented):
I'm struggling to comprehend this one in much detail (not enough coffee?) so I only have very superficial comments. I may have another look later.

rules/aip0123/aip0123.go
noahdietz: I see what you are saying. Let me see if I can come up with a better name...just want to see if the pattern indicates if it has a parent or not by detecting if there are more than two segments (where odd # of segments is a singleton, and event # >= 4 is _technically_ a nested collection)

docs/rules/0123/resource-pattern-singular.md
noahdietz: sure thing! Done

rules/aip0123/aip0123.go
noahdietz: Good catch, thanks.

rules/aip0123/aip0123.go
jskeet: Yup, that feels clearer to me.

jskeet (approved):
(I haven't re-reviewed everything - I'm just approving on the grounds that my comments have all been addressed. Getting another review from someone else would probably be a good idea.)

rules/aip0123/resource_pattern_singular.go
```diff
@@ -0,0 +1,75 @@
...
+		if nested {
+			nn = nestedSingular(res)
+		}
+
+		patterns := res.GetPattern()
+		singular := utils.GetResourceSingular(res)
+
+		if !utils.IsSingletonResource(m) {
+			singular = fmt.Sprintf("{%s}", strcase.SnakeCase(singular))
+			nn = fmt.Sprintf("{%s}", nn)
```
slevenick, line 48: If you have a non-nested singleton will this be empty as `nn` will be unset? Or is that case impossible?

docs/rules/0123/resource-pattern-singular.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Correct** code for this rule:
+
+```proto
+// Correct.
+message Book {
+  option (google.api.resource) = {
+    type: "library.googleapis.com/BookShelf"
+    pattern: "publishers/{publisher}/bookShelves/{book_shelf}"
+    singular: "bookShelf",
```
slevenick, line 55: Should we include the plural: "bookShelves" here as well? If I understand correctly the linter will fail if that doesn't exist as it can't go from singular --> plural

noahdietz: You're right we shouldn't have incomplete examples. Done!

rules/aip0123/resource_pattern_singular.go
noahdietz: I don't think non-nested singletons are possible since they depend on the lifecycle of another resource e.g. when X is created, XSingleton is also created. I don't know of any examples of them either.

That said, this specific line wouldn't be executed b.c the check "if no singleton" above would skip this in that case.

but having top-level singletons shouldn't be a thing...this did remind me though that I needed to better handle that case in the "is root level" check (fixed).

slevenick (approved):

PR #1406, opened 2024-07-17
feat(AIP-217): add various rules for return_partial_success support

noahdietz:
Enforces some of the guidance in https://github.com/aip-dev/google.aip.dev/pull/1385.

* makes `return_partial_success` an allowed field for Standard List APIs
* ensures `return_partial_success` is of type `bool`
* ensures `return_partial_success` is added in presence of an `unreachable` response field


777easy (approved):

rules/aip0217/return_partial_success_with_unreachable.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"github.com/googleapis/api-linter/lint"
+	"github.com/jhump/protoreflect/desc"
+)
+
+var returnPartialSuccessWithUnreachable = &lint.MethodRule{
+	Name: lint.NewRuleName(217, "return-partial-success-with-unreachable"),
+	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		return m.GetInputType().FindFieldByName("return_partial_success") != nil
```
itsStrobe, line 25: dumb question: does the `m.GetInputType()` and `m.GetOutputType()` correspond to the same verb's Req and Resp objects respectively for a given resource?

docs/rules/0132/request-unknown-fields.md
```diff
@@ -28,6 +28,7 @@ comes across any fields other than:
 - `string request_id` ([AIP-155][])
 - `google.protobuf.FieldMask read_mask` ([AIP-157][])
 - `View view` ([AIP-157][])
+- `bool return_partial_succes` ([AIP-217][])
```
itsStrobe, line 31: typo: `return_partial_success`

rules/aip0217/return_partial_success_with_unreachable.go
noahdietz: It's simpler than that. In pure protobuf descriptor protos, these are just `string` fields ([here](https://github.com/protobufjs/protobuf.js/blob/0a0cdb6b0c635d86a9bb96a382b4ff73ee848109/google/protobuf/descriptor.proto#L125-L126)) containing the fully qualified name of the types that are declared in the RPC as the request and response messages. This `protoreflect` library we use is a wrapper that fully resolves those to the actual type. So when we invoke `GetInputType` it provides the actual descriptor for the type declared as the request message for the RPC.

docs/rules/0132/request-unknown-fields.md
noahdietz: Great catch, thanks.

itsStrobe (approved):

Issue #1395, opened 2024-06-04
AIP-154: etag field_behavior


loeffel-io:
AIP-154 mentions

> **Etags on request methods**
> 
> On a request message, the etag field should be given a behavior annotation - either REQUIRED or OPTIONAL. See [AIP-203](https://google.aip.dev/203) for more information.

But there is a linter error when using REQUIRED:

```bash
- file_path: mindful/earth/user/v1/user.proto
  problems:
    - message: Delete RPCs must only require fields explicitly described in AIPs, not "etag".
      location:
        start_position:
            line_number: 282
            column_number: 3
        end_position:
            line_number: 282
            column_number: 59
        path: mindful/earth/user/v1/user.proto
      rule_id: core::0135::request-required-fields
      rule_doc_uri: https://linter.aip.dev/135/request-required-fields
```

message:

```bash
message DeleteUserRequest {
  // The resource name
  // Format: users/{user}
  string name = 1 [
    (google.api.resource_reference) = {type: "user.mindful.com/User"},
    (google.api.field_behavior) = REQUIRED,
    (validate.rules).string = {pattern: "^users\\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}"}
  ];
  // The etag of the user. Must match the server's etag.
  string etag = 2 [(google.api.field_behavior) = REQUIRED];
}
```

noahdietz:
Yep, you are definitely right. Sorry for not seeing this bug earlier, it's been a bit crazy.

Fixing in #1414 and looking at the other RPCs

noahdietz:
We had an internal bug open to look at this for Standard Get as well. I won't proceed with the change yet as there is still some debate there. This will be marked fixed when I release the ApiLinter, which will happen on Monday (no Friday releases) reviews permitting.

loeffel-io:
Thanks @noahdietz 🙏

PR #1414, opened 2024-08-08
fix(AIP-135): allow required etag in Delete

noahdietz:
Allow Standard Field `etag` to be `REQUIRED` in Standard Delete.

Updates #1395


slevenick (approved):
I strongly dislike this rule, but it does match the AIP

Issue #922, opened 2022-01-19
List Response Linter has overly aggressive regex filter


JDrit:
#### Steps to reproduce

The `Firestore.Listen` return a [`ListenResponse`](https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L696) message which represents a single message is a bi-directional stream. The [`List methods: Unknown fields (Response)`](https://linter.aip.dev/132/response-unknown-fields) linter identifies this as a list method since the regular expression [here](https://github.com/googleapis/api-linter/blob/6167aaa7cdb810a7efb4c6544cdfd71d0fe4eb79/rules/aip0132/aip0132.go#L50) matches against **List**enResponse.

This should hopefully just be a simple regex fix.

noahdietz:
Hey there, I'm going through old issues and cleaning thins up. You are most definitely correct. I've proposed a fix in #1420.

Once that is released (and imported for Google), you shouldn't see it again. Thanks for the report!

noahdietz:
Released! Importing for Google use imminently

PR #1420, opened 2024-08-27
fix(AIP-132): refine List request response regex

noahdietz:
Updates all `List` request/response regular expressions to expect a capital letter following `"List"` so as to avoid capturing verbs that start with `"List"`, but which are not `List`, like `"Listen"`.

This is a reasonable refinement to make because all standard Lists are `List{PluralNoun}` so we get `List{PluralNoun}Request` and `List{PluralNoun}Response`. 

Updates #922


slevenick (approved):
Should we do the same for the other regexes?

noahdietz:
> Should we do the same for the other regexes?

Yep, definitely.

Issue #1430, opened 2024-09-17
locations.go sourceInfoRegistry cache is not concurrent


apodznoev:
We found the `locations.go` a useful utility to access locations of HTTP options, in particular: `MethodHTTPRule`.

However, it is not concurrent and might crash like:

```
fatal error: concurrent map writes

goroutine 5017 [running]:
github.com/googleapis/api-linter/locations.sourceInfoRegistryType.sourceInfo(0x14000172e70, 0x140052a1ea0)
        github.com/googleapis/api-linter@v1.64.0/locations/locations.go:86 +0x144
github.com/googleapis/api-linter/locations.pathLocation({0x1022ba130, 0x140052af6c0}, {0x14001072b18, 0x2, 0x14001072b58?})
        github.com/googleapis/api-linter@v1.64.0/locations/locations.go:37 +0xfc
github.com/googleapis/api-linter/locations.FieldResourceReference(0x140052af6c0)
        github.com/googleapis/api-linter@v1.64.0/locations/field_locations.go:35 +0x68
```
With goroutine:
```
goroutine 5024 [runnable]:
github.com/googleapis/api-linter/locations.sourceInfoRegistryType.sourceInfo(0x14000172e70, 0x140039f5180)
        github.com/googleapis/api-linter@v1.64.0/locations/locations.go:81 +0x104
github.com/googleapis/api-linter/locations.pathLocation({0x1022ba090, 0x140037cf8b0}, {0x14004b87af8, 0x2, 0x2?})
        github.com/googleapis/api-linter@v1.64.0/locations/locations.go:37 +0xfc
github.com/googleapis/api-linter/locations.MethodOption(...)
        github.com/googleapis/api-linter@v1.64.0/locations/method_locations.go:54
github.com/googleapis/api-linter/locations.MethodHTTPRule(0x140037cf8b0)
        github.com/googleapis/api-linter@v1.64.0/locations/method_locations.go:37 +0x68
```
or:
```
goroutine 4853 [runnable]:
github.com/jhump/protoreflect/desc/internal.asMapKey(...)
        github.com/jhump/protoreflect@v1.15.6/desc/internal/source_info.go:52
github.com/jhump/protoreflect/desc/internal.SourceInfoMap.Get(0xc002394ae0, {0xc0010bd810, 0x4, 0x5bc2d70?})
        github.com/jhump/protoreflect@v1.15.6/desc/internal/source_info.go:14 +0x3f
[...]

PR #1433, opened 2024-09-23
fix(locations): make source info access concurrent safe

noahdietz:
Refactors internal source code info registry for the `locations` package to leverage a `synx.Mutex` in protecting the `map`s containing the descriptor/path to SourceCodeInfo pairs. Uses the "[mutex hat](https://github.com/go-critic/go-critic/issues/465)" pattern for naming/commenting.

Adds a simple concurrency test to run with `-race`. Note: Running this same test on `main` produced the data race warning.

Executes unit tests in CI with the `-race` flag.

Updates #1430

### DIFF
--- a/locations/locations.go
+++ b/locations/locations.go
@@ -23,6 +23,8 @@
 package locations
 
 import (
+	"sync"
+
 	"github.com/jhump/protoreflect/desc"
 	dpb "google.golang.org/protobuf/types/descriptorpb"
 )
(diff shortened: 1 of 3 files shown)


locations/locations.go
```diff
@@ -37,12 +39,24 @@ func pathLocation(d desc.Descriptor, path ...int) *dpb.SourceCodeInfo_Location {
 	return sourceInfoRegistry.sourceInfo(d.GetFile()).findLocation(fullPath)
 }
 
-type sourceInfo map[string]*dpb.SourceCodeInfo_Location
+type sourceInfo struct {
+	info map[string]*dpb.SourceCodeInfo_Location
+	lock sync.Mutex
```
quartzmo, line 44: The "mutex hat" pattern prefers the mutex above, and with a name that indicates what it controls: https://github.com/go-critic/go-critic/issues/465

noahdietz: Nice! Done.

locations/locations.go
```diff
@@ -37,12 +39,25 @@ func pathLocation(d desc.Descriptor, path ...int) *dpb.SourceCodeInfo_Location {
...
+func newSourceInfo() *sourceInfo {
+	return &sourceInfo{
+		info: map[string]*dpb.SourceCodeInfo_Location{},
+	}
+}
 
 // findLocation returns the Location for a given path.
-func (si sourceInfo) findLocation(path []int32) *dpb.SourceCodeInfo_Location {
+func (si *sourceInfo) findLocation(path []int32) *dpb.SourceCodeInfo_Location {
+	si.infoMu.Lock()
```
slevenick, line 56: Do we need to lock on the read of this map? It seems like we will only write the contents of the info map once per key, and I assume we only try to access it after writing, so there shouldn't be a need to lock on read

I could only see a lock being needed if the call to findLocation can happen while another call to sourceInfo() is currently executing and writing the value for that location

Not that it matters too much to have an extra lock call 🤷‍♀️

slevenick (approved):

locations/locations.go
quartzmo: I think tooling such as the test `-race` flag might complain without the read lock, but I guess you can remove it to see if that happens.

noahdietz: > I could only see a lock being needed if the call to findLocation can happen while another call to sourceInfo() is currently executing and writing the value for that location
>
> Not that it matters too much to have an extra lock call 🤷‍♀️

Yeah that's what I'm being cautious about - i'd rather just synchronize all access. It's not performant enough of an application to really care about the extra lock call and I'm choosing to go one step short of using `sync.Map` (which is not strongly typed) :)

noahdietz: Thanks @slevenick for the point/question though

PR #1436, opened 2024-10-18
chore: upgrade go version to 1.21 and bump bufbuild/protocompile to v0.14.1 

r9deyes:
To use api-linter with protocolbuffers/protobuf v28 need to update `bufbuild/protocompile` (from v0.10.0 to v0.14.1).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.20"
+          go-version-file: 'go.mod'
       - run: go test -race ./...
   lint:
     runs-on: ubuntu-latest
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.20"
+          go-version-file: 'go.mod'
       - name: staticcheck
         uses: dominikh/staticcheck-action@v1.3.0
         with:
@@ -30,6 +30,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.20"
+          go-version-file: 'go.mod'
       - name: Run the quality checker (which catches obvious mistakes, missing docs, etc.)
         run: go run ./.github/quality-checker
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/googleapis/api-linter
 
-go 1.20
+go 1.21
 
 require (
 	bitbucket.org/creachadair/stringset v0.0.12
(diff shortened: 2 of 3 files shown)


noahdietz:
Hey @r9deyes thanks for the PR. We are going to need to also update the Go verison used in our workflows, like to use `go-version-file: 'go.mod'` instead of hard coding a version, like [here](https://github.com/googleapis/api-linter/blob/6ef36a69c0a9ff876e644922b413ea0999891c05/.github/workflows/ci.yaml#L14)

noahdietz:
Looks like we will also need to suppress staticcheck here b.c it doesn't like that `jhump/protoreflect/desc` is deprecated...

r9deyes:
I see a few ways to resolve the deprecation issue in jhump/protoreflect:

1. Update only `bufbuild/protocompile` to v1.14.1 and keep `jhump/protoreflect` at the stable version v1.16.0. This allows to use protobuf v28 as desired.
2. We can simply add `//lint:ignore SA10019` above every import of the deprecated package, so staticcheck will ignore it.
3. Refactor all usages of `jhump/protoreflect/desc`. There are more than 100 files that depend on it. Refactoring internal packages and all rules will take time. Additionally, the new version of `jhump/protoreflect/v2` could facilitate this, but currently, v2 is still in beta ([v2.0.0-beta.2](https://github.com/jhump/protoreflect/tree/v2.0.0-beta.2)).
Personally, I prefer the first option: just a local update of `protocompile`. In future releases, we can refactor to use `jhump/protoreflect/v2` once it is stable.

noahdietz:
> Update only bufbuild/protocompile to v1.14.1 and keep jhump/protoreflect at the stable version v1.16.0. This allows to use protobuf v28 as desired.
> Personally, I prefer the first option: just a local update of protocompile. 

Yes, I agree, let's get you unblocked.

> In future releases, we can refactor to use jhump/protoreflect/v2 once it is stable.

We may need to bump to v1.17.0 sooner though, for Protobuf Editions support. v1.16.0 states it does not have `desc/protoparse` support for editions syntax, which is used for api-linter's `rules/internal/testutils` parsing proto file strings in tests...

It's almost looking like we should migrate *off* of jhump/protoreflect and use protobuf-go (in place of `desc`, as jhump/protoreflect suggests) + protocompile (in place of `jhump/protoreflect/desc/protoparse` also as is suggested) long term.

Since the Rule interfaces are public, we will have to take a v2 of this project to make this happen, so we better think good and long about what we want to do (which I have been in the background, but perhaps it should be done publicly :) )

noahdietz (approved):
Thank you @r9deyes

noahdietz:
Latest release v1.64.7 contains this change @r9deyes !

r9deyes:
Excellent! Thanks for your assistance @noahdietz

PR #1439, opened 2024-10-22
fix(AIP-133): use resource field in signature suggestion

noahdietz:
Fix the way that the expected resource field name is derived for Standard Create method_signatures. When the resource field in the request message uses a name that doesn't exactly mirror the resource type/RPC name, the previous logic suggested a method_signature including a field that didn't exist. Since this rule isn't supposed to also govern the field's name itself, just roll with it.

Also fixes the way the output type is evaluated for the resource parent/no-parent case when the response type is an Operation.

Moves the LRO test to its own case.

Addresses internal feedback http://b/372508475


rules/aip0133/method_signature_test.go
```diff
@@ -75,7 +75,12 @@ func TestMethodSignature(t *testing.T) {
 					Book book = 2;
 					{{.IDField}}
 				}
-				message Book {}
+				message Book {
+				  option (google.api.resource) = {
+				    type: "library.googleapis.com/Book"
+					pattern: "libraries/{library}/books/{book}"
```
slevenick, line 81:
```suggestion
					  pattern: "libraries/{library}/books/{book}"
```

slevenick (approved):

Issue #1285, opened 2023-11-08
AIP 132: HTTP URI should include a `parent` variable 


loeffel-io:
Version: 1.59.1

Given this method

```protobuf
  rpc ListMediaTypes(ListMediaTypesRequest) returns (ListMediaTypesResponse) {
    option (google.api.http) = {
      get: "/v1/mediaTypes",
      additional_bindings: [
        {
          get: "/v1/{parent=contents/*}/mediaTypes"
        },
        {
          get: "/v1/{parent=contentGroups/*}/mediaTypes"
        }
      ],
    };
    option (google.api.method_signature) = "parent";
  }
```

the aip-linter throws `message: HTTP URI should include a parent variable` while at [AIP-127](https://google.aip.dev/127) under _Multiple URI bindings_ this looks allowed or here https://github.com/googleapis/googleapis/blob/master/google/cloud/resourcemanager/v3/tag_keys.proto#L45C19-L45C37

is this expected? Thank you 🙏

loeffel-io:
ping @noahdietz 🙏

noahdietz:
> this looks allowed or here
> https://github.com/googleapis/googleapis/blob/master/google/cloud/resourcemanager/v3/tag_keys.proto#L45C19-L45C37
> 
> is this expected? Thank you 🙏

In looking at the internal changes for this example, it looks like it predates the linter. If this file were linted today it would like produce a finding.

I think in general the linter does not handle "top level" resources all that well, since in Google Cloud, there are very few examples (everything is parented by a Project, Org, Folder, etc.). It is also not super common to have a resource that is both top-level and potentially parented by another resource. For these reasons, exempting a single RPC via the disablement comment or config is preferable to relaxing the implementation and potentially missing out on a real finding.

Can you share more of the example/case you have i.e., the request message, resource definition?

One solution would be to perhaps tweak this rule to be more relaxed if the `string parent` field is not `google.api.field_behavior = REQUIRED` - which it is for Standard List in all cases except for those mentioned above.

That said, if one doesn't want the `parent` field to be required, but still wants an additional binding with `parent`, what will warn them if that is missing? That's the tricky part...

loeffel-io:
Yes, its pretty tricky! Thank you for your detailed answer 

Here is the full proto file for now: 

```proto
syntax = "proto3";

package mindful.earth.entitlement.v1;

[...]

PR #1463, opened 2025-01-30
add buf plugin

### LINKED ISSUE #1095: Enable api-linter on buf
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

**Is your feature request related to a problem? Please describe.**
The API linter is a useful way of ensuring API protobuf definitions are AIP-compliant. [buf](buf.build) is gradually becoming an industry standard for publishing and managing protobufs, but today: 1) There is no way of natively running aip-linter in buf, and 2) The buf default lint rules contradict AIP lint rules. For example, buf recommends that no two RPC methods return the same message, but AIP recommends `Get` and `Create` methods both return the same message.

**Describe the solution you'd like**
Native aip-linter integration into buf as a plugin.

**Describe alternatives you've considered**
Run aip-linter entirely separately, or don't use it at all, if using buf.

**Additional context**
N/A

seanlafferty-ibm:
Closes #1095 

As mentioned in the linked issue, buf has become the predominant protobuf formatting/linting/building tool, and now that buf supports linter plugins, we can surface api-linter rules through buf. This opens up a rich ecosystem of integrations. For example, we gain IDE support:

Screenshot 2025-01-29 at 6 21 08 PM

The implementation is fairly trivial, we loop over all api-linter `Rules`, and convert them to buf `Rules`. This is validated by importing buf's plugin testing helpers.


noahdietz:
Hey @seanlafferty-ibm , unfortunately we won't be able to accept this contribution. The primary concern is the increase in support surface for this project with an ecosystem we aren't ready to get involved with yet. I hope you and the others can understand that, and I really appreciate you taking the time to write this up.

Now, if Buf or another entity, like IBM, is able/willing to host a plugin implementation, we'd be happy to advertise that here in the repo and on the api-linter docsite. I'll mention as much on the related issue.

Thanks again and apologies that it won't work out. Of course, I still really appreciate your input/thoughts on the core functionality of the api-linter e.g. in #1285, and I hope this won't dissuade you from future collaboration.

seanlafferty-ibm:
Sure, thanks for the consideration. If we publish a plugin independently, we can comment on the original issue

Issue #1018, opened 2022-09-16
AIP-215: warning on cross-API references


jskeet:
AIP-215 states:

> APIs should not directly depend on protos defined in other APIs. Instead, they should copy and paste the applicable messages into their own API. When doing so, APIs should keep the field names and numbers the same.

We should lint for this, probably by checking imports. Some caveats to this:

- Various common imports should be allowed, e.g. `google/protobuf/*`, `google/longrunning/*`, `google/rpc/*`
- Only imports that remain after pruning (e.g. removing internal-only dependencies) should generate warnings
- This should probably be done at an "API version" boundary - so `abc/def/v2/xyz/a.proto` should be able to import `abc/def/v2/b.proto` and vice versa; unversioned APIs are sufficiently rare that we can probably either ignore them or restrict them to the exact same package
- Very occasionally there are legitimate reasons for an API to use messages from another API; we should be able to configure those "allowed imports" rather than those APIs having to turn the rule off entirely. (They should be able to be confident they're not *accidentally* importing more than they intended to.)

noahdietz:
After https://github.com/aip-dev/google.aip.dev/pull/1117, this guidance was moved to AIP-215, renaming this issue and updating original text accordingly.

noahdietz:
@shollyman take a look at some of Jon's thoughts here, I think they will help make an initial version more targeted and less error prone.

shollyman:
Thanks for tagging me.  I was originally drawn to validating the import list as well, but the biggest wart there is annotations/extensions.  Framework annotations, data lifecycle annotations, etc can be valid, optional metadata present for a subset of consumers, but they pollute the import list in such a way that I worry import validation alone will yield many false positives.

I've taken the approach in https://github.com/googleapis/api-linter/pull/1467 to compare fielddescriptors and message types they reference.  Arguably you could extend this and do the same for comparing MessageDescriptors that reference request/response types in other packages, but most of the AIP-215 violations I've seen anecdotally come in via message references.

PR #1467, opened 2025-02-04
feat(AIP-215): augment foreign type checking

shollyman:
This PR augments AIP-215 rules to look for package mismatches.  It leverages comparisons between a FieldDescriptor that references a Message type, and compares based on the proto package of the two.

Updates: #1018

internal: b/391975695

### DIFF
--- a/docs/rules/0215/foreign-type-reference.md
+++ b/docs/rules/0215/foreign-type-reference.md
@@ -3,9 +3,9 @@ rule:
   aip: 215
   name: [core, '0215', foreign-type-reference]
   summary: API-specific protos should be in versioned packages.
-permalink: /215/versioned-packages
+permalink: /215/foreign-type-reference
 redirect_from:
-  - /0215/versioned-packages
+  - /0215/foreign-type-reference
 ---
 
 # Versioned packages
@@ -47,8 +47,18 @@ message SomeMessage {
 
 ## Known issues
 
-This package disallows subpackaging, and allows any package that ends with `.type` to
-avoid flagging possible legitimate uses of AIP-213 component packages.
+This check only allows subpackage usage if you're within a versioned path, but generates warnings for unversioned subpackage usage.
+It also ignores if the referenced package is a "common" package like google.api, or if the package path ends in ".type" indicating
+the package is an AIP-213 component package.
+
+Examples of foreign type references and their expected results:
+
+| Calling Package | Referenced Package | Result       |
+| --------------- | ------------------ | ------------ |
+| foo.bar         | foo.xyz            | lint warning |
+| foo.v2.bar      | foo.v2.xyz         | ok           |
+| foo.bar         | foo.type           | ok           |
+| foo.bar         | google.api         | ok           |
 
 ## Disabling
 
--- a/docs/rules/0215/foreign-type-reference.md
+++ b/docs/rules/0215/foreign-type-reference.md
@@ -47,7 +47,7 @@ message SomeMessage {
 
 ## Known issues
 
-This check only allows subpackage usage if you're within a versioned path, but generates warnings for unversioned subpackage usage.
+This check only allows subpackage usage within a versioned path, but generates warnings for unversioned subpackage usage.
 It also ignores if the referenced package is a "common" package like google.api, or if the package path ends in ".type" indicating
 the package is an AIP-213 component package.
 
--- a/docs/rules/0215/foreign-type-reference.md
+++ b/docs/rules/0215/foreign-type-reference.md
@@ -48,7 +48,7 @@ message SomeMessage {
 ## Known issues
 
 This check only allows subpackage usage within a versioned path, but generates warnings for unversioned subpackage usage.
-It also ignores if the referenced package is a "common" package like google.api, or if the package path ends in ".type" indicating
+It also ignores if the referenced package is a "common" package like `google.api`, or if the package path ends in `.type` indicating
 the package is an AIP-213 component package.
 
 Examples of foreign type references and their expected results:
--- a/docs/rules/0215/foreign-type-reference.md
+++ b/docs/rules/0215/foreign-type-reference.md
@@ -2,7 +2,7 @@
 rule:
   aip: 215
   name: [core, '0215', foreign-type-reference]
-  summary: API-specific protos should be in versioned packages.
+  summary: API should not reference foreign types outside of the API scope.
 permalink: /215/foreign-type-reference
 redirect_from:
   - /0215/foreign-type-reference
--- a/docs/rules/0215/foreign-type-reference.md
+++ b/docs/rules/0215/foreign-type-reference.md
@@ -1,0 +1,65 @@
+---
+rule:
+  aip: 215
+  name: [core, '0215', foreign-type-reference]
+  summary: API-specific protos should be in versioned packages.
+permalink: /215/versioned-packages
+redirect_from:
+  - /0215/versioned-packages
+---
+
+# Versioned packages
+
+This rule enforces that none of the fields in an API reference message types in a different
+proto package namespace other than well-known common packages.
+
+## Details
+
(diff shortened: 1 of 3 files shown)


docs/rules/0215/foreign-type-reference.md
```diff
@@ -0,0 +1,65 @@
+---
+rule:
+  aip: 215
+  name: [core, '0215', foreign-type-reference]
+  summary: API-specific protos should be in versioned packages.
```
noahdietz, line 5: Please update this

docs/rules/0215/foreign-type-reference.md
```diff
@@ -0,0 +1,65 @@
...
+permalink: /215/versioned-packages
+redirect_from:
+  - /0215/versioned-packages
```
noahdietz, line 8: Please update these

rules/aip0215/foreign_type_reference.go
```diff
@@ -0,0 +1,61 @@
...
+		return f.GetPackage()
+	}
+	return ""
+}
+
+// check if a package path is an AIP-213 component package path.
+// Valid component packages are expected to end in ".type".
+func isComponentPackage(pkg string) bool {
+	parts := strings.Split(pkg, ".")
+	return parts[len(parts)-1] == "type"
```
noahdietz, line 60: Nit: could be `return strings.HasSuffix(pkg, ".type")` no?

rules/aip0215/foreign_type_reference.go
```diff
@@ -0,0 +1,61 @@
...
+			msgPkg := getPackage(msg)
+			if !utils.IsCommonProto(fd.GetMessageType().GetFile()) {
+				if curPkg != "" && msgPkg != "" && !isComponentPackage(msgPkg) && curPkg != msgPkg {
+					return []lint.Problem{{
+						Message:    fmt.Sprintf("foreign type referenced, current field in %q message in %q", curPkg, msgPkg),
+						Descriptor: fd,
+					}}
+				}
+			}
+		}
```
noahdietz, line 44: I'm kind of a dedent, check assertions, exit early kind of person. How do you feel about this approach? Could be further collapsed into related &&'d conditions I suppose, but I just find a deeply nested conditional hard to grok.

I think the suggestion below is on the extreme end, but perhaps we could check some of the already deeply nested conditions earlier e.g. `if curPkg == "" { return nil }` instead of two levels deeper?

```suggestion
    curPkg := getPackage(fd)
    if curPkg == "" {
      return nil
    }
    msg := fd.GetMessageType()
    if msg == nil {
      return nil
    }
    msgPkg := getPackage(msg)
    if msgPkg == "" {
      return nil
    }
    if curPkg == msgPkg {
      return nil
    }
		if utils.IsCommonProto(fd.GetMessageType().GetFile()) {
	    return nil
		}
		if isComponentPackage(msgPkg) {
			return nil
		}
		return []lint.Problem{{
		  Message:    fmt.Sprintf("foreign type referenced, current field in %q message in %q", curPkg, msgPkg),
		  Descriptor: fd,
	  }}
```

docs/rules/0215/foreign-type-reference.md
```diff
@@ -0,0 +1,65 @@
...
+import "other.proto";
+
+message SomeMessage {
+    OtherMessage other_message = 1;
+}
+```
+
+## Known issues
+
+This package disallows subpackaging, and allows any package that ends with `.type` to
```
noahdietz, line 50: > This package disallows subpackaging,

Is this to say that, related to one of Jon's points on #1018 (reproduced below), that `google.library.v1` cannot depend on `google.library.v1.subpkg.Foo` ?

We have some examples of this today that we'd need to support and I think it is still considered an allowable mechanism because it is ostensibly part of the same version.

Jon's relevant comment:
> This should probably be done at an "API version" boundary - so abc/def/v2/xyz/a.proto should be able to import abc/def/v2/b.proto and vice versa

noahdietz (commented):
Generally, LGTM, but let's continue talking about the heuristics a bit

docs/rules/0215/foreign-type-reference.md
shollyman: The versioned_packages rule already present in the 215 rules defines this regexp:
```
var version = regexp.MustCompile(`\.v[\d]+(p[\d]+)?(alpha|beta|eap|test)?[\d]*`)
```

I think if we normalized the package(s) to the last "version" segment if present we're closer to the intent.  It would mean packages within a version would be allowed, but subpackages without a version segment would be disallowed:

* `foo.v2.a` references `foo.v2.b` -> allowed
* `foo.a` references `foo.b` -> lint warning generated

noahdietz: Yeah that makes sense to me!

rules/aip0215/foreign_type_reference.go
shollyman: inlined as part of the simplification.

rules/aip0215/foreign_type_reference.go
shollyman: Refactored this to align with this style.

docs/rules/0215/foreign-type-reference.md
shollyman: Done.

docs/rules/0215/foreign-type-reference.md
shollyman: Done.

docs/rules/0215/foreign-type-reference.md
shollyman: Updated the PR with the change.  I did end up defining a slightly different regex to capture the prefix in one shot, but it does yield an ambiguous case:  is the versioning segment allowed to be the first segment of a package path?

e.g. while `foo.v1` is an allowed path (and thus something like a reference to `foo.v1.bar` is allowed), is `v2` considered a version path if its the only segment, or is it just a package name?

docs/rules/0215/foreign-type-reference.md
```diff
@@ -47,8 +47,18 @@ message SomeMessage {
 
 ## Known issues
 
-This package disallows subpackaging, and allows any package that ends with `.type` to
-avoid flagging possible legitimate uses of AIP-213 component packages.
+This check only allows subpackage usage if you're within a versioned path, but generates warnings for unversioned subpackage usage.
```
noahdietz, line 50:
```suggestion
This check only allows subpackage usage within a versioned path, but generates warnings for unversioned subpackage usage.
```

Take the "you" out of it

docs/rules/0215/foreign-type-reference.md
```diff
@@ -47,8 +47,18 @@ message SomeMessage {
...
+It also ignores if the referenced package is a "common" package like google.api, or if the package path ends in ".type" indicating
```
noahdietz, line 51:
```suggestion
It also ignores if the referenced package is a "common" package like `google.api`, or if the package path ends in `.type` indicating
```

docs/rules/0215/foreign-type-reference.md
```diff
@@ -47,8 +47,18 @@ message SomeMessage {
...
+the package is an AIP-213 component package.
+
+Examples of foreign type references and their expected results:
+
+| Calling Package | Referenced Package | Result       |
```
noahdietz, line 56: I'll be honest, IDK how this will render on the linter site, but let's find out!

noahdietz (approved):
Nice, looking really good. Love the examples you gave.

docs/rules/0215/foreign-type-reference.md
noahdietz: > but it does yield an ambiguous case: is the versioning segment allowed to be the first segment of a package path?

This won't realistically happen for Google, but I think the way you've done it would be OK for others

PR #1472, opened 2025-02-14
docs(AIP-215): fix incorrect heading for `foreign-type-reference`

Juneezee:
This appears to be a copy-paste mistake in PR https://github.com/googleapis/api-linter/pull/1467.


noahdietz (approved):
Indeed it was thank you! Amazing how easy it is to miss those things sometimes...

Issue #1465, opened 2025-01-31
unexpected lint warning when providing mutiple files


xinnjie:
#### Environment details

```
$ api-linter --version                                                                                                                                                                                  [22:41:48]
api-linter 1.68.0

$ tree proto                                                                                                                                                                                            [22:44:30]
proto
├── google
│   └── api
│       ├── annotations.proto
│       └── http.proto
└── ....
```

#### Steps to reproduce

1. `api-linter` works great when only provided with one file
```
$ api-linter -I proto proto/google/api/http.proto                                                                                                                                                       [22:45:42]
- file_path: proto/google/api/http.proto
  problems:
    - message: 'Proto package and directory structure mismatch: The proto package must match the proto directory structure.'
      location:
        start_position:
            line_number: 17
            column_number: 1
        end_position:
            line_number: 17
            column_number: 19
        path: proto/google/api/http.proto
      rule_id: core::0191::proto-package
      rule_doc_uri: https://linter.aip.dev/191/proto-package
    ....
```

```
$ api-linter -I proto proto/google/api/annotations.proto                                                                                                                                                [22:47:37]
- file_path: proto/google/api/annotations.proto
  problems:
    - message: 'Proto package and directory structure mismatch: The proto package must match the proto directory structure.'
      location:
        start_position:
            line_number: 17
            column_number: 1
        end_position:
            line_number: 17
            column_number: 19
        path: proto/google/api/annotations.proto
      rule_id: core::0191::proto-package
      rule_doc_uri: https://linter.aip.dev/191/proto-package
```
3. Having unexpected lint warning when providing mutiple files
```
$ api-linter -I proto proto/google/api/http.proto proto/google/api/annotations.proto                                                                                                                    [22:44:53]
2025/01/31 22:45:42 google/api/http.proto:29:9: symbol "google.api.Http" already defined at proto/google/api/http.proto:29:9
google/api/http.proto:33:21: symbol "google.api.Http.rules" already defined at proto/google/api/http.proto:33:21
google/api/http.proto:41:8: symbol "google.api.Http.fully_decode_reserved_expansion" already defined at proto/google/api/http.proto:41:8
google/api/http.proto:308:9: symbol "google.api.HttpRule" already defined at proto/google/api/http.proto:308:9
google/api/http.proto:313:10: symbol "google.api.HttpRule.selector" already defined at proto/google/api/http.proto:313:10
[...]

PR #1493, opened 2025-06-26
feat(integration-tests): add test harness for cli invocation

quirogas:
## Create TestMultipleFilesFromParentDir Integration Test

`TestMultipleFilesFromParentDir`  emulates the issue described in https://github.com/googleapis/api-linter/issues/1465. This change ensures our integration test more robustly covers cases with nested package structures and proper import path resolution from the project root.

### Directory structure

```
.
└── grandparent
    └── parent
        ├── a.proto
        └── b.proto
```

### Command:
```
api-linter -I grandparent grandparent/parent/a.proto grandparent/parent/b.proto
```

### DIFF
--- a/cmd/api-linter/integration_test.go
+++ b/cmd/api-linter/integration_test.go
@@ -183,7 +183,7 @@ func TestMultipleFilesFromParentDir(t *testing.T) {
 	// https://github.com/googleapis/api-linter/issues/1465
 
 	// Skipping until Issue # 1465 is addressed
-	t.Skip()
+	t.Skip("Skipping until Issue # 1465 is addressed")
 
 	projDir, err := os.MkdirTemp("", "proj")
 	if err != nil {


cmd/api-linter/integration_test.go
```diff
@@ -178,6 +178,79 @@ func TestBuildErrors(t *testing.T) {
...
+	if err := os.Chdir(projDir); err != nil {
+		t.Fatal(err)
+	}
+	defer os.Chdir(oldWD)
+
+	args := []string{
+		"-I", "proto",
+		"proto/a/a.proto",
+		"proto/b/b.proto",
+	}
```
noahdietz: This is a good case to have, but the exact case in the issue involved two protos in the same package i.e. `proto/a/a.proto` and `proto/a/b.proto` each with `package a;`. This example _should_ still exercise the same situation but maybe a second test for this would be reasonable as well?

noahdietz (commented):
Great job with all of the temp file stuff.

quirogas:
I think we should re-open https://github.com/googleapis/api-linter/issues/1465 too.

noahdietz:
> I think we should re-open #1465 too.

Agreed

quirogas:
Skipping test for now, since the bug is still present.

cmd/api-linter/integration_test.go
```diff
@@ -178,6 +178,73 @@ func TestBuildErrors(t *testing.T) {
 	}
 }
 
+func TestMultipleFilesFromParentDir(t *testing.T) {
+	// This test addresses a previously found bug:
+	// https://github.com/googleapis/api-linter/issues/1465
+
+	// Skipping until Issue # 1465 is addressed
+	t.Skip()
```
noahdietz: Tip: You can put this comment in the `Skip` and it will emit it in the test output
```suggestion
	t.Skip("Skipping until Issue # 1465 is addressed")
```

https://pkg.go.dev/testing#hdr-Skipping

noahdietz (approved):

PR #1496, opened 2025-07-02
fix(cli): unexpected lint warning when providing multiple files

quirogas:
## Bug 

When running `api-linter` with the `-I` flag for an import path that is a parent of the proto files, the linter would fail with a "symbol already defined" error. This was caused by a two-stage file resolution process that resolved the same file twice with different paths. Additionally, the linter would fail when using absolute file paths without also passing `.` as an import path.

For example, the following commands would fail:

```sh
api-linter -I grandparent grandparent/parent/a.proto grandparent/parent/b.proto
```

```sh
api-linter /Users/user/home/project/grandparent/parent/a.proto
```

-----

## Fix 

Fixes \#1465

The file path resolution logic in `cmd/api-linter/cli.go` has been refactored to use a single, consistent resolution step. We now resolve imports **before** passing them to the proto parser. This ensures that all proto files are resolved with the same set of import paths that the parser uses, preventing duplicate parsing and fixing the bugs described above.

### DIFF
--- a/cmd/api-linter/cli.go
+++ b/cmd/api-linter/cli.go
@@ -296,9 +296,9 @@ func getOutputFormatFunc(formatType string) formatFunc {
 	return yaml.Marshal
 }
 
-func resolveImports(protoImportPaths []string) []string {
+func resolveImports(imports []string) []string {
 	// If no import paths are provided, default to the current directory.
-	if len(protoImportPaths) == 0 {
+	if len(imports) == 0 {
 		return []string{"."}
 	}
 
@@ -307,10 +307,10 @@ func resolveImports(protoImportPaths []string) []string {
 	if err != nil {
 		// Fallback: If we can't get CWD, return only the provided paths and "."
 		seen := map[string]bool{
-		  ".": true,
+			".": true,
 		}
 		result := []string{"."} // Always include "."
-		for _, p := range protoImportPaths {
+		for _, p := range imports {
 			if !seen[p] {
 				seen[p] = true
 				result = append(result, p)
@@ -330,10 +330,10 @@ func resolveImports(protoImportPaths []string) []string {
 	// Initialize resolvedImports with "." and track its canonical absolute path.
 	resolvedImports := []string{"."}
 	seenAbsolutePaths := map[string]bool{
-	  evaluatedCwd: true, // Mark canonical CWD as seen
+		evaluatedCwd: true, // Mark canonical CWD as seen
 	}
 
-	for _, p := range protoImportPaths {
+	for _, p := range imports {
 		absPath, err := filepath.Abs(p)
 		if err != nil {
 			// If we can't get the absolute path, treat it as an external path


cmd/api-linter/cli.go
```diff
@@ -165,21 +165,7 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
...
-			return err
-		}
-	}
-	// Then resolve again against ".", the local directory.
-	// This is necessary because ResolveFilenames won't resolve a path if it
-	// relative to *at least one* of the given import paths, which can result
-	// in duplicate file parsing and compilation errors, as seen in #1465 and
-	// #1471. So we resolve against local (default) and flag specified import
-	// paths separately.
-	protoFiles, err = protoparse.ResolveFilenames([]string{"."}, protoFiles...)
```
noahdietz, line 182: I _think_ not including `"."`  in the `ResolveFilenames` call is actually what caused the issue described in https://github.com/googleapis/api-linter/issues/1465#issuecomment-2667709003 where the user had to include `--proto-path=.` when using in combination with `--descriptor-set-in` because this is what I tried originally in #1466

Do we have a test runCli test using `--descriptor-set-in`?

quirogas: We do have test using `--descriptor-set-in`. However, I'm going to look into replicating the exact behaviour in the comment  and write a test for it .

quirogas: After some digging I found that the reason the `--descriptor-set-in` was failing had nothing to do with the flag itself, but the way `ImportPahts` were resolved. While the previous state should've been able to handle the  error, it was hard to say for sure (since the error is hard to replicate). However, adding the `resolveImports` function should allow us to handle both cases safely.

noahdietz: Interesting! Can I ask how you tested this to verify no `--descriptor-set-in` regression? Perhaps we should add a `--descriptor-set-in` test with no `--proto-path` flag to the integration tests? I didn't see one on first glance, just an "error" test that exercises the flag.

cmd/api-linter/cli.go
```diff
@@ -305,3 +295,74 @@ func getOutputFormatFunc(formatType string) formatFunc {
...
+		return []string{"."}
+	}
+
+	// Get the absolute path of the current working directory.
+	cwd, err := os.Getwd()
+	if err != nil {
+		// Fallback: If we can't get CWD, return only the provided paths and "."
+		seen := make(map[string]struct{})
+		var result []string
+		result = append(result, ".") // Always include "."
```
noahdietz: consolidate this declaration and initialization into one line composite literal
```suggestion
		result := []string{"."} // Always include "."
```

https://google.github.io/styleguide/go/best-practices.html#composite-literals

cmd/api-linter/cli.go
```diff
@@ -305,3 +295,74 @@ func getOutputFormatFunc(formatType string) formatFunc {
...
+	// If no import paths are provided, default to the current directory.
+	if len(protoImportPaths) == 0 {
+		return []string{"."}
+	}
+
+	// Get the absolute path of the current working directory.
+	cwd, err := os.Getwd()
+	if err != nil {
+		// Fallback: If we can't get CWD, return only the provided paths and "."
+		seen := make(map[string]struct{})
```
noahdietz: If we use a `map[string]bool` then the subsequent `if` condition becomes simpler due to the default value returned if the key isn't present being `false` for `bool` https://go.dev/play/p/4gmEsbyfY8f

```go
seen := make(map[string]bool)
...
if !seen[p] {
  seen[p] = true
  ...
}
```

This ultimately makes the code read better in the condition and in the set line in comparison to `seen[p]= struct{}{}`

cmd/api-linter/cli.go
```diff
@@ -305,3 +295,74 @@ func getOutputFormatFunc(formatType string) formatFunc {
...
+	evaluatedCwd, err := filepath.EvalSymlinks(cwd)
+	if err != nil {
+		// Fallback to Clean if EvalSymlinks fails (e.g., path does not exist)
+		evaluatedCwd = filepath.Clean(cwd)
+	}
+
+	// Initialize resolvedImports with "." and track its canonical absolute path.
+	resolvedImports := []string{"."}
+	seenAbsolutePaths := make(map[string]struct{})
+	seenAbsolutePaths[evaluatedCwd] = struct{}{} // Mark canonical CWD as seen
```
noahdietz: Same comment re:using `bool` and these can also be combined into a composite literal initialization
```suggestion
	seenAbsolutePaths := map[string]bool{
	  evaluatedCwd: true, // Mark canonical CWD as seen
	}
```

cmd/api-linter/cli_test.go
```diff
@@ -54,3 +57,141 @@ func TestNewCli(t *testing.T) {
...
+			protoImportPaths: []string{"./relative_dir", "/tmp/external_dir", "test_dir"},
+			setupTempDir: func(t *testing.T) string {
+				tmpDir := t.TempDir()
+				if err := os.Mkdir(filepath.Join(tmpDir, "relative_dir"), 0755); err != nil {
+					t.Fatalf("Failed to create relative_dir: %v", err)
+				}
+				if err := os.Mkdir(filepath.Join(tmpDir, "test_dir"), 0755); err != nil {
+					t.Fatalf("Failed to create test_dir: %v", err)
+				}
+				if err := os.MkdirAll("/tmp/external_dir", 0755); err != nil {
```
noahdietz: Can this be done without hard coding the `/tmp` path and instead using a different TempDir separate from `tmpDir`? Forcing `/tmp` could reduce portability of these tests I think.

cmd/api-linter/cli_test.go
```diff
@@ -54,3 +57,141 @@ func TestNewCli(t *testing.T) {
...
+			// Dynamically set protoImportPaths for this specific test case
+			var actualProtoImportPaths []string
+			if test.name == "CurrentDirAsAbsolutePath" {
+				actualProtoImportPaths = []string{currentTestDir}
+			} else {
+				actualProtoImportPaths = test.protoImportPaths
+			}
+
+			if err := os.Chdir(currentTestDir); err != nil {
+
```
noahdietz, line 188: Nit: remove extraneous new line
```suggestion
```

cmd/api-linter/cli_test.go
```diff
@@ -54,3 +57,141 @@ func TestNewCli(t *testing.T) {
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// Set up a temporary directory for the test if a setup function is provided.
+			// Change to this directory.
+			var currentTestDir string
+			if test.setupTempDir != nil {
+				currentTestDir = test.setupTempDir(t)
+			} else {
+				currentTestDir = t.TempDir()
+			}
```
noahdietz: Let's simplify this. Define a func variable in this test that just executes `t.TempDir`, set it on all test cases manually in the above slice of tests, and consolidate this into one line `ctd := test.setupTempDir(t)`

`defaultTempDir = func(t *testing.T)string { return t.TempDir() }`

then all of the tests that don't need a special dir set `setupTempDir: defaultTempDir,`

There is already a few examples of this in your test cases, so let's just make it consistent

cmd/api-linter/cli_test.go
```diff
@@ -54,3 +57,141 @@ func TestNewCli(t *testing.T) {
...
+
+			// Dynamically set protoImportPaths for this specific test case
+			var actualProtoImportPaths []string
+			if test.name == "CurrentDirAsAbsolutePath" {
+				actualProtoImportPaths = []string{currentTestDir}
+			} else {
+				actualProtoImportPaths = test.protoImportPaths
+			}
```
noahdietz: This has a bit of a "smell" to me :) ideally we are able to look at the test cases and understand what the input will be without having to look at the test code. This is "logic in the test" which is a "no no" https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

Can we think of another mechanism for doing this OR split this into a separate test?

cmd/api-linter/cli.go
quirogas: I replicated the bug by using a debugger to interpret how Vim was resolving the file paths. Although I couldn't fully replicate the bug within Vim because it was breaking my configuration, I was able to investigate the potential cause. You can reference the configuration [[here](https://github.com/fredrikaverpil/dotfiles/blob/92d9a1190b0cf2a7041b2724062e818da3400acb/nvim-fredrik/lua/fredrik/plugins/lang/protobuf.lua#L139)](https://github.com/fredrikaverpil/dotfiles/blob/92d9a1190b0cf2a7041b2724062e818da3400acb/nvim-fredrik/lua/fredrik/plugins/lang/protobuf.lua#L139).

In an iteration of the code there was a chance for the imports to be empty, regardless of the `--descriptor-set-in` flag being present. This was equivalent to passing an empty slice for the import paths:

```go
protofiles, err = protoparse.ResolveFilenames([]string{}, protoFiles...) // imports are empty
```

This, in turn, executes the following code:

```go
func ResolveFilenames(importPaths []string, fileNames ...string) ([]string, error) {
    if len(importPaths) == 0 { // importPaths is empty
        if containsAbsFilePath(fileNames) { // Vim was passing absolute paths, so this would be true.
            // We have to do this as otherwise parseProtoFiles can result in duplicate symbols.
            // For example, assume we import "foo/bar/bar.proto" in a file "/home/alice/dev/foo/bar/baz.proto"
            // as we call ParseFiles("/home/alice/dev/foo/bar/bar.proto","/home/alice/dev/foo/bar/baz.proto")
            // with "/home/alice/dev" as our current directory. Due to the recursive nature of parseProtoFiles,
            // it will discover the import "foo/bar/bar.proto" in the input file, and call parse on this,
            // adding "foo/bar/bar.proto" to the parsed results, as well as "/home/alice/dev/foo/bar/bar.proto"
            // from the input file list. This will result in a
            // 'duplicate symbol SYMBOL: already defined as field in "/home/alice/dev/foo/bar/bar.proto'
            // error being returned from ParseFiles.
            return nil, errNoImportPathsForAbsoluteFilePath
        }
        return fileNames, nil
    }
    // ... more logic
}
```

which only runs into errors if the files passed where passed by absolute path, so adding `-I "."` would've resolved the issue with the file paths but exposed the caller to potential duplicate symbol errors. While this didn't occur in their specific case, it could've, which reintroduces the original issue.

There could be many reasons why the linter was receiving an absolute path, all related to Vim's configuration, which explains why the error was isolated to that environment. [[Here](https://github.com/fredrikaverpil/dotfiles/blob/92d9a1190b0cf2a7041b2724062e818da3400acb/nvim-fredrik/lua/fredrik/plugins/lang/protobuf.lua#L13)](https://github.com/fredrikaverpil/dotfiles/blob/92d9a1190b0cf2a7041b2724062e818da3400acb/nvim-fredrik/lua/fredrik/plugins/lang/protobuf.lua#L13) is an example that causes a file to be passed as an absolute path.

noahdietz: I see, well done debugging that and putting the pieces together! I think you've covered this case in the resolveImports already then right?

cmd/api-linter/cli_test.go
quirogas: Good catch. I'll re-write this to make sure is portable.

quirogas: Looking more into  `TempDir()`, it will always return a new unique directory, so the solution is to call it twice.

cmd/api-linter/cli.go
```diff
@@ -305,3 +295,73 @@ func getOutputFormatFunc(formatType string) formatFunc {
...
+	if len(protoImportPaths) == 0 {
+		return []string{"."}
+	}
+
+	// Get the absolute path of the current working directory.
+	cwd, err := os.Getwd()
+	if err != nil {
+		// Fallback: If we can't get CWD, return only the provided paths and "."
+		seen := make(map[string]bool)
+		seen["."] = true
```
noahdietz: Use composite literal when initializing a variable with a  non-default value https://google.github.io/styleguide/go/best-practices.html#composite-literals
```suggestion
		seen := map[string]bool{
		  ".": true,
		}
```

cmd/api-linter/cli.go
```diff
@@ -305,3 +295,73 @@ func getOutputFormatFunc(formatType string) formatFunc {
...
+	evaluatedCwd, err := filepath.EvalSymlinks(cwd)
+	if err != nil {
+		// Fallback to Clean if EvalSymlinks fails (e.g., path does not exist)
+		evaluatedCwd = filepath.Clean(cwd)
+	}
+
+	// Initialize resolvedImports with "." and track its canonical absolute path.
+	resolvedImports := []string{"."}
+	seenAbsolutePaths := make(map[string]bool)
+	seenAbsolutePaths[evaluatedCwd] = true // Mark canonical CWD as seen
```
noahdietz:
```suggestion
	seenAbsolutePaths := map[string]bool{
	  evaluatedCwd: true, // Mark canonical CWD as seen
	}
```

cmd/api-linter/cli.go
```diff
@@ -305,3 +295,73 @@ func getOutputFormatFunc(formatType string) formatFunc {
 	}
 	return yaml.Marshal
 }
+
+func resolveImports(protoImportPaths []string) []string {
```
noahdietz: Nit: maybe simplify the name `protoImportPaths` to `imports`  https://go.dev/wiki/CodeReviewComments#variable-names

cmd/api-linter/cli_test.go
noahdietz: Nice!

cmd/api-linter/cli_test.go
noahdietz: Looks much cleaner, thanks

cmd/api-linter/cli_test.go
noahdietz: Good change, thanks!

PR #1498, opened 2025-07-14
fix(AIP-158): clarify pluralized response field finding

quirogas:
## Description

The error message for the `response-plural-first-field` rule (AIP-158) was generic and did not provide sufficient context when a resource defines a `(google.api.resource).plural` annotation. This could be confusing for developers when dealing with irregular, non-standard pluralizations, or simply mismatched pluralizations.

## Fix

The rule's logic has been updated to check for the presence of the `(google.api.resource).plural` annotation. If the annotation exists, the error message now explicitly states that the suggested plural form is based on that annotation. This provides clearer guidance to the developer. If no annotation is found, the rule falls back to the original generic message.

Internal bug http://b/430460300

### DIFF
--- a/rules/aip0136/response_message_name.go
+++ b/rules/aip0136/response_message_name.go
@@ -53,8 +53,17 @@ var responseMessageName = &lint.MethodRule{
 			// just exit.
 			return nil
 		}
+		res := utils.GetResource(response)
+		responseResourceType := res.GetType()
 		requestResourceType := utils.GetResourceReference(m.GetInputType().FindFieldByName("name")).GetType()
-		responseResourceType := utils.GetResource(response).GetType()
+
+		// Check to see if the custom method uses the resource type name as the target
+		// field name and use that instead if `name` is not present as well.
+		// AIP-144 methods recommend this naming style.
+		resourceFieldType := utils.GetResourceReference(m.GetInputType().FindFieldByName(utils.GetResourceSingular(res))).GetType()
+		if requestResourceType == "" && resourceFieldType != "" {
+			requestResourceType = resourceFieldType
+		}
 
 		// Short-circuit: Output type is the resource being operated on
 		if utils.IsResource(response) && responseResourceType == requestResourceType {
--- a/rules/aip0136/response_message_name_test.go
+++ b/rules/aip0136/response_message_name_test.go
@@ -102,13 +102,15 @@ func TestResponseMessageName(t *testing.T) {
 			testName        string
 			MethodName      string
 			RespMessageName string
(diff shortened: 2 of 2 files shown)


rules/aip0158/response_plural_first_field.go
```diff
@@ -37,13 +41,15 @@ var responsePluralFirstField = &lint.MessageRule{
 		want := utils.GetResourcePlural(utils.GetResource(firstField.GetMessageType()))
 		if want != "" {
 			want = strcase.SnakeCase(want)
+			message = fmt.Sprintf("The name of the first field of a paginated response should be the snake_case plural of the resource, which is %q based on the `(google.api.resource).plural` annotation.", want)
```
noahdietz:
```suggestion
			message = "Paginated resource response field name should be the snake_case form of the resource type plural defined in the `(google.api.resource)` annotation.",
```

Let's let the `Suggestion` carry the `want` rather than the message itself, since we have one that we can actually "Apply fix" for. Then, just tighten up the language a bit to keep it more concise. WDYT?

This eliminates the `fmt` import to btw.

rules/aip0158/response_plural_first_field_test.go
```diff
@@ -103,3 +103,43 @@ func TestResponsePluralFirstField(t *testing.T) {
 		}
 	})
 }
+
+func TestResponsePluralFirstFieldWithFieldBehavior(t *testing.T) {
```
noahdietz: We already have a test for this, no? The above LibraryBooks test.

quirogas: Yes, I cloned the test for ease of debugging.  It's now removed.

rules/aip0158/response_plural_first_field_test.go
```diff
@@ -62,6 +62,8 @@ func TestResponsePluralFirstField(t *testing.T) {
 	}{
 		{"ValidResourcePlural", "library_books", testutils.Problems{}},
 		{"InvalidResourcePlural", "books", testutils.Problems{{Suggestion: "library_books"}}},
+		{"ValidResourcePluralAnnotation", "library_books", testutils.Problems{}},
+		{"InvalidResourcePluralAnnotation", "book", testutils.Problems{{Suggestion: "library_books"}}},
```
quirogas: @noahdietz let me know if this aligns with the behaviour we want.

noahdietz: You know what I think we have these two test cases already, on the lines immediately preceding the lines you've added. :)

PR #1501, opened 2025-07-15
feat: add relative time segments rule

quirogas:
###  Description

AIP-142 recommends that fields representing a duration or a time offset have clarifying comments. Specifically, for
google.protobuf.Duration fields ending in _offset, it's crucial to document what the offset is relative to, in order to avoid
ambiguity for API consumers.

This change introduces a new linter rule to enforce this guideline.

internal bug http://b/118149385


rules/aip0142/duration_offset_comment.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+var durationOffsetComment = &lint.FieldRule{
+	Name: lint.NewRuleName(142, "duration-offset-comment"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		return utils.GetTypeName(f) == "google.protobuf.Duration" && strings.HasSuffix(f.GetName(), "_offset")
+	},
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		comment := f.GetSourceInfo().GetLeadingComments()
+		if comment == "" {
```
noahdietz: We already have https://linter.aip.dev/192/has-comments so I'm not sure if this is the most useful heuristic. If there is a comment `// Start offset` this rule does not help us with the guidance it is intended to enforce.

Is there another heuristic that we can use to evaluate a comment for its compliance? Do you think `strings.Contains(comment, "relative")` would be useful/accurate enough? Perhaps take a look at existing fields in googleapis.

quirogas: Yes, I think that is a good point. I think relative is a good candidate.

rules/aip0142/duration_offset_comment.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+var durationOffsetComment = &lint.FieldRule{
+	Name: lint.NewRuleName(142, "duration-offset-comment"),
+	OnlyIf: func(f *desc.FieldDescriptor) bool {
+		return utils.GetTypeName(f) == "google.protobuf.Duration" && strings.HasSuffix(f.GetName(), "_offset")
+	},
+	LintField: func(f *desc.FieldDescriptor) []lint.Problem {
+		comment := f.GetSourceInfo().GetLeadingComments()
+		if comment == "" || (!strings.Contains(strings.ToLower(comment), "relative") && !strings.Contains(strings.ToLower(comment), "in respect") && !strings.Contains(strings.ToLower(comment), "of the")) {
```
noahdietz: Is "of the" really a good signal?  I'm a little skeptical 😅 My gut says that this could easily appear in a sentence not related to the offset relation point.

noahdietz: Move the `strings.ToLower` call to the `comment` initialization line
```suggestion
		comment := strings.ToLower(f.GetSourceInfo().GetLeadingComments())
		if comment == "" || (!strings.Contains(comment, "relative") && !strings.Contains(comment, "in respect") && !strings.Contains(comment, "of the")) {
```

rules/aip0142/duration_offset_comment_test.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+		{
+			name:      "IrrelevantNoOffsetSuffix",
+			Comment:   "",
+			FieldName: "start",
+			FieldType: "google.protobuf.Duration",
+			problems:  nil,
+		},
+		{
+			name:      "IrrelevantNotDuration",
```
noahdietz: Nit: I like to name these `Skip` test cases
```suggestion
			name:      "SkipNoOffsetSuffix",
			Comment:   "",
			FieldName: "start",
			FieldType: "google.protobuf.Duration",
			problems:  nil,
		},
		{
			name:      "SkipNotDuration",
```

rules/aip0142/duration_offset_comment.go
quirogas: When looking at current examples, I noticed "of the" appeared in a few descriptions, so I thought including it would prevent the linting rule from triggering in those situations. I'm not too in love with that idea, though. Since I think we'll end up with some false positives either way, perhaps we can remove it and see how it does? Let me know what you think.

noahdietz: Agreed, I think that is the right choice (to remove and see how it does). Let's do that.

noahdietz (approved):

PR #1513, opened 2025-07-31
feat!: Migrate from jhump/protoreflect to google.golang.org/protobuf

quirogas:
# Migrate from jhump/protoreflect to google.golang.org/protobuf

This pull request executes a large-scale migration of the entire codebase from the third-party `jhump/protoreflect` library to the official `google.golang.org/protobuf` library and its associated APIs (`protoreflect`, `protocompile*`, etc.).

The primary goal of this change is **100% functional equivalence**. This PR is intended as a pure dependency swap with no intentional behavioral changes.

> Note: in the case of `protocompile` the migration was to it successor `bufbuild/protocompile`

### Key Changes:

*   **Dependency Swap:** All imports and usages of `github.com/jhump/protoreflect` have been removed and replaced with `google.golang.org/protobuf`. The `go.mod` and `go.sum` files have been updated accordingly.
*   **API Usage Update:** The codebase has been updated to use the modern `protoreflect` interfaces for descriptor traversal, option access, and type handling.
*   **Test Infrastructure:** The test suite's proto parsing logic has been migrated from `jhump/protoparse` to `bufbuild/protocompile`, which is the recommended tool for working with the new `protoreflect` library. This is consolidated in the `rules/internal/testutils/compile.go` file (this is to keep the test logic relatively similar).

### BREAKING CHANGE:

The public API of the linter rules has changed. Any functions or rule definitions that previously exposed types from the `jhump/protoreflect` library (e.g., `desc.MessageDescriptor`, `desc.FieldDescriptor`) now expose the official `protoreflect.MessageDescriptor`, `protoreflect.FieldDescriptor`, etc. types from `google.golang.org/protobuf`.

Consumers who use the `api-linter` as a library and have written custom rules will need to update their code to use the new `protoreflect` types.

### For Reviewers:

This is a large but highly mechanical change. The vast majority of diffs are simple type and method name replacements.

When reviewing, please focus on:

1.  **Functional Equivalence:** The most critical task is to spot any potential logic drift. The changes should be a 1:1 mapping of the old API calls to the new ones.
2.  **Test Scrutiny:** The tests are our primary defense against regressions. While CI is passing, it's important to ensure that tests were not inadvertently weakened during the migration. 
3.  **Consistency:** Ensure the migration has been applied consistently across the entire codebase.

Given the size of the PR, it may be helpful to review  directory-by-directory, starting with `lint/`, then `locations/`, then `rules/internal`, then `rules/`, then `cmd/` (helps put everything together as changes the previous are dependencies to the other).

### DIFF
--- a/internal/version.go
+++ b/internal/version.go
@@ -14,4 +14,4 @@
 package internal
 
 // Version is the current tagged release of the library.
-const Version = "2.0.0-rc.2"
+const Version = "2.0.0-beta.1"


lint/testutils_test.go
noahdietz, line 1: This file should just be reverted, the test code should just use builders. See other comment in this directory for an example.

lint/lint_test.go
```diff
@@ -103,11 +99,7 @@ func TestLinter_run(t *testing.T) {
 }
 
 func TestLinter_LintProtos_RulePanics(t *testing.T) {
-	fd, err := builder.NewFile("test.proto").Build()
-	if err != nil {
-		t.Fatalf("Failed to build the file descriptor.")
-	}
-
```
noahdietz: @quirogas  This should actually be using the descriptor builders not this template-based `buildFile` function.

https://go.dev/play/p/GFheUaTmFuE

locations/locations.go
```diff
@@ -25,18 +25,78 @@ package locations
...
+// This function is necessary because the modern `protoreflect` API does not
+// provide a way to get this path directly from a descriptor, unlike the older
+// `jhump/protoreflect` library. While `SourceLocations.ByDescriptor()` can find
+// the location for an entire descriptor, it does not provide the `[]int32` path
+// itself. This path is the essential "key" needed to look up the locations of
+// a descriptor's sub-elements (e.g., a field's name vs. its type).
+//
+// This function reconstructs the path by traversing upwards from the descriptor
+// to the file root, collecting the field numbers and indexes at each step.
+func getDescriptorPath(d protoreflect.Descriptor) []int32 {
```
quirogas: @noahdietz, I'd appreciate it if you could take a closer look at this function—not so much the code itself, but rather a better alternative. As mentioned in the comments, this function emulates some lost functionality. I attempted a different approach, but it resulted in a larger set of codebase changes.

noahdietz: In looking at the `protoreflect` docs, `SourceLocation` has a `Path SourcePath` property which is a type-aliased `[]int32`. Is there and issue using this? 

image
image

cmd/api-linter/cli.go
```diff
@@ -138,60 +141,77 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
...
+	// warning handler to maintain the same behavior of ignoring them.
+	var collectedErrors []error
+	rep := reporter.NewReporter(func(err reporter.ErrorWithPos) error {
+		collectedErrors = append(collectedErrors, err)
+		return nil // Returning nil signals the compiler to continue.
+	}, nil)
+
+	compiler := protocompile.Compiler{
+		Resolver:       protocompile.WithStandardImports(protocompile.CompositeResolver(resolvers)),
+		SourceInfoMode: protocompile.SourceInfoStandard,
```
noahdietz: We might want to consider using the "extra" modes here
image

I'm not sure if the extra comments would be useful, but the options locations I think would be. A little surprised we haven't needed it though I suppose.

lint/problem.go
```diff
@@ -76,11 +76,28 @@ func (p Problem) MarshalYAML() (interface{}, error) {
...
-	loc := p.Location
-	if loc == nil && p.Descriptor != nil {
-		loc = p.Descriptor.GetSourceInfo()
+	var fl fileLocation
+	if p.Location != nil {
+		// If Location is set, use it.
+		fl = fileLocationFromPBLocation(p.Location, p.Descriptor)
+	} else if p.Descriptor != nil {
+		// Otherwise, use the descriptor's location.
+		// This is the protobuf-go idiomatic way to get the source location.
```
noahdietz, line 85:
```suggestion
		// This is the protobuf-go idiomatic way to get the source location.
		// Note: ParentFile() called on a FileDescriptor returns itself.
```

This was a question I had, and is apparent in the documentation for this API, but had me worried at first, so thought it might be useful to include to help make this all more readable.

lint/rule_enabled_test.go
```diff
@@ -47,7 +48,7 @@ func TestRuleIsEnabled(t *testing.T) {
 		{"MessageDisabled", "", "api-linter: a::b::c=disabled", false},
 		{"NameNotMatch", "", "api-linter: other=disabled", true},
 		{"RegexpNotMatch", "", "api-lint: a::b::c=disabled", true},
-		{"AliasDisabled", "", "api-linter: d::e::f=disabled", false},
+		{"AliasDisabled", "", " api-linter: d::e::f=disabled", false},
```
noahdietz: Nit: Extraneous white space added here?

lint/rule_enabled_test.go
```diff
@@ -87,26 +104,39 @@ func TestRuleIsEnabledFirstMessage(t *testing.T) {
...
+			{
+				Name: proto.String("SecondMessage"),
+			},
+		},
+		SourceCodeInfo: &descriptorpb.SourceCodeInfo{
+			Location: []*descriptorpb.SourceCodeInfo_Location{
+				{
+					Path:            []int32{4, 0}, // message_type 0
+					Span:            []int32{1, 1, 1, 1},
+					LeadingComments: proto.String(" api-linter: test=disabled"),
```
noahdietz: Is this leading white space necessary?
```suggestion
					LeadingComments: proto.String("api-linter: test=disabled"),
```

lint/rule_enabled_test.go
```diff
@@ -115,35 +145,62 @@ func TestRuleIsEnabledParent(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		SourceCodeInfo: &descriptorpb.SourceCodeInfo{
+			Location: []*descriptorpb.SourceCodeInfo_Location{
+				{
+					Path:            []int32{4, 0}, // message_type 0
+					Span:            []int32{1, 1, 1, 1},
+					LeadingComments: proto.String(" api-linter: test=disabled"),
```
noahdietz: Is this leading white space necessary?

```suggestion
					LeadingComments: proto.String("api-linter: test=disabled"),
```

locations/descriptor_locations.go
```diff
@@ -15,14 +15,14 @@
...
 
 // DescriptorName returns the precise location for a descriptor's name.
 //
 // This works for any descriptor, regardless of type (message, field, etc.).
-func DescriptorName(d desc.Descriptor) *dpb.SourceCodeInfo_Location {
+func DescriptorName(d protoreflect.Descriptor) *dpb.SourceCodeInfo_Location {
 	// All descriptors seem to have `string name = 1`, so this conveniently works.
 	return pathLocation(d, 1)
-}
+}
```
noahdietz: Missing a new line here?

locations/descriptor_locations_test.go
```diff
@@ -54,4 +54,4 @@ func TestFileDescriptorName(t *testing.T) {
 	if got := DescriptorName(f); got != nil {
 		t.Errorf("%v", got)
 	}
-}
+}
```
noahdietz: Missing a new line here?

locations/field_locations_test.go
```diff
@@ -97,8 +97,8 @@ func TestFieldOption(t *testing.T) {
 		  }];
 		}
 	`)
-	loc := FieldOption(f.GetMessageTypes()[0].GetFields()[0], apb.E_ResourceReference)
+	loc := FieldOption(f.Messages().Get(0).Fields().Get(0), apb.E_ResourceReference)
 	if diff := cmp.Diff(loc.GetSpan(), []int32{4, 19, 6, 3}); diff != "" {
 		t.Error(diff)
 	}
-}
+}
```
noahdietz: Missing new line here?

locations/file_locations.go
```diff
@@ -68,24 +68,24 @@ func FilePhpNamespace(f *desc.FileDescriptor) *dpb.SourceCodeInfo_Location {
...
+func FileImport(f protoreflect.FileDescriptor, index int) *dpb.SourceCodeInfo_Location {
 	return pathLocation(f, 3, index) // 3 == dependency
 }
 
 // FileCCEnableArenas returns the location of the `cc_enable_arenas` option.
-func FileCCEnableArenas(f *desc.FileDescriptor) *dpb.SourceCodeInfo_Location {
+func FileCCEnableArenas(f protoreflect.FileDescriptor) *dpb.SourceCodeInfo_Location {
 	return pathLocation(f, 8, 31) // 8 == (file) options, 31 == cc_enable_arenas
-}
+}
```
noahdietz: Missing new line here?

locations/message_locations.go
```diff
@@ -15,13 +15,13 @@
...
 	dpb "google.golang.org/protobuf/types/descriptorpb"
 )
 
 // MessageResource returns the precise location of the `google.api.resource`
 // annotation.
-func MessageResource(m *desc.MessageDescriptor) *dpb.SourceCodeInfo_Location {
+func MessageResource(m protoreflect.MessageDescriptor) *dpb.SourceCodeInfo_Location {
 	return pathLocation(m, 7, int(apb.E_Resource.TypeDescriptor().Number())) // MessageDescriptor.options == 7
-}
+}
```
noahdietz: Missing new line here?

locations/message_locations_test.go
```diff
@@ -30,8 +30,8 @@ func TestMessageResource(t *testing.T) {
 		  };
 		}
 	`)
-	loc := MessageResource(f.GetMessageTypes()[0])
+	loc := MessageResource(f.Messages().Get(0))
 	if diff := cmp.Diff(loc.GetSpan(), []int32{4, 2, 7, 4}); diff != "" {
 		t.Error(diff)
 	}
-}
+}
```
noahdietz: Missing new line here?

locations/method_locations.go
```diff
@@ -16,40 +16,40 @@ package locations
...
+func MethodSignature(m protoreflect.MethodDescriptor, index int) *dpb.SourceCodeInfo_Location {
 	return pathLocation(m, 4, int(apb.E_MethodSignature.TypeDescriptor().Number()), index) // MethodDescriptor.options == 4
 }
 
 // MethodOption returns the precise location of the method's option with the given field number, if any.
-func MethodOption(m *desc.MethodDescriptor, fieldNumber int) *dpb.SourceCodeInfo_Location {
+func MethodOption(m protoreflect.MethodDescriptor, fieldNumber int) *dpb.SourceCodeInfo_Location {
 	return pathLocation(m, 4, fieldNumber) // MethodDescriptor.options == 4
-}
+}
```
noahdietz: Missing new line here?

locations/method_locations_test.go
```diff
@@ -141,10 +141,10 @@ func TestMethodOption(t *testing.T) {
...
 			// field number of the deprecated option == 33
-			loc := MethodOption(f.GetServices()[0].GetMethods()[test.methodIdx], 33)
+			loc := MethodOption(f.Services().Get(0).Methods().Get(test.methodIdx), 33)
 			if diff := cmp.Diff(loc.GetSpan(), test.want); diff != "" {
 				t.Errorf("Diff: %s", diff)
 			}
 		})
 	}
-}
+}
```
noahdietz: Missing new line here?

noahdietz (commented):
I've reviewed `cmd` `lint` and `locations` so far

cmd/api-linter/cli.go
noahdietz: https://pkg.go.dev/github.com/bufbuild/protocompile#pkg-constants

rules/internal/utils/http.go
```diff
@@ -35,15 +34,29 @@ func HasHTTPRules(m *desc.MethodDescriptor) bool {
...
 	rules := []*HTTPRule{}
 
 	// Get the method options.
-	opts := m.GetMethodOptions()
+	opts := m.Options()
 
 	// Get the "primary" rule (the direct google.api.http annotation).
-	if x := proto.GetExtension(opts, apb.E_Http); x != nil {
-		httpRule := x.(*apb.HttpRule)
+	if opts.ProtoReflect().Has(apb.E_Http.TypeDescriptor()) {
```
noahdietz: We should see if we can simply migrate to the equivalent `proto.GetExtension` ([docs](https://pkg.go.dev/github.com/bufbuild/protocompile#pkg-constants)). We spoke offline about this

rules/aip0123/resource_name_field_test.go
```diff
@@ -30,7 +30,7 @@ func TestResourceNameField(t *testing.T) {
 		problems testutils.Problems
 	}{
 		{"ValidBothPresent", `option (google.api.resource) = { type: "foo" };`, `string name = 1;`, nil},
-		{"InvalidNoField", `option (google.api.resource) = { type: "foo" };`, ``, testutils.Problems{{Message: "`name`"}}},
+		{"InvalidNoField", `option (google.api.resource) = { type: "foo" };`, ``, testutils.Problems{{Message: "Message `Book` has no `name` field."}}},
```
noahdietz: Nit: This shouldn't need to be fully spelled out. Not urgent to fix though

rules/aip0123/name_never_optional.go
```diff
@@ -15,29 +15,29 @@
...
+	LintMessage: func(m protoreflect.MessageDescriptor) []lint.Problem {
 		f := "name"
 		if nf := utils.GetResource(m).GetNameField(); nf != "" {
 			f = nf
 		}
-		field := m.FindFieldByName(f)
+		field := m.Fields().ByName(protoreflect.Name(f))
 
-		if field.IsProto3Optional() {
+		if field.HasOptionalKeyword() {
```
noahdietz, line 40: Nit: This is more of a note to self - `protoreflect` notably differentiates between `HasOptionalKeyWord` and `HasPresence` which in `proto3` is enabled by using the `optional` keyword, but in Editions, is not (there is a field-level feature option for field_presence). To support Editions, we may need to adjust this logic, but that is already tracked in our internal bug list for said work.

rules/aip0126/upper_snake_values_test.go
```diff
@@ -34,24 +34,25 @@ func TestUpperSnake(t *testing.T) {
...
-				t.Fatalf("Could not build an enum with values %v", test.enumValues)
+			var enumValues string
+			for i, v := range test.enumValues {
+				enumValues += fmt.Sprintf("%s = %d;\n", v, i)
 			}
+			f := testutils.ParseProto3Tmpl(t, `
+				enum Number {
+					{{.EnumValues}}
+				}
+			`, struct{ EnumValues string }{enumValues})
```
noahdietz, line 45: This is a bit roundabout, and in general we've built descriptors from literals, but I suppose it can stay as it is relatively readable.

rules/aip0128/resource_reconciling_field.go
```diff
@@ -46,4 +45,4 @@ var resourceReconcilingField = &lint.MessageRule{
 
 		return nil
 	},
-}
+}
```
noahdietz: Nit: missing new line at EoF
```suggestion
}

```

rules/aip0131/http_uri_name_test.go
```diff
@@ -28,8 +28,8 @@ func TestHttpNameField(t *testing.T) {
 		problems   testutils.Problems
 	}{
 		{"Valid", "/v1/{name=publishers/*/books/*}", "GetBook", testutils.Problems{}},
-		{"InvalidVarName", "/v1/{book=publishers/*/books/*}", "GetBook", testutils.Problems{{Message: "`name`"}}},
-		{"NoVarName", "/v1/publishers/*/books/*", "GetBook", testutils.Problems{{Message: "`name`"}}},
+		{"InvalidVarName", "/v1/{book=publishers/*/books/*}", "GetBook", testutils.Problems{{Message: "HTTP URI should include a `name` variable."}}},
```
noahdietz: Nit: Another case where the test want doesn't need to be fully spelled out but can be fixed later.

rules/aip0131/request_required_fields_test.go
```diff
@@ -92,13 +92,13 @@ func TestRequiredFieldTests(t *testing.T) {
...
-				dbr = f.FindMessage("GetBookRequest").FindFieldByName(test.problematicFieldName)
+				dbr = f.Messages().Get(0).Fields().ByName(protoreflect.Name(test.problematicFieldName))
 			}
 			if diff := test.problems.SetDescriptor(dbr).Diff(requestRequiredFields.Lint(f)); diff != "" {
 				t.Error(diff)
 			}
 		})
 	}
-}
+}
```
noahdietz: Nit: Missing new line at EoF
```suggestion
}

```

I won't comment on these anymore assuming all examples will be taken care of wholesale

rules/aip0132/http_uri_parent_test.go
```diff
@@ -29,8 +29,8 @@ func TestHTTPURIParent(t *testing.T) {
 		problems     testutils.Problems
 	}{
 		{"Valid", "/v1/{parent=publishers/*/books/*}", "ListBooks", "string parent = 1;", nil},
-		{"InvalidVarParent", "/v1/{book=publishers/*/books/*}", "ListBooks", "string parent = 1;", testutils.Problems{{Message: "`parent` variable"}}},
-		{"InvalidNoVarParent", "/v1/publishers/*/books/*", "ListBooks", "string parent = 1;", testutils.Problems{{Message: "`parent` variable"}}},
+		{"InvalidVarParent", "/v1/{book=publishers/*/books/*}", "ListBooks", "string parent = 1;", testutils.Problems{{Message: "HTTP URI should include a `parent` variable."}}},
+		{"InvalidNoVarParent", "/v1/publishers/*/books/*", "ListBooks", "string parent = 1;", testutils.Problems{{Message: "HTTP URI should include a `parent` variable."}}},
```
noahdietz: Another place where the test.want was unnecessarily expanded.

I'll stop commenting on this for now. It's not a huge deal.

rules/aip0132/request_show_deleted_required_test.go
```diff
@@ -36,14 +36,14 @@ func TestRequestShowDeletedRequired(t *testing.T) {
...
-			if diff := test.problems.SetDescriptor(f.GetMessageTypes()[1]).Diff(problems); diff != "" {
+			if diff := test.problems.SetDescriptor(f.Messages().Get(1)).Diff(problems); diff != "" {
 				t.Error(diff)
 			}
 		})
 	}
 }
 
-// Regression test for https://github.com/googleapis/api-linter/issues/854.
+// Regression test for https://github.com/googleapis/api-linter/v2/issues/854.
```
noahdietz:
```suggestion
// Regression test for https://github.com/googleapis/api-linter/issues/854.
```
Valid Issue link caught in refactoring :)

noahdietz (commented):
I've now reviewed through `rules/aip0132`

Here are some high level comments that don't need to be integrated in this PR/review:
1. It would be great if we consistently used `string` or `protoreflect.Name` in places so that we don't have type switching in the same file (we could change function signatures, simply declare variables with a type rather than lazy init and switching later, etc.) - this should be done later
2. Meta comment: If only the `protoreflect` "list" types (`Fields`, `FieldDescriptors`, etc.), were iterable via `range` - that seems to be a huge source of diff - I wonder if there is something we can do there in the future to make this code more concise 
image

rules/aip0126/upper_snake_values_test.go
noahdietz: I actually like this a lot in the other places I've seen it happen because it made the test code way more readable in cases where there were a lot of cases to run and a lot of file setup to do. Seeing those diffs actually makes me think this a reasonable thing to do in general. I think that contradicts what I've previously said about retaining similarity in "approach" to test input while refactoring, but at least in the `rules/aip0*` tests, this makes a lot of sense to align on templating.

rules/internal/utils/http.go
quirogas: Alright, after some additional digging, I think the function below is the correct way to get the extension. The issue here is that `Get()` may return a `dynamicpb`, which can't be cast. The workaround is to marshal and then unmarshal the `Message` to avoid complicated traversals through the `Descriptor`. However, since we know what the message looks like, we could modify the `parseRule` function instead and do the traversal there.

noahdietz: Can we not just cast the the `m.Options().(*descriptorpb.MethodOptions)` and use `proto.GetExtension` ?

The docs say `m.Options()` for a `MethodDescriptor` will be `google.protobuf.MethodOptions` or `nil` if the descriptor has no options. We currently aren't handling the `nil` case of `m.Options()` and `if != nil` we should be able to safely cast to `descriptorpb.MethodOptions`.

image

Why is marshaling necessary?

quirogas: The issue isn't with getting the `Options` per se. The problem is that in some cases, the descriptor will be a `descriptorpb` and not a `protobuf`. Consequently, when we get the `Options`, the type may be `*descriptorpb.MethodOptions`, which we would have to cast to `protobuf.MethodOptions` before passing it to `proto.GetExtension`. At that point, we would run into the same issue that the marshalling and unmarshalling is trying to resolve.

noahdietz: You mentioned previously that this seemed to only be an edge case in the test cases and that our integration tests on googleapis were 100% fine with some previous iteration(?) of the code. Is that true?
If so, is there something we are missing in our test setup that is making this an issue? e.g. directly and incorrectly initializing Method Options such that we have this issue?

My gut says that we should not need to be doing `Marshal`/`Unmarshal` for what seems like functionality the APIs seem to support "natively". Sorry for pushing on this so much, I will try to test it myself as well to better understand.

quirogas: I misspoke about the overall repercussions. Initially, I thought this didn't affect the linter when working with `.proto` files, as the issue seemed to stem from dynamic descriptors, which I assumed weren't needed.  Hower during my search I found cases where not marshalling/unmarshalling causes linter panics outside of the tests as well.

noahdietz: Ok. Let's file something to come back to this. We don't need to block on it for this RC and if it works - even as a temporary bandage - we can keep things moving.

locations/locations.go
quirogas: Yes, this is a great solution! Now we can eliminate this function altogether. Thank you!

internal/version.go
```diff
@@ -14,4 +14,4 @@
 package internal
 
 // Version is the current tagged release of the library.
-const Version = "1.70.1"
+const Version = "2.0.0-rc.2"
```
noahdietz: Actually, reading the GO mod docs again, we need this to be `-beta.1`

https://go.dev/doc/modules/version-numbers#pre-release-version
```suggestion
const Version = "2.0.0-beta.1"
```

PR #1523, opened 2025-08-20
feat(AIP-160): Add rule to validate filter field name

shwoodard:
Validates that the field for filtering is called "filter" and not "filters" for List and custom methods' request messages.

### DIFF
--- a/rules/aip0160/filter_field_name.go
+++ b/rules/aip0160/filter_field_name.go
@@ -1,0 +1,50 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aip0160
+
+import (
+	"fmt"
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+const (
+	nameMessageFmt = `Guidance: use the field, "string filter", not "string %s"`
+	nameSuggestion = "filter"
+)
+
+var filterFieldName = &lint.MethodRule{
+	Name: lint.NewRuleName(160, "filter-field-name"),
+	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		return utils.IsListMethod(m) || utils.IsCustomMethod(m)
+	},
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		for _, f := range m.GetInputType().GetFields() {
+			if f.GetName() == "filters" {
+				problems = append(problems, lint.Problem{
+					Message:    fmt.Sprintf(nameMessageFmt, f.GetName()),
+					Descriptor: f,
+					Location:   locations.DescriptorName(f),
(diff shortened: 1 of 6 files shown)


rules/aip0160/filter_field.go
```diff
@@ -0,0 +1,51 @@
...
+var filterField = &lint.MethodRule{
+	Name: lint.NewRuleName(160, "filter-field"),
+	OnlyIf: func(m *desc.MethodDescriptor) bool {
+		return utils.IsListMethod(m) || utils.IsCustomMethod(m)
+	},
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		for _, f := range m.GetInputType().GetFields() {
+			if f.GetName() == "filters" ||
+				(f.GetName() == "filter" && utils.GetTypeName(f) != "string") {
```
noahdietz: Would you be open to separating these into two separate rules? `synonyms` and `filter-field-type`?

We want these to be more granularly configurable, so that disabling one Linter rule does not disable two "checks" simultaneously.

shwoodard: Done, this pr only validates the name and https://github.com/googleapis/api-linter/pull/1524 only validates the type.

rules/aip0160/filter_field_name.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+const (
+	nameMessageFmt = `Guidance: use the field, "string filter", not "string %s"`
+	nameSuggestion = "string filter"
```
noahdietz: If this suggestion is applied to the `location.DescriptorName` it should be simply `filter` - we don't want to include the type in the suggestion that replaces the `FieldDescriptorProto.name`

rules/aip0160/filter_field_name_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/googleapis/api-linter/rules/internal/testutils"
+)
+
+func TestFiltersFieldName(t *testing.T) {
+	tests := []struct {
+		testName   string
+		FieldType  string
+		FieldName  string
+		MethodName string
+		err        bool
```
noahdietz: The typical test pattern is to put `problems testutils.Problems{}` and just specify the desired finding or `nil`. I'm guessing this pattern was used to be more DRY?

I'm OK with keeping this as is, but maybe rename from `err` to `invalid` or something else since `err` is so commonly `error` in Go :)

rules/aip0160/filter_field_name.go
shwoodard: Fixed.

rules/aip0160/filter_field_name_test.go
shwoodard: This was modeled after `TestCount` in `rules/aip0141/count_suffix_test.go`. Nonetheless, good catch! I usually use `wantError`, but I didn't change it after I copied the aforementioned file.

Unless you feel strongly about including the `testutils.Problems{}` in the test cases, yes, I think it is more DRY this way.

Changed to `wantError`.

noahdietz: `wantError` is fine with me.

I _personally_ prefer specifying the expected `testutils.Problems` in each test to avoid putting logic in tests ([public TotT](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html)), but these tests are so simple I'm not too worried about it. Since we already have the pattern, we'd want to refactor and commit to one or the other wholesale, and one more won't hurt that refactor.

shwoodard: @noahdietz ftr, I don't know that the Go authors would agree with the TotT or at least that it applies in this scenario. IMO, keeping the repetition caused by having similar `tesutils.Problems{}` makes the files hard to read either by making the line length long or the table of tests very long. This makes the file harder to navigate. It doesn't seem like logic in the tests, but rather substitution. For these reasons, IMO, at least for table tests, the parameterized solution is preferable.

PR #1524, opened 2025-08-21
feat(AIP-160): Validate filter field type

shwoodard:
Validates that the field for filtering, called "filter," is of type `string`.


shwoodard:
Please merge after https://github.com/googleapis/api-linter/pull/1523, which will make the PR atomic: it will only show changes necessary to add validation for type.

docs/rules/0160/filter-field-type.md
```diff
@@ -0,0 +1,81 @@
...
+## Disabling
+
+If you need to violate this rule, use a leading comment above the method.
+Remember to also include an [aip.dev/not-precedent][] comment explaining why.
+
+```proto
+// (-- api-linter: core::0160::filter-field-type=disabled
+//     aip.dev/not-precedent: We need to do this because reasons. --)
+rpc ListBooks(ListBooksRequest) returns (ListBooksResponse) {
+  string filters = 1;
```
noahdietz:
```suggestion
  bytes filter = 1;
```

copy pasta error: should be an invalid type not field name

rules/aip0160/filter_field_type.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/googleapis/api-linter/lint"
+	"github.com/googleapis/api-linter/locations"
+	"github.com/googleapis/api-linter/rules/internal/utils"
+	"github.com/jhump/protoreflect/desc"
+)
+
+const (
+	typeMessageFmt = `Guidance: use the field, "string filter", not "%s filter"`
+	typeSuggestion = "string filter"
```
noahdietz: This should just be `string` if the suggestion will be on the `locations.FieldType`

rules/aip0160/filter_field_type.go
```diff
@@ -0,0 +1,50 @@
...
+		return utils.IsListMethod(m) || utils.IsCustomMethod(m)
+	},
+	LintMethod: func(m *desc.MethodDescriptor) []lint.Problem {
+		var problems []lint.Problem
+		for _, f := range m.GetInputType().GetFields() {
+			if f.GetName() == "filter" && utils.GetTypeName(f) != "string" {
+				problems = append(problems, lint.Problem{
+					Message:    fmt.Sprintf(typeMessageFmt, utils.GetTypeName(f)),
+					Descriptor: f,
+					Location:   locations.DescriptorName(f),
```
noahdietz: We can use `locations.FieldType` ([docs](https://pkg.go.dev/github.com/googleapis/api-linter/locations#FieldType))

rules/aip0160/filter_field_type_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestFiltersFieldType(t *testing.T) {
+	tests := []struct {
+		testName   string
+		FieldType  string
+		FieldName  string
+		MethodName string
+		err        bool
+	}{
+		{"Valid", "string", "filter", "ListBooks", false},
+		{"Valid name filters", "string", "filters", "ListBooks", false},
```
noahdietz:
```suggestion
		{"Skip non-filter", "string", "filters", "ListBooks", false},
```

docs/rules/0160/filter-field-type.md
shwoodard: 🙏🏻

rules/aip0160/filter_field_type.go
shwoodard: Done.

rules/aip0160/filter_field_type.go
shwoodard: Good catch! Thanks, done.

rules/aip0160/filter_field_type_test.go
shwoodard: Done.

noahdietz (approved):

PR #1525, opened 2025-08-22
chore(AIP-141) change variable name to wantErr.

shwoodard:
Given that `err` is very commonly assigned as an `error`, it is better to use a different variable name, e.g. `wantErr`.


noahdietz:
> I like the change and the reasoning behind it. tiny comment though, this pattern is already present in most of our tests, so I think it would be better if we adopted it everywhere.

Yeah we were discussing this in https://github.com/googleapis/api-linter/pull/1523#discussion_r2294286706 and we may want to see about refactoring to one pattern or the other. @shwoodard if you want to change all other instances of `test.err` to `test.wantErr` in this PR, that would be great, but also don't feel like you have to, @quirogas and I can do it in a separate change.

That said, @quirogas we may want to reduce noise in `main` branch while you are working on v2 to avoid gnarly merges for you :)

shwoodard:
> I like the change and the reasoning behind it. tiny comment though, this pattern is already present in most of our tests, so I think it would be better if we adopted it everywhere.

I could only find the one instance of this,

```sh
$ find ./rules -name "*_test.go" | xargs grep -E "err[^b]+bool"
./rules/aip0141/count_suffix_test.go:		err       bool
```

PR #1535, opened 2025-09-12
chore(v2): catch up with main

noahdietz:
This PR catches up the `v2` branch with recent changes on `main`. This includes:
* dependency updates
* two new rules in new rule group AIP-160 (migrated from v1 to v2 in this PR, obviously)
* some bug fixes


.release-please-manifest.json
```diff
@@ -1,3 +1,3 @@
 {
-    ".": "1.70.1"
+    ".": "1.71.0"
```
noahdietz, line 2: Note: we will update this by forcing the major version bump once we merge v2 into main and release

PR #1538, opened 2025-09-17
chore(internal): add test util

quirogas:
Add a singular version of  `ParseProtoStrings`  for cases where only one file is needed.

### DIFF
--- a/rules/internal/testutils/parse.go
+++ b/rules/internal/testutils/parse.go
@@ -43,6 +43,7 @@ import (
 //
 // It dedents the string before parsing.
 func ParseProtoStrings(t *testing.T, src map[string]string) map[string]protoreflect.FileDescriptor {
+	t.Helper()
 	filenames := []string{}
 	for k, v := range src {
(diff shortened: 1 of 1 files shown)


rules/internal/testutils/parse.go
```diff
@@ -93,6 +93,14 @@ func ParseProto3String(t *testing.T, src string) protoreflect.FileDescriptor {
 	})["test.proto"]
 }
 
+// ParseProtoString parses a string representing a proto file, and returns
+// a FileDescriptor.
+//
+// It dedents the string before parsing.
+func ParseProtoString(t *testing.T, src string) protoreflect.FileDescriptor {
+	return ParseProtoStrings(t, map[string]string{"test.proto": src})["test.proto"]
```
shwoodard, line 104: Should call `t.Helper()` for this line.

noahdietz: @quirogas we should add this to all of these functions. https://pkg.go.dev/testing#T.Helper

PR #1541, opened 2025-09-19
fix(AIP-140): restrict `uri` naming suggestions to fields with URI in comments

stephanemoore:
AIP-140 was recently updated to clarify that field names may use `url` when the field can only represent a URL. The current URI lint rule for AIP-140 will unconditionally discourage using `url` in field names. This change updates the URI lint rule to be more conservative by only suggesting renaming the field to use `uri` when the field's comments use the term URI.

Related update to AIP-140:
https://github.com/aip-dev/google.aip.dev/pull/1551

### DIFF
--- a/rules/aip0140/uri_test.go
+++ b/rules/aip0140/uri_test.go
@@ -36,6 +36,7 @@ func TestURI(t *testing.T) {
 		{"ValidURLPrefix", "", "url_foo", nil},
 		{"ValidURLSuffix", "", "foo_url", nil},
 		{"ValidURLIntermixed", "", "foo_url_bar", nil},
+		{"ValidIntraWordURIComment", "// A URL to a curio.", "url", nil},
 		{"Invalid", "// A URI.", "url", testutils.Problems{{Message: "uri", Suggestion: "uri"}}},
 		{"InvalidPrefix", "// A uri.", "url_foo", testutils.Problems{{Message: "uri", Suggestion: "uri_foo"}}},
 		{"InvalidSuffix", "// A URI.", "foo_url", testutils.Problems{{Message: "uri", Suggestion: "foo_uri"}}},


rules/aip0140/uri_test.go
```diff
@@ -23,21 +23,27 @@ import (
...
-		{"InvalidSuffix", "foo_url", testutils.Problems{{Message: "uri", Suggestion: "foo_uri"}}},
-		{"InvalidIntermixed", "foo_url_bar", testutils.Problems{{Message: "uri", Suggestion: "foo_uri_bar"}}},
+		{"Valid", "", "uri", nil},
+		{"ValidPrefix", "", "uri_foo", nil},
+		{"ValidSuffix", "", "foo_uri", nil},
+		{"ValidIntermixed", "", "foo_uri_bar", nil},
+		{"ValidURL", "", "url", nil},
+		{"ValidURLPrefix", "", "url_foo", nil},
+		{"ValidURLSuffix", "", "foo_url", nil},
+		{"ValidURLIntermixed", "", "foo_url_bar", nil},
```
noahdietz, line 38: Can you please also add a case for mid-word instances of "uri" - I know the regex should be on word boundary, but let's make sure that is covered

```suggestion
		{"ValidURLIntermixed", "", "foo_url_bar", nil},
	  {"ValidIntraWordURIComment", "// A URL to a curio.", "url", nil},
```

noahdietz, line 38: Can you also please add a confirming case with a comment?
```suggestion
		{"ValidURLIntermixed", "", "foo_url_bar", nil},
		{"ValidWithURIComment", "// A URI.", "uri", nil},
```

stephanemoore: Good idea 👍 I added the suggested case.

stephanemoore: Good idea 👍 I added the suggested case.

PR #1545, opened 2025-09-30
fix(cli):  resolve import path resolution

### LINKED ISSUE #1519: `api-linter` fails to resolve the import path
#### Environment details

  - OS: Windows
  - Package version: `1.70.2`

#### Steps to reproduce

  1. Create a project structured like below

        Image

        Set the `test.proto` text as

        ```protobuf
        syntax = "proto3";
        
        package api.v1;
        
        message Test {
          string name = 1 [(google.api.field_behavior) = IDENTIFIER];
        }
        ```

        Set the `field_behavior.proto` text as https://github.com/googleapis/googleapis/blob/67fe1ece997ef298d8d86b026195ea2fd588c538/google/api/field_behavior.proto

  2. Set the console `cwd` to the project root.
  3. Run the console command

      ```powershell
      api-linter --proto-path .\third_party .\api\v1\test.proto
      ```

      or

      ```powershell
      api-linter --proto-path third_party .\api\v1\test.proto
      ```

This returns an error:

```console
2025/08/07 14:56:11 api/v1/test.proto:5:8: could not resolve path "google/api/field_behavior.proto": open google\api\field_behavior.proto: The system cannot find the path specified.
```

The problem isn't reproduced via v. `1.67.*`

quirogas:
This PR fixes an issue where the linter failed to resolve proto imports correctly, particularly in two key scenarios.

### Fix

The underlying issue was that the compiler attempted to process all proto files at once, which led to errors when dependencies were located in different paths. The fix is to compile each proto file individually, allowing the compiler to correctly resolve the distinct import paths for each file.

Additionally, the `resolveImports` function was updated to correctly handle relative paths, ensuring they are preserved and passed to the compiler.

---

#### 1. Imports from Mixed Roots (Relative and `-I` flag) ( #1519)

The linter can now correctly resolve imports when one dependency is relative to the current working directory and another is located in a path specified by the `-I` flag.

**Testing Structure:**

```console
.
├── api
│   ├── common
│   │   └── common.proto
│   └── v1
│       └── test.proto
└── third_party
    └── google
        └── api
            └── field_behavior.proto
```

**File Contents:**

**`api/common/common.proto`**
```proto
syntax = "proto3";
package api.common;
message Foo {}
```

**`api/v1/test.proto`**
```proto
syntax = "proto3";
package api.v1;
import "api/common/common.proto";
message Test {
  string name = 1 [(google.api.field_behavior) = IDENTIFIER];
  api.common.Foo foo = 2;
}
```

**`third_party/google/api/field_behavior.proto`**
```proto
syntax = "proto3";
package google.api;
// ... content
```

**Command:**

The following command now works as expected, resolving both the `third_party` import and the local `api/common` import:

```sh
api-linter -I third_party api/v1/test.proto
```

---

#### 2. Linting Multiple Files with a Common Root (#1465)

The linter now correctly handles cases where multiple files are passed as arguments and share a common import root that is not part of their package path. This avoids duplicate dependency errors.

**Testing Structure:**

```console
.
└── proto
    ├── google
    │   └── api
    │       ├── annotations.proto
    │       └── http.proto
    └── ...
```

**Command:**

The following command correctly lints both files without issues:

```shell
api-linter -I proto proto/google/api/http.proto proto/google/api/annotations.proto
```

### Important Usage Note

It is important to note that the following command **will not work** for the second scenario:

```shell
# This will fail
api-linter proto/google/api/http.proto proto/google/api/annotations.proto
```

The linter cannot automatically infer that `proto` is the intended import root for the dependencies between `http.proto` and `annotations.proto`. Adding the complexity to infer this is unnecessary.

Instead, users should follow one of two conventions:
1.  Run the `api-linter` from the root of the proto definitions (e.g., `cd proto && api-linter google/api/http.proto ...`).
2.  Use the `-I` flag to explicitly specify the import root, as shown in the working example above.

A new integration test has been added to cover these scenarios and prevent regressions.

--- 

fixes #1519

### DIFF
--- a/cmd/api-linter/cli.go
+++ b/cmd/api-linter/cli.go
@@ -184,9 +184,11 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
 		Reporter:       rep,
 	}
 
+	// Compile each file individually to avoid possible collisions
+	// between a linted file that imports other files that are also being linted.
+	// Otherwise, both the import resolver and the file will be "duplicated".
 	var compiledFiles linker.Files
 	for _, protoFile := range c.ProtoFiles {
-		// Compile each file individually.
 		// The compiler returns a slice of files, even for a single input file.
 		f, err := compiler.Compile(context.Background(), protoFile)
 		// After compilation, check if the handler collected any errors.


cmd/api-linter/cli.go
```diff
@@ -183,25 +184,31 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
...
-	// After compilation, check if the handler collected any errors.
-	// This is the primary source of truth for parse errors when using a
-	// custom reporter that continues on error.
-	if len(collectedErrors) > 0 {
-		errorStrings := make([]string, len(collectedErrors))
-		for i, e := range collectedErrors {
-			errorStrings[i] = e.Error()
+	var compiledFiles linker.Files
+	for _, protoFile := range c.ProtoFiles {
+		// Compile each file individually.
```
noahdietz: Nit: Why? The code is self-explanatory as far as what it is doing, the comment should be providing rationale.

https://testing.googleblog.com/2017/07/code-health-to-comment-or-not-to-comment.html

cmd/api-linter/integration_test.go
```diff
@@ -277,6 +277,131 @@ func TestMultipleFilesFromParentDir(t *testing.T) {
...
+		syntax = "proto3";
+
+		package api.common;
+
+		message Foo {}
+	`); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := writeFile(filepath.Join(thirdPartyDir, "field_behavior.proto"), `
```
noahdietz: Can we make the test simpler (just import a type, not use an option) and not use a real proto (avoid using a copy of the real `google/api/field_behavior.proto`) in the test? It is unlikely, but global proto registry collisions could happen in this test given that we force register the common protos.

noahdietz:
Hey @CuriousStork would you be interested in testing this out ? We think it fixes the issue you reported and the one we were trying to fix when we broke this case - your repro was really helpful.

CuriousStork:
@noahdietz Hello, I'd like to test it. However, I guess I need to compile a binary from the source code? I'm not familiar with Go, so this may take some time.

noahdietz:
> @noahdietz Hello, I'd like to test it. However, I guess I need to compile a binary from the source code? I'm not familiar with Go, so this may take some time.

Ah apologies! If you have Go installed on your machine, you can compile/install in path by do the following `go install github.com/googleapis/api-linter/v2/cmd/api-linter@fix-resolve-import-path`. If your [GOPATH](https://go.dev/wiki/GOPATH) is part of your PATH, it will be accessible from there. 

If you don't mind trying, it would be great, but I understand not having the time to and waiting for us to push a pre-release.

CuriousStork:
@noahdietz Thank you and @quirogas for fixing the issue.

In short, it seems to be fixed.

I've made a very quick smoke test on Ubuntu with my real project files.

`api-linter --version` output for the compiled binary:

```console
api-linter 2.0.0-beta.4
```

Running with `--proto-path=third_party` was successful. The project root directory import was correctly inferred w/o explicitly adding the root to `--proto-path`. I've tested a couple of files: one from `third_party` and another from the main project API section. The latter imports both from `third_party` and another directory.

PS: I guess testing on Windows isn't necessary.

Issue #1547, opened 2025-10-07
Allow view field in AIP-133 AIP-134 and AIP-135 Soft Delete request messages


noahdietz:
As per https://github.com/aip-dev/google.aip.dev/pull/1562, we will allow the AIP-157 `View` pattern to be used with the other standard methods, specifically AIP-133 Create, AIP-134 Update, and AIP-135 Soft Delete (AIP-164). This will allow services to return a partial resource in the response for those standard methods `iff` there is a need to do so.

Thus, we need to update the `request-unknown-fields` rules ([example](https://linter.aip.dev/133/request-unknown-fields)) to allow a `view`/`_view`-suffixed field, similar to Get ([example](https://linter.aip.dev/131/request-unknown-fields)).

noahdietz:
cc @quirogas

PR #1548, opened 2025-10-13
chore(rules): allow AIP-157 View field in standard methods

quirogas:
feat(rules): Allow AIP-157 View field in standard methods

This change updates the `request-unknown-fields` rules for AIP-133 (Create), AIP-134 (Update), and AIP-135 (Delete) to allow the use of a `view` field, as per the guidance in AIP-157 and recent changes in [aip-dev/google.aip.dev#1562](https://github.com/aip-dev/google.aip.dev/pull/1562). This allows services to return a partial resource in the response for these standard methods.

fixes:  #1547

### DIFF
--- a/rules/aip0133/request_unknown_fields_test.go
+++ b/rules/aip0133/request_unknown_fields_test.go
@@ -127,10 +127,6 @@ func TestUnknownFields(t *testing.T) {
 				}
 			`, test)
 
-			// Correct the field type for the ViewField test case.
-			if test.testName == "ViewField" {
-				test.FieldType = "BookView"
-			}
 			problems := unknownFields.Lint(f)
 			var d protoreflect.Descriptor = f.Messages().Get(0)
 			if test.problematicFieldName != "" {
--- a/rules/aip0135/request_unknown_fields_test.go
+++ b/rules/aip0135/request_unknown_fields_test.go
@@ -34,11 +34,11 @@ func TestUnknownFields(t *testing.T) {
 		{"AllowMissing", "DeleteBookRequest", "allow_missing", "bool", testutils.Problems{}},
 		{"RequestId", "DeleteBookRequest", "request_id", "string", testutils.Problems{}},
 		{"ValidateOnly", "DeleteBookRequest", "validate_only", "bool", testutils.Problems{}},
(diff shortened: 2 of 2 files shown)


rules/aip0133/request_unknown_fields_test.go
```diff
@@ -101,11 +117,20 @@ func TestUnknownFields(t *testing.T) {
...
+				}
 				message {{.MessageName}} {
 					{{.FieldType}} {{.FieldName}} = 1;
 				}
 			`, test)
 
+			// Correct the field type for the ViewField test case.
+			if test.testName == "ViewField" {
+				test.FieldType = "BookView"
+			}
```
noahdietz: Why is this necessary? Let's avoid adding logic to our tests ([TotT](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html)) and put it in a different/sub test if it is too hard to modify the table test.

Applies to here and below.

rules/aip0135/request_unknown_fields_test.go
```diff
@@ -34,11 +34,11 @@ func TestUnknownFields(t *testing.T) {
...
-		        {"ValidView", "DeleteBookRequest", "view", "BookView", testutils.Problems{}},
-				{"SuffixedView", "DeleteBookRequest", "custom_view", "BookView", testutils.Problems{}},
-				{"Invalid", "DeleteBookRequest", "application_id", "string", testutils.Problems{{
-					Message: "Unexpected field",
-				}}},		{"Irrelevant", "RemoveBookRequest", "application_id", "string", testutils.Problems{}},
+		{"ValidView", "DeleteBookRequest", "view", "BookView", testutils.Problems{}},
+		{"SuffixedView", "DeleteBookRequest", "custom_view", "BookView", testutils.Problems{}},
+		{"Invalid", "DeleteBookRequest", "application_id", "string", testutils.Problems{{
+			Message: "Unexpected field",
+		}}}, {"Irrelevant", "RemoveBookRequest", "application_id", "string", testutils.Problems{}},
```
noahdietz, line 41: Nit: what is going on with the formatting here, should the "Irrelevant" case be on its own line?

PR #1549, opened 2025-10-16
chore(v2): catch up with main (again)

noahdietz:
This PR catches up the `v2` branch with recent changes on `main`. This includes:
* Go version update
* change to AIP-140 URI field name linting
* rule doc fix

Note: It looks like many more commits are being merged than there are changes in this PR b.c I previously caught up the branch with these changes in #1535, but it was squashed, so it looks like they weren't!

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -20,7 +20,7 @@
   ],
   force: {
     constraints: {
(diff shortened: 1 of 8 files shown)


quirogas (approved):
LGTM!

PR #1559, opened 2025-10-24
fix(aip-136): Allow SetIamPolicy method

### LINKED ISSUE #1507: AIP-136: Allow SetIamPolicy
Hey @noahdietz,

it would be great to support `SetIamPolicy` as special edge case.

Currently AIP-136 do not allow SetIamPolicy. I think this worked in earlier versions.

Otherwise i really like `core::0136::response-message-name`.

Example:

```proto
  // (-- api-linter: core::0136::response-message-name=disabled
  //     aip.dev/not-precedent: Google. --)     <---- would love to remove this
  rpc SetIamPolicy(google.iam.v1.SetIamPolicyRequest) returns (google.iam.v1.Policy) {
    option (google.api.http) = {
      post: "/v1/{resource=projects/*/users/*}:setIamPolicy"
      body: "*"
    };
    option (google.api.method_signature) = "resource,policy";
  }
```

Thank you 🙏

quirogas:
This change addresses an issue where the `core::0136::response-message-name` linter rule was flagging the standard `SetIamPolicy` method.

  Fixes #1507

### DIFF
--- a/rules/internal/testutils/parse.go
+++ b/rules/internal/testutils/parse.go
@@ -30,13 +30,13 @@ import (
 	// These imports cause the common protos to be registered with
 	// the protocol buffer registry, and therefore make the call to
 	// `proto.FileDescriptor` work for the imported files.
+	_ "cloud.google.com/go/iam/apiv1/iampb"
 	_ "cloud.google.com/go/longrunning/autogen/longrunningpb"
 	_ "google.golang.org/genproto/googleapis/api/annotations"
 	_ "google.golang.org/genproto/googleapis/api/httpbody"
 	_ "google.golang.org/genproto/googleapis/type/date"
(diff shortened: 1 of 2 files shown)


noahdietz:
@quirogas please run `go mod tidy` if you haven't yet. There is something up with the dependency changes in this PR.

rules/internal/testutils/parse.go
```diff
@@ -36,6 +36,7 @@ import (
 	_ "google.golang.org/genproto/googleapis/type/date"
 	_ "google.golang.org/genproto/googleapis/type/datetime"
 	_ "google.golang.org/genproto/googleapis/type/timeofday"
+	_ "google.golang.org/genproto/googleapis/iam/v1"
```
noahdietz: Yeah we need to be using `cloud.google.com/go/iam/apiv1/iampb` here

PR #1561, opened 2025-10-28
feat(rules): allow AIP-157 view field in standard methods

### LINKED ISSUE #1547: Allow view field in AIP-133 AIP-134 and AIP-135 Soft Delete request messages

quirogas:
This is based on the [v2 implementation](https://github.com/googleapis/api-linter/pull/1548/files) of the same feature.

closes #1547

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,7 @@ name: ci
 on:
   pull_request:
     branches:
-      - main
+      - v1
 jobs:
   unit-tests:
     runs-on: ubuntu-latest


noahdietz:
We need CI to run on `v1`: https://github.com/googleapis/api-linter/pull/1562

rules/aip0133/request_unknown_fields_test.go
```diff
@@ -74,6 +74,20 @@ func TestUnknownFields(t *testing.T) {
 			testutils.Problems{},
 			nil,
 		},
+		{
+			"ViewField",
+			"CreateBookRequest",
+			[]field{{"view", builder.FieldTypeEnum(builder.NewEnum("BookView").AddValue(builder.NewEnumValue("BOOK_VIEW_UNSPECIFIED")).AddValue(builder.NewEnumValue("BOOK_VIEW_BASIC")).AddValue(builder.NewEnumValue("BOOK_VIEW_FULL")))}},
```
noahdietz: Nit: Does `gofmt` do anything to change this? Feels like too long of a line :)

noahdietz: Hey sorry you resolved but with no reply - does `gofmt` care about this or no? I'm guessing not given that the `lint` CI doesn't complain...

noahdietz: Turns out we aren't checking format in the `lint` job actually...

quirogas: My bad; I didn't realize I hadn't posted my comment. No, `gofmt` doesn't care about this, but I also feel like it's very long.

noahdietz: Well if you end up changing it, great, if not, no big deal. Approving to unblock as this is only a nit

PR #1564, opened 2025-11-07
fix(aip-0134): check word boundaries in synonyms rule

eddiesoller:
Fixes incorrect flagging of method names like "Settle" that start with synonym prefixes but are different words.

The synonyms rule was using simple prefix matching, which incorrectly flagged method names like "Settle", "PatchyClouds", or "PutterAround".

This PR adds word boundary detection by checking if the character after the prefix is uppercase (PascalCase word boundary) or if we've reached the end of the string.

Now "SetBook" correctly fails while "Settle" correctly passes.

### DIFF
--- a/rules/aip0134/synonyms.go
+++ b/rules/aip0134/synonyms.go
@@ -37,7 +37,7 @@ var synonyms = &lint.MethodRule{
 				synLen := len(syn)
 				nameLen := len(name)
 				// Check for word boundary: either exact match or next char is uppercase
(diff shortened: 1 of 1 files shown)


eddiesoller:
@gemini-code-assist review

eddiesoller:
@gemini-code-assist review

noahdietz:
Thanks @eddiesoller nice patch, and good use of Gemini code review. Isn't it particular? 😄

PR #1565, opened 2025-11-12
fix(aip-123): prevent panic in getParentIDVariable with single-variable patterns

eddiesoller:
## Summary
Fixes a panic in `getParentIDVariable` when processing resource patterns with only one variable that aren't root-level or singleton patterns.

## Problem
The function attempted to access `variables[len(variables)-2]`, which results in index -1 (panic) when there's only one variable. This occurred with patterns like `"nonCollectionPrefix/projects/{project}"`.

## Solution
Added a bounds check to return empty string when there are fewer than 2 variables. Invalid patterns are still caught by the `resource-name-components-alternate` rule.

## Testing
- Added integration test to verify no panic on invalid patterns
- Added unit tests for validation rule coverage
- All existing tests pass

## Changes
- Add bounds check in `getParentIDVariable()` (aip0123.go)
- Add integration test `TestAIP0123_InvalidPatternsDontPanic`
- Add unit tests in `TestResourceNameComponentsAlternate` for invalid patterns

### DIFF
--- a/rules/aip0123/aip0123.go
+++ b/rules/aip0123/aip0123.go
@@ -115,7 +115,7 @@ func getParentIDVariable(pattern string) string {
 		return variables[len(variables)-1]
 	}
 
-	// If there are fewer than 2 variables, there is no parent ID variable.
+	// If there are fewer than 2 variables for a non-singleton resource, there is no parent ID variable.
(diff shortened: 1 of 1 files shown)


eddiesoller:
@gemini-code-assist review

PR #1578, opened 2025-12-03
fix(AIP-133): avoid returning an error when return type is a message

### LINKED ISSUE #1577: "Create methods: Required fields" returns confusing message
#### Steps to reproduce

1. Create a proto file `reproduce_issue.proto` with a `Create` method that does not return the resource itself or an LRO:

    ```proto
    syntax = "proto3";

    package example.v1;

    option java_multiple_files = true;
    option go_package = "example/v1;v1";

    service LibraryService {
      // Creates a book.
      rpc CreateBook(CreateBookRequest) returns (CreateBookResponse) {}
    }

    message Book {
        type: "library.googleapis.com/Book"
        pattern: "books/{book}"
        singular: "book"
        plural: "books"
      };

      string name = 1;
      string author = 2;
    }

    message CreateBookRequest {
      // The parent resource where this book will be created.
      // Format: publishers/{publisher}
      string parent = 1 [
        (google.api.field_behavior) = REQUIRED,
        (google.api.resource_reference) = {
          child_type: "library.googleapis.com/Book"
        }
      ];

      // The book to create.
      Book book = 2 [(google.api.field_behavior) = REQUIRED];

      // The ID to use for the book, which will become the final component of
      // the book's resource name.
      string book_id = 3 [(google.api.field_behavior) = REQUIRED];
    }

    message CreateBookResponse {
      string status_message = 1;
      // Note: Does not return the Book resource.
    }
    ```

2. Create a `linter-config.yaml` to isolate the relevant rule:

    ```yaml
    disabled_rules:
      - "all"
    enabled_rules:
      - "core::0133::request-required-fields"
    ```

3. Run the API Linter:

    ```console
    api-linter reproduce_issue.proto --config linter-config.yaml -I path/to/googleapis
    ```

#### Observed Behavior
The linter incorrectly flags `book_id` and `book` as unexpected required fields:
- `Create RPCs must only require fields explicitly described in AIPs, not "book".`
- `Create RPCs must only require fields explicitly described in AIPs, not "book_id".`

This occurs because `aip0133.request-required-fields` relies on the response message type to identify the resource. Since `CreateBookResponse` is not the `Book` resource, the inference fails.

#### Expected Behavior
The linter should recognize `CreateBook` implies the resource is `Book` and accept `book_id` and `book` as valid required fields, as per AIP-133, even if the response is custom.

#### Suggested Improvement
Update `rules/aip0133/request_required_fields.go`:
If the response message is not the resource (or its LRO), attempt to derive the resource noun from the RPC name (e.g., `CreateBook` -> `Book`).

quirogas:
The `core::0133::request-required-fields` rule previously failed to correctly identify the resource when a `Create` RPC returned a custom response message instead of the resource itself or a standard LRO. This resulted in false positives where valid required fields (like `book_id` or `book`) were flagged as unexpected. Additionally, the rule did not strictly validate the types of the allowed required fields.

### Solution
This PR improves the rule by:
1.  **Fallback Resource Inference:** If the resource cannot be determined from the RPC's return type, the rule now attempts to infer the resource name from the RPC method name (e.g., inferring `Book` from `CreateBook`).
2.  **Strict Type Validation:** The rule now enforces that:
    *   `parent` field must be a `string`.
    *   The resource ID field (e.g., `book_id`) must be a `string`.
    *   The resource field (e.g., `book`) must be a `message`.
    
    If a field has the correct name but the wrong type, a specific error message is now returned.

### Changes
*   Modified `rules/aip0133/request_required_fields.go` to implement the fallback logic and refactor the field validation to use a map of expected types.
*   Updated `rules/aip0133/request_required_fields_test.go` to include test cases for custom response types and invalid field types.

Fixes #1577

### DIFF
--- a/docs/rules/0133/request-required-fields.md
+++ b/docs/rules/0133/request-required-fields.md
@@ -37,15 +37,6 @@ message CreateBookRequest {
 }
 ```
 
-```proto
-// Incorrect: The `book` field should be a message type (Book).
-message CreateBookRequest {
-  string parent = 1 [(google.api.field_behavior) = REQUIRED];
-  string book = 2 [(google.api.field_behavior) = REQUIRED];
-  string book_id = 3;
-}
-```
-
 **Correct** code for this rule:
 
 ```proto
--- a/rules/aip0133/request_required_fields.go
+++ b/rules/aip0133/request_required_fields.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"bitbucket.org/creachadair/stringset"
 	"github.com/googleapis/api-linter/v2/lint"
 	"github.com/googleapis/api-linter/v2/rules/internal/utils"
 	"github.com/stoewer/go-strcase"
@@ -34,8 +35,10 @@ var requestRequiredFields = &lint.MethodRule{
 			return nil
 		}
 
-		r := utils.GetResource(ot)
-		resourceMsgName := utils.GetResourceSingular(r)
+		var resourceMsgName string
+		if r := utils.GetResource(ot); r != nil {
+			resourceMsgName = utils.GetResourceSingular(r)
+		}
 
 		if resourceMsgName == "" {
 			if noun := utils.GetResourceMessageName(m, "Create"); noun != "" {
@@ -46,11 +49,11 @@ var requestRequiredFields = &lint.MethodRule{
 		snakeResourceName := strings.ToLower(strcase.SnakeCase(resourceMsgName))
 
 		// Rule check: Establish that there are no unexpected fields.
-		allowedRequiredFields := map[string]protoreflect.Kind{
-			"parent":                                protoreflect.StringKind,
-			fmt.Sprintf("%s_id", snakeResourceName): protoreflect.StringKind,
-			snakeResourceName:                       protoreflect.MessageKind,
-		}
+		allowedRequiredFields := stringset.New(
+			"parent",
+			fmt.Sprintf("%s_id", snakeResourceName),
+			snakeResourceName,
+		)
 
 		problems := []lint.Problem{}
 		for i := 0; i < m.Input().Fields().Len(); i++ {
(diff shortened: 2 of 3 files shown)


rules/aip0133/request_required_fields.go
```diff
@@ -34,33 +33,45 @@ var requestRequiredFields = &lint.MethodRule{
...
-			if !allowedRequiredFields.Contains(string(f.Name())) {
+			expectedKind, allowed := allowedRequiredFields[string(f.Name())]
+			if !allowed {
 				problems = append(problems, lint.Problem{
 					Message:    fmt.Sprintf("Create RPCs must only require fields explicitly described in AIPs, not %q.", f.Name()),
 					Descriptor: f,
 				})
+			} else if f.Kind() != expectedKind {
+				problems = append(problems, lint.Problem{
+					Message:    fmt.Sprintf("The required field %q must be of type %v.", f.Name(), expectedKind),
```
noahdietz: I like the thought, but this should be a separate rule. Each rule should try to have only one type of finding so that granular enablement/disablement is possible. We don't want them to disable this rule for having a non-standard required field, but then happen to also mess up their request resource field type and not know.

Please remove this logic and consider adding it as a separate rule in a diff pr.

docs/rules/0133/request-required-fields.md
```diff
@@ -37,6 +37,15 @@ message CreateBookRequest {
 }
 ```
 
+```proto
+// Incorrect: The `book` field should be a message type (Book).
+message CreateBookRequest {
+  string parent = 1 [(google.api.field_behavior) = REQUIRED];
+  string book = 2 [(google.api.field_behavior) = REQUIRED];
+  string book_id = 3;
+}
```
noahdietz: Ditto re removing this "extra rule" from the docs.

rules/aip0133/request_required_fields.go
quirogas: Thank you! I totally agree with you— funny enough I was debating on creating separating this check when  I wrote this. #1579 tracks this. I should be able to whip a PR for this reasonably fast.

noahdietz: Nice, thanks!

Issue #1574, opened 2025-11-26
core::0133::request-id-field false positive for AIP-155 idempotency keys


mo-cmyk:
## Summary

The `core::0133::request-id-field` rule flags Create requests as missing a client-specified ID field (e.g. `book_id`), but this creates false positives when `request_id` is used as an **idempotency key** per [AIP-155](https://google.aip.dev/155).

## Problem

AIP-133 allows client-specified resource IDs via fields like `{resource}_id` (e.g., `book_id` in `CreateBookRequest`).

AIP-155 recommends `request_id` as an idempotency key - a completely different use case:
- Purpose: deduplicate mutation requests from network retries
- Behavior: server returns cached response for duplicate request IDs
- Not a resource identifier

The linter strictly enforces AIP-133 and requires `{resource}_id`, failing to recognize that `request_id` satisfies the idempotency requirement.

## Example

```protobuf
message CreateBookRequest {
  string parent = 1 [(google.api.resource_reference).child_type = "library.googleapis.com/Book"];
  
  Book book = 2 [(google.api.field_behavior) = REQUIRED];
  
  // Idempotency key per AIP-155 - NOT a client-specified resource ID
  string request_id = 3 [
    (google.api.field_behavior) = OPTIONAL,
    (google.api.field_info).format = UUID4
  ];
}
```

This triggers:
```
core::0133::request-id-field: create methods should contain a singular `string book_id` field.
```

## Expected Behavior

The rule should not flag the missing `{resource}_id` when a valid AIP-155 idempotency key is present:
- Field name is exactly `request_id`
- Field behavior is `OPTIONAL`
- Field has `UUID4` format annotation

## Suggested Fix

Update the rule to check for `request_id` as a valid alternative when `{resource}_id` is missing.

## References

- [AIP-133: Create](https://google.aip.dev/133) - Client-specifiedgh issue edit 1574 --repo googleapis/api-linter --body-file - <<EOF
## Summary

The `core::0133::request-id-field` rule flags Create requests as missing a client-specified ID field (e.g. `book_id`), but this creates false positives when `request_id` is used as an **idempotency key** per [AIP-155](https://google.aip.dev/155).

## Problem

AIP-133 allows client-specified resource IDs via fields like `{resource}_id` (e.g., `book_id` in `CreateBookRequest`).

AIP-155 recommends `request_id` as an idempotency key - a completely different use case:
- Purpose: deduplicate mutation requests from network retries
[...]

PR #1155, opened 2023-05-19
feat(AIP-133): add request-id-field for non-df API

toumorokoshi:
Corresponds to the change to https://github.com/aip-dev/google.aip.dev/pull/1019.

resource_id is now required regardless of declarative_friendly.

fixes #1107

PR #1581, opened 2025-12-04
docs(AIP-133): remove declarative-friendly requirement from spec

mo-cmyk:
## Summary

- Fixes the `request-id-field` rule to only apply to declarative-friendly resources
- Aligns implementation with rule documentation which states it's for "declarative-friendly create methods"
- Adds test cases for non-declarative-friendly resources

## Problem

The `core::0133::request-id-field` rule documentation says it applies to "declarative-friendly create methods", but the implementation applied to ALL Create request messages. This caused false positives for non-declarative-friendly resources.

## Solution

Added `IsDeclarativeFriendlyMessage` check to the `OnlyIf` condition, matching the pattern used by other rules in the same package (e.g., `response_lro.go` at line 27).

## Test plan

- [x] Added test cases for non-declarative-friendly resources (`ValidNotDeclarativeFriendly`, `ValidNotDeclarativeFriendlyWithRequestId`)
- [x] All existing tests pass with `style: DECLARATIVE_FRIENDLY` annotation
- [x] Full test suite passes: `go test ./...`

Updates #1574

### DIFF
--- a/rules/aip0133/request_id_field_test.go
+++ b/rules/aip0133/request_id_field_test.go
@@ -31,8 +31,6 @@ func TestRequestIDField(t *testing.T) {
 		{"InvalidMissing", "", problems},
 		{"InvalidType", "bytes book_id = 2;", problems},
 		{"InvalidRepeated", "repeated string book_id = 2;", problems},
-		// request_id (AIP-155) is NOT a substitute for {resource}_id (AIP-133)
-		{"InvalidRequestIdOnly", "string request_id = 2;", problems},
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			f := testutils.ParseProto3Tmpl(t, `


rules/aip0133/request_id_field_test.go
```diff
@@ -21,16 +21,19 @@ import (
 )
 
 func TestRequestIDField(t *testing.T) {
-	problems := testutils.Problems{{Message: "`string book_id`"}}
+problems := testutils.Problems{{Message: "create methods should contain a singular `string book_id` field."}}
```
quirogas: We usually match only the key component of the message to give our testing framework some flexibility.
```suggestion
problems := testutils.Problems{{Message: "`string book_id`"}}
```

docs/rules/0133/request-id-field.md
```diff
@@ -15,15 +15,29 @@ client-specified ID field, as mandated in [AIP-133][].
...
 
-This rule looks at any `Create` method connected to a resource and complains if
-it lacks a client-specified ID (e.g. `string book_id`) field.
+This rule looks at any `Create` method connected to a declarative-friendly
+resource (one with `style: DECLARATIVE_FRIENDLY` in its `google.api.resource`
+annotation) and complains if it lacks a client-specified ID (e.g. `string
+book_id`) field.
+
+**Note:** This rule only applies to declarative-friendly resources. Resources
+without the `DECLARATIVE_FRIENDLY` style are not checked by this rule.
```
quirogas: This rule should still apply to methods that are not connected to a non-declarative-friendly resource.

quirogas (changes requested):
@mo-cmyk thank you for submitting this pull request. I think request_idshouldn't generate a lint message. However, we should discuss a few points in the original issue before agreeing on an overall solution.

mo-cmyk:
Hi @quirogas, thank you for the review!

I've addressed your feedback:

1. **Test message format** - Reverted to `"string book_id"` as per your suggestion ✓

2. **Documentation** - I've simplified the wording. Removed the verbose note and kept it concise.

Regarding your comment on the documentation:
> "This rule should still apply to methods that are not connected to a non-declarative-friendly resource"

I want to make sure I understand correctly. Currently the implementation:
- ✅ Applies to declarative-friendly resources (requires `{resource}_id`)
- ❌ Skips non-declarative-friendly resources

Are you saying the rule should also apply when there's **no resource connection at all**? If so, I can update the `OnlyIf` condition to handle that case.

Happy to discuss further on issue #1574 to align on the best approach. The original problem was false positives for `request_id` (AIP-155 idempotency key) - should that case also be considered?

mo-cmyk:
Thanks for the clarification on issue #1574, @quirogas! I've completely reworked this PR based on your feedback.

## Updated Approach

I now understand that:
- `{resource}_id` (AIP-133) = Resource Identity
- `request_id` (AIP-155) = Idempotency key (separate purpose)
- The rule should apply to **all** Create methods (as decided in PR #1155)

## Changes in this PR

1. **Code**: Reverted to original behavior - rule applies to all Create request messages (not just declarative-friendly)
2. **Documentation**: Removed the outdated "declarative-friendly" wording that was missed when PR #1155 updated the logic
3. **Tests**: Reverted to original test cases

This is now essentially a **documentation fix** to align the docs with the intended behavior.

Could you please re-review? Thank you!

mo-cmyk:
**Update**: Rebased on latest `main` and added a test case per your feedback on #1574.

### Changes
- Rebased on `main` (includes the AIP-216 fix from PR #1582)
- Added test case `InvalidRequestIdOnly` confirming the linter correctly flags missing `{resource}_id` even when `request_id` is present

The linter behaves as expected - `request_id` (AIP-155) is not treated as a substitute for `{resource}_id` (AIP-133). 

PR is ready for merge whenever you're ready @quirogas!

quirogas:
Removed test to keep this PR as a docs update only.

PR #1598, opened 2026-01-23
chore: return back golangci-lint

alexandear:
This PR re-enables linting via [golangci-lint](https://golangci-lint.run/), which was mistakenly replaced with staticcheck in #1084. Golangci-lint has never been deprecated.

Golangci-lint already includes [staticcheck](https://golangci-lint.run/docs/linters/configuration/#staticcheck) rules, so nothing is broken. In the future, we can also enable [other linters](https://golangci-lint.run/docs/linters/configuration/).

### DIFF
--- a/rules/internal/utils/casing.go
+++ b/rules/internal/utils/casing.go
@@ -36,8 +36,7 @@ func toCamelCase(s string, makeNextUpper bool, makeNextLower bool) string {
 			asLower = append(asLower, r)
 		} else if isUpper(r) {
 			if makeNextLower {
-				r = r | ' '           // make lowercase
-				makeNextLower = false //nolint:ineffassign
+				r = r | ' ' // make lowercase
 			}
 			asLower = append(asLower, r)
 		} else if isNumber(r) {


lint/rule.go
```diff
@@ -312,7 +312,7 @@ type DescriptorRule struct {
 	// is applicable.
 	OnlyIf func(protoreflect.Descriptor) bool
 
-	//lint:ignore U1000 ignored via golint previously
+	//nolint:unused // field is required to prevent positional parameters
```
noahdietz, line 315: Do we need the intervening `//` ?
```suggestion
	//nolint:unused field is required to prevent positional parameters
```

alexandear: @noahdietz Yes, we do.

`nolintlint.require-explanation` requires that `//`. See tests for the `nolintlint` linter: https://github.com/golangci/golangci-lint/blob/2ed365c7944938b15a969dd36be9637b6c9c0d9e/pkg/golinters/nolintlint/internal/nolintlint_test.go#L52

noahdietz: Great, thank you!

PR #1600, opened 2026-02-04
feat(cli): add support for linting descriptor sets only via --skip-compilation

quirogas:
Add the ability  to lint specific proto files directly from pre-compiled `FileDescriptorSet` files, bypassing the source compilation step.

## Key Changes
- **New Flag `--skip-compilation`**: Added a boolean flag that, when enabled, instructs the linter to look up requested files in the descriptor sets provided via `--descriptor-set-in` instead of compiling them from source.
- **Improved Descriptor Resolution**:
    - Refactored internal descriptor gathering logic into specialized methods: `getDescriptorsFromDescriptorSet` and `getDescriptorsFromSource`.
    - Consolidated descriptor set loading into a reusable helper `createRegistryFromDescriptorSets`, allowing for the aggregation of multiple descriptor sets.

## Integration Tests
Added comprehensive end-to-end tests in `cmd/api-linter/integration_test.go`:
- `TestSkipCompilation_Success`:
    - **SingleFile**: Verifies linting a single file from one descriptor set.
    - **MultipleDescriptorSets**: Verifies aggregating multiple descriptor sets and linting files from both.
- `TestSkipCompilation_Errors`:
    - **NoFileToLint**: Ensures the linter errors out if no files are specified with `--skip-compilation`.
    - **NoDescriptorSet**: Ensures the linter errors out if no descriptor sets are provided.

## Usage Example
To lint specific files using pre-compiled descriptors:
```bash
api-linter --descriptor-set-in=service.desc --skip-compilation google/api/service.proto
```

To aggregate multiple descriptor sets:
```bash
api-linter --descriptor-set-in=set1.desc --descriptor-set-in=set2.desc --skip-compilation file1.proto file2.proto
```

Update b/468026535

### DIFF
--- a/cmd/api-linter/cli.go
+++ b/cmd/api-linter/cli.go
@@ -225,7 +225,7 @@ func (c *cli) getDescriptorsFromDescriptorSet() ([]protoreflect.FileDescriptor, 
 		for _, fd := range fileDescriptors {
 			filenames = append(filenames, fd.Path())
 		}
-		return nil, fmt.Errorf("files found in descriptors %v, files request for linting %v", filenames, c.ProtoFiles)
+		return nil, fmt.Errorf("files found in descriptors %v, files requested for linting %v", filenames, c.ProtoFiles)
 	}
 
 	return fileDescriptors, nil
--- a/cmd/api-linter/integration_test.go
+++ b/cmd/api-linter/integration_test.go
@@ -604,7 +604,7 @@ func TestSkipCompilation_Errors(t *testing.T) {
 				"dummy.proto",
 				"missing.proto",
 			},
-			wantErrString: "files found in descriptors [dummy.proto], files request for linting [dummy.proto missing.proto]",
+			wantErrString: "files found in descriptors",
 		},
 	}
 


cmd/api-linter/cli.go
```diff
@@ -82,6 +85,7 @@ func newCli(args []string) *cli {
 	fs.BoolVar(&versionFlag, "version", false, "Print version and exit.")
 	fs.StringArrayVarP(&protoImportFlag, "proto-path", "I", nil, "The folder for searching proto imports.\nMay be specified multiple times; directories will be searched in order.\nThe current working directory is always used.")
 	fs.StringArrayVar(&protoDescFlag, "descriptor-set-in", nil, "The file containing a FileDescriptorSet for searching proto imports.\nMay be specified multiple times.")
```
noahdietz: Update the comment here to include reference to skip-compilation

cmd/api-linter/cli.go
```diff
@@ -144,10 +149,101 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
...
+	files, err := createRegistryFromDescriptorSets(c.ProtoDescPath...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Identify the files to lint.
+	filesToLint := make(map[string]bool)
+	for _, f := range c.ProtoFiles {
+		filesToLint[f] = true
+	}
```
noahdietz: I don't think this is necessary, `slices.Contains` is an api we can us in `files.RangeFiles`. https://pkg.go.dev/slices#Contains

Can delete this highlighted code and below replace `if filesToLink[fd.Path()] {` with `if slices.Contains(c.ProtoFiles, fd.Path()) {` and delete `delete(filesToLint, fd.Path()`

cmd/api-linter/cli.go
```diff
@@ -144,10 +149,101 @@ func (c *cli) lint(rules lint.RuleRegistry, configs lint.Configs) error {
...
+		return true // continue iteration
+	})
+
+	if len(filesToLint) > 0 {
+		var missing []string
+		for f := range filesToLint {
+			missing = append(missing, f)
+		}
+		sort.Strings(missing)
+		return nil, fmt.Errorf("files not found in descriptor set(s): %s", strings.Join(missing, ", "))
```
noahdietz: I think this is trying to provide a better error more experience and I like that. That might be reason enough to keep the above. But I think we can also just say `if len(fileDescriptors) != len(c.ProtoFiles)` and then do a 
```go
for _, f := range fileDescriptors { 
  filenames = append(filenames, fd.Path())
}
return nil, fmt.Errorf("files found in descriptors %v, files request for linting %v", filenames, c.ProtoFiles)`

I leave it to you to decide

File: cmd/api-linter/cli.go [hunk 4]
```diff
@@ -293,6 +350,17 @@ func (r *resolver) FindFileByPath(path string) (protocompile.SearchResult, error
 // object. It then wraps this object in our custom resolver so that it can be
 // used by the protocompile.Compiler to resolve imports.
 func loadFileDescriptorsAsResolver(filePaths ...string) (protocompile.Resolver, error) {
+	files, err := createRegistryFromDescriptorSets(filePaths...)
+	if err != nil {
+		return nil, err
+	}
+	if files == nil {
+		return nil, nil
```
noahdietz, line 351: Is this safe to do? Should we really be returning `nil` when there isn't anything? Just want to make sure the code invoking this method is prepared for that and we don't get `panic` nil dereference.

noahdietz (commented):
Please also change the [docs/index.md](https://github.com/googleapis/api-linter/blob/main/docs/index.md).

cmd/api-linter/cli.go
quirogas: Yes, we do. Returning nil, if anything, avoids invoking a resolver that won't be able to find anything.

cmd/api-linter/cli.go
```diff
@@ -211,29 +211,21 @@ func (c *cli) getDescriptorsFromDescriptorSet() ([]protoreflect.FileDescriptor,
...
-		for f := range filesToLint {
-			missing = append(missing, f)
+	if len(fileDescriptors) < len(c.ProtoFiles) {
+		var filenames []string
+		for _, fd := range fileDescriptors {
+			filenames = append(filenames, fd.Path())
 		}
-		sort.Strings(missing)
-		return nil, fmt.Errorf("files not found in descriptor set(s): %s", strings.Join(missing, ", "))
+		return nil, fmt.Errorf("files found in descriptors %v, files request for linting %v", filenames, c.ProtoFiles)
```
noahdietz: typo: `files requested...`

cmd/api-linter/integration_test.go
```diff
@@ -575,9 +601,10 @@ func TestSkipCompilation_Errors(t *testing.T) {
 			args: []string{
 				"--descriptor-set-in=internal/testdata/dummy.protoset",
 				"--skip-compilation",
+				"dummy.proto",
 				"missing.proto",
 			},
-			wantErrString: "files not found in descriptor set(s): missing.proto",
+			wantErrString: "files found in descriptors [dummy.proto], files request for linting [dummy.proto missing.proto]",
```
noahdietz: Nit: might not want to use a full error string here (because now you have to change it after I pointed out a typo). Pick a partial string e.g. the request file names porttion

PR #1621, opened 2026-04-09
fix: interpret span correctly for github format

### LINKED ISSUE #1071: GitHub Action output format generates incorrect line numbers

SoMuchForSubtlety:
The old location logic had two bugs
- keeping locations zero-indexed
- interpreting end column as end line in case of span size 3

~I decided to reuse the existing correct implementation from the lint package. I can re-implement it if you would prefer not to expose this API.~

I also fixed newlines, github does not support `\n`, but at least partially supports `%0A`, see https://github.com/actions/toolkit/issues/193

Closes #1071

### DIFF
--- a/cmd/api-linter/github_actions.go
+++ b/cmd/api-linter/github_actions.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/googleapis/api-linter/v2/lint"
+	dpb "google.golang.org/protobuf/types/descriptorpb"
 )
 
 // formatGitHubActionOutput returns lint errors in GitHub actions format.
@@ -34,8 +35,8 @@ func formatGitHubActionOutput(responses []lint.Response) []byte {
 			fmt.Fprintf(&buf, "::error file=%s", response.FilePath)
 
 			if problem.Location != nil {
-				loc := lint.FileLocationFromPBLocation(problem.Location, nil)
-				fmt.Fprintf(&buf, ",line=%d,endLine=%d,col=%d,endColumn=%d", loc.Start.Line, loc.End.Line, loc.Start.Column, loc.End.Column)
+				start, end := fileLocationFromPBLocation(problem.Location)
+				fmt.Fprintf(&buf, ",line=%d,endLine=%d,col=%d,endColumn=%d", start.Line, end.Line, start.Column, end.Column)
 			}
 
 			// GitHub uses :: as control characters (which are also used to delimit
@@ -53,4 +54,30 @@ func formatGitHubActionOutput(responses []lint.Response) []byte {
 	}
 
 	return buf.Bytes()
+}
+
+type position struct {
+	Line   int
+	Column int
+}
+
+// Implementation copied from lint/problem.go
+func fileLocationFromPBLocation(l *dpb.SourceCodeInfo_Location) (start, end position) {
+	start = position{
+		Line:   int(l.Span[0]) + 1,
+		Column: int(l.Span[1]) + 1,
+	}
+
+	if len(l.Span) == 4 {
+		end = position{
+			Line:   int(l.Span[2]) + 1,
+			Column: int(l.Span[3]),
+		}
+	} else {
+		end = position{
+			Line:   int(l.Span[0]) + 1,
+			Column: int(l.Span[2]),
+		}
+	}
+	return start, end
 }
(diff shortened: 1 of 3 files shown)


quirogas:
Hey @SoMuchForSubtlety, thank you so much for submitting this PR. We definitely do not want to expose this API, so I think a small re-implementation would be appropriate. Additionally, if you don't mind, could you also reference https://github.com/googleapis/api-linter/issues/1071 to determine if this change addresses that issue?

SoMuchForSubtlety:
@quirogas any chance of getting another review?

Issue #1628, opened 2026-05-13
Batch method rules do not respect google.api.resource singular annotation


stevenwarejones:
## Description

The batch method rules for AIP-0231, 0233, 0234, and 0235 (`request-requests-field`, `request-names-field`, `response-resource-field`) do not consult the `google.api.resource` annotation when determining the singular form of a resource name. Instead, they rely solely on `go-pluralize` for singularization, which can produce incorrect results.

For example, given a resource with an explicit `singular` annotation:

```proto
message ImpressionMetadata {
  option (google.api.resource) = {
    type: "example.com/ImpressionMetadata"
    singular: "impressionMetadata"
    plural: "impressionMetadata"
  };
}

message BatchUpdateImpressionMetadataRequest {
  repeated UpdateImpressionMetadataRequest requests = 2;
}
```

The linter reports:
```
The "requests" field on Batch Update Request should be a "UpdateImpressionMetadatumRequest" type
```

The `singular: "impressionMetadata"` annotation is ignored, and `go-pluralize` incorrectly applies a Latin grammar rule (`(?i)(dat)a$` → `$1um`) to produce "Metadatum".

## Reproducible Example

Running the linter on the following proto incorrectly reports that `requests` should be of type `UpdateImpressionMetadatumRequest`, even though the resource annotation explicitly declares `singular: "impressionMetadata"`:

```proto
syntax = "proto3";

import "google/api/resource.proto";

message ImpressionMetadata {
  option (google.api.resource) = {
    type: "example.googleapis.com/ImpressionMetadata"
    pattern: "dataProviders/{data_provider}/impressionMetadata/{impression_metadata}"
    singular: "impressionMetadata"
    plural: "impressionMetadata"
  };
}

message BatchUpdateImpressionMetadataRequest {
  repeated UpdateImpressionMetadataRequest requests = 1;
}

message UpdateImpressionMetadataRequest {}
```

## Root Cause

These 8 batch rules use `go-pluralize` directly to singularize the resource name extracted from the batch message name, without first checking if the resource message has a `google.api.resource` annotation with an explicit `singular` field.

This is the same class of bug that was fixed in #1573 for the `plural-method-name` rule, which now reads the `plural` field from `google.api.resource` before falling back to `go-pluralize`.

## Affected Rules

- `core::0231::request-names-field`
- `core::0231::response-resource-field`
- `core::0233::request-requests-field`
- `core::0233::response-resource-field`
- `core::0234::request-requests-field`
[...]

PR #1629, opened 2026-05-13
fix: Respect google.api.resource singular annotation in batch method rules

stevenwarejones:
## Summary

Batch method rules (AIP-0231, 0233, 0234, 0235) for `request-requests-field`, `request-names-field`, and `response-resource-field` used `go-pluralize` to singularize the resource name extracted from batch message names. This produced incorrect results for words like "Metadata", which `go-pluralize` singularizes to "Metadatum" using Latin grammar rules.

This PR adds a `DeriveResourceSingular()` helper that first checks for a `google.api.resource` annotation with a `singular` field on sibling or (transitively) imported resource messages before falling back to `go-pluralize`. This is analogous to the fix in #1573 for the `plural-method-name` rule.

When no `google.api.resource` annotation is found, the function falls back to `go-pluralize`, preserving existing behavior for all other resources.

## Affected Rules

- `core::0231::request-names-field`
- `core::0231::response-resource-field`
- `core::0233::request-requests-field`
- `core::0233::response-resource-field`
- `core::0234::request-requests-field`
- `core::0234::response-resource-field`
- `core::0235::request-names-field`
- `core::0235::response-resource-field`

## Changes

- **`rules/internal/utils/string_pluralize.go`**: Add `DeriveResourceSingular()`, which searches messages in the current file and its transitive imports (via the existing `utils.GetAllDependencies` helper) for a `google.api.resource` annotation with a matching `singular` value, converting it to UpperCamelCase. Falls back to `go-pluralize` if no annotation is found.
- **8 rule files across AIP-0231, 0233, 0234, 0235**: Replace `pluralize.NewClient().Singular()` calls with `utils.DeriveResourceSingular()`.
- **Rule tests across AIP-0231, 0233, 0234, 0235**: Add `...ResourceAnnotation` test cases for `ImpressionMetadata` with a `google.api.resource` annotation, verifying both the `request-*-field` and `response-resource-field` rules.

## Test plan

- Added resource annotation tests for all 8 affected rules across AIP-0231, 0233, 0234, and 0235
- Added unit tests for `DeriveResourceSingular()` covering: annotation in same file, annotation in imported file, fallback to go-pluralize, uncountable nouns, message name matching, and no-annotation Latin words
- Added multi-file import tests (`TestDeriveResourceSingularImportedFile` for a directly imported file, and `TestDeriveResourceSingularTransitiveImport` for a 3-file `service.proto -> common.proto -> resource.proto` chain) exercising the transitive import search path
- `go test -race ./...` passes (0 failures)
- `golangci-lint run ./...` reports 0 issues
- Verified against [cross-media-measurement](https://github.com/world-federation-of-advertisers/cross-media-measurement) protos: removing all 6 `Metadatum` disable comments produces zero false positives with the fixed linter

Issue: #1628

### DIFF
--- a/rules/aip0231/response_resource_field.go
+++ b/rules/aip0231/response_resource_field.go
@@ -29,7 +29,7 @@ var resourceField = &lint.MessageRule{
 	OnlyIf: isBatchGetResponseMessage,
 	LintMessage: func(m protoreflect.MessageDescriptor) []lint.Problem {
 		// The singular form the resource message name; the first letter capitalized.
-		pluralName := strings.TrimSuffix(strings.TrimPrefix(string(m.Name()), "BatchGet"), "Response")
+		pluralName := strings.TrimPrefix(strings.TrimSuffix(string(m.Name()), "Response"), "BatchGet")
 		resourceMsgName := utils.ResourceSingular(pluralName, m)
 
 		for i := 0; i < m.Fields().Len(); i++ {
--- a/rules/aip0231/response_resource_field.go
+++ b/rules/aip0231/response_resource_field.go
@@ -18,8 +18,8 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/gertd/go-pluralize"
 	"github.com/googleapis/api-linter/v2/lint"
+	"github.com/googleapis/api-linter/v2/rules/internal/utils"
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
@@ -29,8 +29,8 @@ var resourceField = &lint.MessageRule{
 	OnlyIf: isBatchGetResponseMessage,
 	LintMessage: func(m protoreflect.MessageDescriptor) []lint.Problem {
 		// The singular form the resource message name; the first letter capitalized.
-		plural := strings.TrimSuffix(strings.TrimPrefix(string(m.Name()), "BatchGet"), "Response")
-		resourceMsgName := pluralize.NewClient().Singular(plural)
+		pluralName := strings.TrimSuffix(strings.TrimPrefix(string(m.Name()), "BatchGet"), "Response")
+		resourceMsgName := utils.ResourceSingular(pluralName, m)
 
 		for i := 0; i < m.Fields().Len(); i++ {
 			fieldDesc := m.Fields().Get(i)
--- a/rules/aip0231/response_resource_field.go
+++ b/rules/aip0231/response_resource_field.go
@@ -30,7 +30,7 @@ var resourceField = &lint.MessageRule{
 	LintMessage: func(m protoreflect.MessageDescriptor) []lint.Problem {
 		// The singular form the resource message name; the first letter capitalized.
 		pluralName := strings.TrimPrefix(strings.TrimSuffix(string(m.Name()), "Response"), "BatchGet")
-		resourceMsgName := utils.ResourceSingular(pluralName, m)
+		resourceMsgName := utils.DeriveResourceSingular(pluralName, m)
 
 		for i := 0; i < m.Fields().Len(); i++ {
 			fieldDesc := m.Fields().Get(i)
--- a/rules/internal/utils/string_pluralize.go
+++ b/rules/internal/utils/string_pluralize.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/rules/internal/utils/string_pluralize.go
+++ b/rules/internal/utils/string_pluralize.go
@@ -54,9 +54,8 @@ func ResourceSingular(pluralName string, m protoreflect.MessageDescriptor) strin
 		if s := findResourceSingularInFile(pluralName, f); s != "" {
 			return s
 		}
-		// Also search directly imported files, since the resource message
-		// is often defined in a separate file (e.g. impression_metadata.proto)
-		// from the service file (impression_metadata_service.proto).
+		// Search direct imports only (not transitive); sufficient for standard proto
+		// layouts where the resource is in a sibling file imported by the service file.
 		imports := f.Imports()
 		for i := 0; i < imports.Len(); i++ {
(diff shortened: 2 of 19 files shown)


rules/aip0231/response_resource_field.go
```diff
@@ -29,8 +29,8 @@ var resourceField = &lint.MessageRule{
 	OnlyIf: isBatchGetResponseMessage,
 	LintMessage: func(m protoreflect.MessageDescriptor) []lint.Problem {
 		// The singular form the resource message name; the first letter capitalized.
-		plural := strings.TrimSuffix(strings.TrimPrefix(string(m.Name()), "BatchGet"), "Response")
-		resourceMsgName := pluralize.NewClient().Singular(plural)
+		pluralName := strings.TrimSuffix(strings.TrimPrefix(string(m.Name()), "BatchGet"), "Response")
```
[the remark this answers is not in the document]
stevenwarejones: Fixed in 9c9e3a5 — reordered to `TrimPrefix(TrimSuffix(...))` to match all other rules.

rules/internal/utils/string_pluralize.go
```diff
@@ -32,3 +34,70 @@ func ToPlural(s string) string {
...
+		}
+		// Also search directly imported files, since the resource message
+		// is often defined in a separate file (e.g. impression_metadata.proto)
+		// from the service file (impression_metadata_service.proto).
+		imports := f.Imports()
+		for i := 0; i < imports.Len(); i++ {
+			if s := findResourceSingularInFile(pluralName, imports.Get(i).FileDescriptor); s != "" {
+				return s
+			}
+		}
```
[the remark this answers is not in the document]
stevenwarejones: Good callout. Intentionally keeping this as direct imports only — in standard proto layouts the resource message is in a sibling file directly imported by the service file, so single-depth is sufficient. Recursive search would add complexity for a case we haven't seen in practice. Updated the comment to make this intentional, and added a multi-file test (TestResourceSingularImportedFile) that exercises the import search path. See 9c9e3a5.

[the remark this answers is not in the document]
stevenwarejones: On reflection, agreed — implemented recursive search with cycle detection via a visited set in 027891b. Also added  that exercises a 3-file chain (service.proto → common.proto → resource.proto) to verify the annotation is found through transitive imports.

rules/internal/utils/string_pluralize.go
```diff
@@ -32,3 +34,80 @@ func ToPlural(s string) string {
...
+// covers the file's entire import graph (recursively), since the resource
+// message may be defined in a separate file from the service and
+// request/response messages.
+//
+// If a matching resource annotation is found, its singular is returned (in
+// UpperCamelCase). Otherwise, it falls back to the go-pluralize library.
+//
+// This avoids incorrect singularization of words like "Metadata" (which
+// go-pluralize converts to "Metadatum" using Latin grammar rules).
+func ResourceSingular(pluralName string, m protoreflect.MessageDescriptor) string {
```
noahdietz: Let's use a different function name as I think this is a bit misleading alongside other utilities like `GetResourceSingular` (which arguably should've been named `ResourceSingular`).

This might be something like `DeriveResourceSingular` as it tries to find a resource from a derived name and then get the singular.

rules/internal/utils/string_pluralize_test.go
```diff
@@ -41,3 +43,182 @@ func TestPluralize(t *testing.T) {
...
+			}
+		`,
+	})
+
+	serviceFile := files["service.proto"]
+	m := serviceFile.Messages().Get(0)
+	got := ResourceSingular("ImpressionMetadata", m)
+	if got != "ImpressionMetadata" {
+		t.Errorf("ResourceSingular(\"ImpressionMetadata\") = %q, want \"ImpressionMetadata\"", got)
+	}
```
noahdietz, line 225: Please make `"ImpressionMetadata"` a variable `want`, then use that throughout and change the failure message from using `\"ImpressionMetadata\"` to `%q` 

Here and everywhere similar

noahdietz (commented):
Hi there, thanks for the extensive improvement. Couple of comments.

rules/internal/utils/string_pluralize_test.go
stevenwarejones: Done. Hoisted the literal into want := "ImpressionMetadata" and switched the failure messages to %q in both standalone import tests.

rules/internal/utils/string_pluralize.go
stevenwarejones: Done. Renamed to DeriveResourceSingular.

stevenwarejones:
@noahdietz - updated the PR description as well. lmk if you have any other things you'd like done. thx

noahdietz:
Need a second reviewer @quirogas

We will take a look at this next week. Thanks again for the PR and patience

noahdietz (approved):

quirogas (approved):
Thank you for submitting this fix! 

if you don't mind could you also add `DeriveResourceSingular`  to[ api-linter/rules/aip0235/request_parent_field.goL33](https://github.com/googleapis/api-linter/blob/219332a448ee2a93fd3107b86107b9b8cf9dedb6/rules/aip0235/request_parent_field.go#L33) before submitting 🚀

stevenwarejones:
> Thank you for submitting this fix!
> 
> if you don't mind could you also add `DeriveResourceSingular` to[ api-linter/rules/aip0235/request_parent_field.goL33](https://github.com/googleapis/api-linter/blob/219332a448ee2a93fd3107b86107b9b8cf9dedb6/rules/aip0235/request_parent_field.go#L33) before submitting 🚀

done

stevenwarejones:
> > Thank you for submitting this fix!
> > if you don't mind could you also add `DeriveResourceSingular` to[ api-linter/rules/aip0235/request_parent_field.goL33](https://github.com/googleapis/api-linter/blob/219332a448ee2a93fd3107b86107b9b8cf9dedb6/rules/aip0235/request_parent_field.go#L33) before submitting 🚀
> 
> done

@noahdietz - do you need to re-approve again to run the unit tests?

noahdietz:
Thanks @stevenwarejones !

PR #1636, opened 2026-06-26
ci: zizmor fix

suztomo:
The changes are from:

```
zizmor  --fix=all --gh-token=$(gh auth token) .github/workflows
```

go/github-zizmor-help?polyglot=github-com#auto-fixes-via-the-fix-flag

b/527976233


noahdietz (approved):
Thanks Tomo

PR #1639, opened 2026-08-03
chore(ci): fix action scan findings

noahdietz:
Address action scan issues.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,11 +61,10 @@ jobs:
         run: echo "raw_version=${TAG_NAME#v}" >> $GITHUB_OUTPUT
         shell: bash
       - name: Upload the ${{ matrix.osarch.os }}/${{ matrix.osarch.arch }} release.
-        uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5 # v1
         env:
           GITHUB_TOKEN: ${{ github.token }}
-        with:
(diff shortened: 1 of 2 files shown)


quirogas (approved):

.github/workflows/release.yaml
```diff
@@ -61,11 +61,10 @@ jobs:
...
-        with:
-          upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./api-linter.tar.gz
-          asset_name: api-linter-${{ steps.raw_tag.outputs.raw_version }}-${{ matrix.osarch.os }}-${{ matrix.osarch.arch }}.tar.gz
-          asset_content_type: application/tar+gzip
+          RAW_VERSION: ${{ steps.raw_tag.outputs.raw_version }}
+        run: |
+          gh release upload "$TAG_NAME" \
+            "./api-linter.tar.gz#api-linter-${RAW_VERSION}-${{ matrix.osarch.os }}-${{ matrix.osarch.arch }}.tar.gz" \
+            --clobber
```
quirogas, line 70: Could this overwrite releases accidentally?

quirogas (approved):

.github/workflows/release.yaml
noahdietz: It will just overwrite the release asset with the same name e.g. if it is retried. from the flag docs `Overwrite existing assets of the same name`. I think this is fine, especially if we need to press "retry" or do it locally.

